PR #2138, opened 2023-07-14
feat: Create Login w/Sentry Button (CODE-3592)

nicholas-codecov:
# Description

This PR brings in a new login button (hidden behind a flag) that will allow users to login via Sentry, once the new auth model is released on the API.

# Notable Changes

- Create new `loginProvider` utils as they will be different from our source control providers
- Update `LoginPage` to support having a Sentry login button when flag is enabled
- Update `LoginButton` to support a login with Sentry when flag is enabled or if not enabled and on `/login/sentry` redirect to `/login` or just return `null`

# Screenshots

`/login`:

Screenshot 2023-07-27 at 8 49 20 AM

`/login/sentry`:

Screenshot 2023-07-27 at 8 49 12 AM


adrian-codecov (approved):
Seems good but could you provide a screenshot again of how these look like please? Don't think they loaded properly

adrian-codecov (approved):

PR #2159, opened 2023-07-26
fix: Force All Contents Table Searches to Lower Case

### LINKED ISSUE #2157: File search is case sensitive, it should not be
On the coverage tab if you search for a file with caps it won't return a relevant result that is all lowercase

nicholas-codecov:
# Description

This PR closes #2157 by forcing all search param strings to lower case before sending to the API so that if the user so the search value param matches that of the requesting user.

# Notable Changes

- Update all `*ContentsTables` hooks to send search param string to lower case.

### DIFF
--- a/src/pages/CommitDetailPage/subRoute/CommitDetailFileExplorer/hooks/useRepoCommitContentsTable.js
+++ b/src/pages/CommitDetailPage/subRoute/CommitDetailFileExplorer/hooks/useRepoCommitContentsTable.js
@@ -1,4 +1,5 @@
 import isEqual from 'lodash/isEqual'
+import isString from 'lodash/isString'
 import { useCallback, useMemo } from 'react'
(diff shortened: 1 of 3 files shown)


adrian-codecov (approved):
Did you play around with this to ensure we're not case sensitive on the API? I don't think we are but worth double checking

nikosatwork:
Hi @nicholas-codecov. I have been following this PR - is there any progress?

It is not clear if search is still case sensitive, however it maps to lowercase search? For example, if there are these two files:
`Class.cs`
`Anotherclass.cs`
and in the search it is typed:
`Class`
then would this result in both files be listed or just `Anotherclass.cs`?

unity-nikos:
I think that PR makes more sense - thank you for the update Nicholas

nikosatwork:
I think that PR makes more sense - thank you for the update Nicholas

PR #2174, opened 2023-08-01
fix: Adjust Upgrade Plan Form Resolver When Seat Count is Lower then Activated Members (CODE-3642)

nicholas-codecov:
# Description

Since we're giving a vast amount of seats when users are trialing we can't block them from upgrading to an amount of seats that is lower then the current activated amount so this bypasses that check in the form resolver.

# Notable Changes

- Update `getSchema` function to return bypass the seat count vs activated member check if the current plan is on trial and trial is currently ongoing.
- Update `UpgradeForm` to pass along `trialStatus` value
- Update tests

### DIFF
--- a/src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/UpgradeForm.spec.jsx
+++ b/src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/UpgradeForm.spec.jsx
@@ -730,45 +730,47 @@ describe('UpgradeForm', () => {
       })
     })
 
-    describe('when the user chooses less than the number of active users', () => {
-      it('does not display an error', async () => {
-        const { user } = setup({
-          includeSentryPlans: true,
-          trialStatus: TrialStatuses.ONGOING,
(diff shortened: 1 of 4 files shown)


src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/UpgradeForm.spec.jsx
```diff
@@ -730,11 +731,12 @@ describe('UpgradeForm', () => {
     })
 
     describe('when the user chooses less than the number of active users', () => {
```
adrian-codecov, line 733: nitpick: add a "when a customer is undergoing a trial"

adrian-codecov (approved):

src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/UpgradeForm.spec.jsx
nicholas-codecov: Fixed

PR #2175, opened 2023-08-01
feat, ref: Fire mutation

RulaKhaled:
# Description
-> Firing trial mutation on continue to app 
-> Adjusting a bug of request param used in `useAccessGate` 

# Notable Changes
Move logic to App, since it's where we're handling home page redirects anyway!

# Link to Sample Entry

# Legal Boilerplate


src/App.jsx
```diff
@@ -26,15 +28,41 @@ const OwnerPage = lazy(() => import('./pages/OwnerPage'))
 const PullRequestPage = lazy(() => import('./pages/PullRequestPage'))
 const RepoPage = lazy(() => import('./pages/RepoPage'))
 
+const SetUpActions = Object.freeze({
```
adrian-codecov, line 31: Optional: I find this and the HomePageRedirect component could be in a different file. Feels it's taking away from what the app page is meant to represent

src/layouts/BaseLayout/hooks/useUserAccessGate.js
```diff
@@ -1,34 +1,10 @@
...
-  const { setup_action: setupAction } = params
-
-  if (setupAction === SetUpActions.REQUEST && matchProvider.isExact) {
-    updateDefaultOrg({ username })
-
-    return history.push(`/gh/${username}?setup_action=${SetUpActions.REQUEST}`)
-  }
-}
-
 // eslint-disable-next-line complexity, max-statements
```
adrian-codecov, line 8: Does this need to be here now that we got rid of the onboarding hook?

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
```diff
@@ -88,25 +101,28 @@ function DefaultOrgSelector() {
...
   })
 
-  const onSubmit = (data) => {
-    if (!data?.select) {
-      updateDefaultOrg({ username: currentUser?.user?.username })
-      return history.push(`/${provider}/${currentUser?.user?.username}`)
-    }
-
+  const onSubmit = () => {
     const segmentEvent = {
```
adrian-codecov, line 105: Optional: we're getting rid of Segment, so it might be worth not adding this at all

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
```diff
@@ -64,6 +67,16 @@ function DefaultOrgSelector() {
   const { data: currentUser, isLoading: userIsLoading } = useUser()
   const { mutate: updateDefaultOrg } = useUpdateDefaultOrganization()
 
+  const selectedOrg = orgValue?.org?.username ?? currentUser?.user?.username
```
adrian-codecov, line 70: I remember at one point we were using the information from the form itself, has that changed? Did we test changing the form's value and ensuring that's the one we ended up with?

src/pages/DefaultOrgSelector/DefaultOrgSelector.spec.jsx
```diff
@@ -60,12 +73,14 @@ describe('DefaultOrgSelector', () => {
     myOrganizationsData,
     useUserData,
     isValidUser = true,
+    trialStatus = 'NOT_STARTED',
+    planName = 'users-inappm',
```
adrian-codecov, line 77: Nitpick: this scenario could never happen, so maybe change this to either `users-basic` or `users-free`

adrian-codecov (approved):
Left some comments but otherwise approved

src/layouts/BaseLayout/hooks/useUserAccessGate.js
RulaKhaled: yes unfortunately

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
RulaKhaled: ohhh

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
RulaKhaled: this solved the select problem we had, Nick helped me come up with this solution and it resolved the issue with state

PR #2178, opened 2023-08-03
fix: Display Pre-Trial Benefits on Downgrade Card (CODE-3647)

nicholas-codecov:
# Description

Because the free plan card currently displays the current benefits the user has on their account this would display the trial plan benefits but that's not whats supposed to be displayed. So the API will have a new field introduced that will allow us to get the information for the plan the user had previously.

# Notable Changes

- Update `usePlanData` to fetch the pre-trial info
- Update `FreePlanCard` to display the pre-trial benefits if the user is currently on a trial
- Update tests

Closes codecov/applications-team#66


adrian-codecov (approved):
Looks good here

PR #2182, opened 2023-08-04
fix/rfc: use API_URL instead of BASE_URL for auth button hrefs

matt-codecov:
if i run Codecov locally and load `localhost:3000/login/gh` and click the "login" button, it sends me to `localhost:3000/login/gh` which is where i already was and nothing happens. if i modify it to send me to `localhost:8000/login/gh`, which is my local API instance, it will send me to a GitHub authorization page as i'd expect.

i don't understand why this isn't a problem in staging and production. am i missing anything?

# Legal Boilerplate

### DIFF
--- a/src/services/navigation/useNavLinks/useNavLinks.spec.js
+++ b/src/services/navigation/useNavLinks/useNavLinks.spec.js
@@ -36,13 +36,19 @@ describe('useNavLinks', () => {
     })
 
     it('Returns the correct link with nothing passed', () => {
+      // This is supposed to have REACT_APP_API_URL in front of it, but a lot
+      // of tests rely on that value being blank in `.env.test` so the test
+      // appears to assert that it is a same-site URL.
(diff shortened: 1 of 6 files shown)


src/services/navigation/useNavLinks/useNavLinks.spec.js
```diff
@@ -81,12 +96,15 @@ describe('useNavLinks', () => {
       )
       setup(['/gh/doggo/squirrel-locator'])
 
+      // This is supposed to have REACT_APP_API_URL in front of it, but a lot
+      // of tests rely on that value being blank in `.env.test` so the test
+      // appears to assert that it is a same-site URL.
       expect(
         hookData.result.current.signIn.path({
           to: 'htts://app.codecov.io/gh/codecov',
```
matt-codecov, line 104: is the missing p here intentional?

terry-codecov (approved):

PR #2194, opened 2023-08-11
ref + feat: Revert revert code 3571 enhance org selector onboarding flow to integrate codecov app installation

RulaKhaled:
# Description
Reverting the revet 
Hide request banner behind a feature flag 
We don't show it for self hosted 
tests 

# Legal Boilerplate


src/shared/GlobalTopBanners/RequestInstallBanner/RequestInstallBanner.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+  const { setup_action: setupAction } = params
+
+  const { defaultOrgSelectorPage: showBanner } = useFlags({
+    defaultOrgSelectorPage: false,
+  })
+
+  if (
+    !showBanner ||
+    (provider && providerToName(provider) !== 'Github') ||
```
RulaKhaled, line 25: was thinking we can alter the function itself a bit to handle these use-cases?

nicholas-codecov (approved):
Round Two 🥊

PR #2207, opened 2023-08-15
ref: Update useBranch to Use Repository GQL Field

nicholas-codecov:
# Description

Continuing along with dealing with codecov/engineering-team#355. This PR updates the `useBranch` hook to use the new `repository` field and handle any of the possible errors that may arise from the returned union type.

# Notable Changes

- Refactor `useBranch` to TS, use the new `repository` field, and handle errors from the union return type
- Update tests


src/services/branches/useBranch.tsx
```diff
@@ -0,0 +1,131 @@
...
+      Api.graphql({
+        provider,
+        query,
+        signal,
+        variables: {
+          owner,
+          repo,
+          branch,
+        },
+      }).then((res) => {
```
adrian-codecov, line 91: What would happen if we try to use this useBranch hook with a `select` statement somewhere in our codebase? If I recall correctly, the select would go first than the then, which is okay if we don't intend this hook with a `select`, is this the case?

adrian-codecov (approved):
Everything looks good, curious about the question I left though!

src/services/branches/useBranch.tsx
nicholas-codecov: That is not how `select` works. So `queryFn` requires a promise to be returned if it's not a promise it will not work. `select` runs off the data which is returned by the `queryFn` so it would have to wait until the promise that is returned is resolved. Here's a little example showing off that `select` is not ran until the promise has resolved in the `queryFn`.

> [!NOTE]
> _This code follows the same conventions that occur when you use `Api.graphql`_

Screenshot 2023-08-28 at 10 20 37 AM

Screenshot 2023-08-28 at 10 18 51 AM

PR #2208, opened 2023-08-15
ref: Quick Refactor to Parse the Whole Response of Converted Hooks

nicholas-codecov:
# Description

This quick refactor updates hooks that have already been converted over to use the new `repository` field to slightly adjust the return value as well as parse the whole response rather then a small portion.

# Notable Changes

- Update `useCommitHeaderData`
- Update `useCommitPageData`
- Update `usePullHeadData`
- Update `usePullPageData`


src/pages/CommitDetailPage/Header/hooks/useCommitHeaderData.tsx
```diff
@@ -132,7 +138,7 @@ export const useCommitHeaderData = ({
         }
 
         return {
-          commit: data?.repository?.commit,
+          commit: data?.owner?.repository?.commit ?? null,
```
RulaKhaled, line 141: so just so i understand this better, we can't return undefined here?

src/pages/PullRequestPage/hooks/usePullPageData.tsx
```diff
@@ -49,12 +49,18 @@ const RepositorySchema = z.object({
...
+      isCurrentUserPartOfOrg: z.boolean(),
+      repository: z
+        .discriminatedUnion('__typename', [
+          RepositorySchema,
+          RepoNotFoundErrorSchema,
+          RepoOwnerNotActivatedErrorSchema,
+        ])
+        .nullable(),
+    })
+    .nullable(),
```
RulaKhaled, line 63: do we plan to take out the queries to their own files?

RulaKhaled (commented):
two quick comments

src/pages/CommitDetailPage/Header/hooks/useCommitHeaderData.tsx
nicholas-codecov: It's because of the way that GraphQL works. If you request a field it will either return the value or `null` it will never return `undefined` so we don't wan't the TS value to reflect something it could never be. Which if we set to undefined here it won't pass the type validation that zod runs and will error out.

src/pages/PullRequestPage/hooks/usePullPageData.tsx
nicholas-codecov: I don't see any benefit as to removing the queries as when they're only used in a single location. I also think it's somewhat of a bad practice in these instances to move a string to it's own file.

RulaKhaled (approved):

PR #2210, opened 2023-08-15
Update Trial Copy and Tests

eliatcodecov:
# Description
Updates trial copy from "Trial Pro Team" to "Try Codecov Pro"


adrian-codecov (approved):
🥇

PR #2219, opened 2023-08-28
ref: useBranchCoverageMeasurements to TS and repository GQL Field

nicholas-codecov:
# Description

Continuing along with dealing with https://github.com/codecov/engineering-team/issues/355. This PR updates the `useBranchCoverageMeasurements` hook to use the new `repository` field and handle any of the possible errors that may arise from the returned union type.

# Notable Changes

- Refactor `useBranchCoverageMeasurements` to TS, use the new repository field, and handle errors from the union return type
- Update tests


src/pages/RepoPage/CoverageTab/CoverageTab.spec.jsx
```diff
@@ -191,21 +223,17 @@ describe('Coverage Tab', () => {
   it('renders the sunburst chart', async () => {
     setup()
 
-    render(<CoverageTab />, { wrapper: wrapper(['/gh/test-org/test-repo']) })
+    console.error = () => {}
```
terry-codecov, line 226: Doesnt this need to be cleaned up? Should we spy on the error instead?

PR #2225, opened 2023-08-29
Converts Icon to TypeScript

terry-codecov:
# Description
A fully typed Icon component. Kicking off the TS rust!

# Code Example
```
<Icon name="search" />
```

# Notable Changes
* To make use of type inference/auto complete I'm dropping support for camel case variations on the name prop.
* I updated TSX components using Icon to be camel case. Once we're fully over to TypeScript we can remove the camelCase function from Icon.
* Added data-icon for use with the new Table component. Not used in this pr.
* Fixed the svg module to correctly type .svg files, they're named exports not default exports.

# Screenshots

Screenshot 2023-08-29 at 6 50 35 PM
Screenshot 2023-08-29 at 6 51 08 PM
Screenshot 2023-08-29 at 6 51 30 PM
Screenshot 2023-08-29 at 6 52 14 PM

# Legal Boilerplate


src/ui/Icon/Icon.tsx
```diff
@@ -0,0 +1,76 @@
...
+  if (!IconSvg || !isValidKey(iconClasses, size)) return null
+
+  return (
+    <IconSvg
+      data-testid={label}
+      data-icon={label}
+      className={iconClasses[size]}
+    />
+  )
+}
```
nicholas-codecov, line 74: Just to take this to the next TypeScript level and make sure we don't introduce any errors because the variants don't have all of the same values for name, can you switch the props to the following so when you set the variant to `developer` you'll only see the names that are possible for it:

```tsx
interface CommonProps {
  size?: 'sm' | 'md' | 'lg' | 'flex'
  label?: string
}

type DeveloperIconProps = {
  name: keyof DeveloperIconCollection
  variant?: 'developer'
}

type OutlineIconProps = {
  name: keyof OutlineIconCollection
  variant?: 'outline'
}

type SolidIconProps = {
  name: keyof SolidIconCollection
  variant?: 'solid'
}

type IconProps = (DeveloperIconProps | OutlineIconProps | SolidIconProps) &
  CommonProps

function Icon({
  name,
  variant = 'outline',
  size = 'md',
  label = '',
}: IconProps) {
  const IconSvg = get(iconComponentCollection, variant, name)
  if (!IconSvg || !isValidKey(iconClasses, size)) return null

  return (
    <IconSvg
      data-testid={label}
      data-icon={label}
      className={iconClasses[size]}
    />
  )
}
```

nicholas-codecov (changes requested):
Quick change to have the types be narrowed to the given variant so we will get errors if we're trying to access a icon that does not belong to the given variant.

src/ui/Icon/Icon.tsx
terry-codecov: I wasn't able to get the type narrowing working correctly with `type IconProps = (DeveloperIconProps | OutlineIconProps | SolidIconProps) &
  CommonProps` but moving the intersection CommonProps up to each IconProps and unioning them seemed to have done the trick.
  
  ```
type OutlineIconProps = {
  name: keyof OutlineIconCollection
  variant: 'outline'
} & CommonProps
```

If anyone can why explain why this makes the difference to me that would be great. I would have assumed both ways were equivalent.

File: src/ui/Icon/Icon.tsx [hunk 2]
```diff
@@ -0,0 +1,91 @@
...
+
+type SolidIconProps = {
+  name: keyof SolidIconCollection
+  variant: 'solid'
+} & CommonProps
+
+type DeveloperIconProps = {
+  name: keyof DeveloperIconCollection
+  variant: 'developer'
+} & CommonProps
```
nicholas-codecov, line 69: You will need to set `variant` here as optional with `?:` or you'll get build errors wherever a variant isn't passed in. Unless you want to force a variant to be passed.

nicholas-codecov (changes requested):
Quick change is needed to fix the build.

src/ui/Icon/Icon.tsx
terry-codecov: Whoops my bad, on it.

nicholas-codecov (approved):
LGTM :shipit:

PR #2235, opened 2023-08-31
feat: Create Provider Sync Page

nicholas-codecov:
# Description

This PR creates a page that allows user to sync various git providers with their account when signing in with a non-git provider i.e. Sentry, Okta, etc. This PR also includes changes to the `useUserAccessGate` hook which adds in a check to see if the user has synced any providers and if not then redirect them to the `/sync` route.

# Notable Changes

- Create `SyncProviderPage`, `SyncProviderButton` components
- Update `useUserAccessGate` to support checks for synced providers
- Update `BaseLayout` to add check for synced providers after TOS and before selecting default org
- Add `/sync` route to `App`

# Screenshots

Screenshot 2023-08-31 at 9 02 25 AM


src/App.jsx
```diff
@@ -55,6 +56,13 @@ const MainAppRoutes = () => (
         <AccountSettings />
       </BaseLayout>
     </SentryRoute>
+    {!config.IS_SELF_HOSTED && (
+      <SentryRoute path="/sync" exact>
```
RulaKhaled, line 60: Why didn't you move the sync page to base layout?

src/layouts/BaseLayout/BaseLayout.jsx
```diff
@@ -25,24 +26,34 @@ const FullPageLoader = () => (
...
+  if (showAgreeToTerms && !isFullExperience) {
     return (
       <Suspense fallback={null}>
         <TermsOfService />
       </Suspense>
     )
+  }
+
+  if (redirectToSyncPage && !isFullExperience) {
+    return <Redirect to="/sync" />
```
RulaKhaled, line 47: what if i'm already in /sync route? does this mean i'll still get redirected to /sync?

src/layouts/BaseLayout/hooks/useUserAccessGate.js
```diff
@@ -31,47 +34,88 @@ function useOnboardingRedirect({ username }) {
...
+    data: internalUser,
+    isLoading: internalUserIsLoading,
+    isSuccess: internalUserIsSuccess,
+    isFetching: internalUserIsFetching,
+  } = useInternalUser({
+    suspense: false,
+    enabled: !config.IS_SELF_HOSTED,
+  })
+
+  const foundUser = !userData && userIsSuccess
```
RulaKhaled, line 68: can you change this name? maybe `missingUser`

src/layouts/BaseLayout/hooks/useUserAccessGate.js
```diff
@@ -31,47 +34,88 @@ function useOnboardingRedirect({ username }) {
...
+  const foundInternalUser = !internalUser && internalUserIsSuccess
+  const isGuest = foundUser || foundInternalUser
```
RulaKhaled, line 70: also found internal user please

src/layouts/BaseLayout/hooks/useUserAccessGate.js
```diff
@@ -31,47 +34,88 @@ function useOnboardingRedirect({ username }) {
...
 
-  // If we don't have a provider, we cant check the TOS agreement, so we alow the user to access the full experience.
-  if (!provider) {
-    return {
-      isFullExperience: true,
-      isLoading: false,
-      showAgreeToTerms,
-      showDefaultOrgSelector,
-    }
+  const onSyncPage = currentRoute.path === '/sync'
```
RulaKhaled, line 87: can you do `const match = useRouteMatch('/sync');` to check if match.isExact, it directly checks if there is an exact match for the '/sync' route, which is more concise and clearer in its intent

src/layouts/BaseLayout/hooks/useUserAccessGate.js
```diff
@@ -31,47 +34,88 @@ function useOnboardingRedirect({ username }) {
...
+      (userIsLoading && userIsFetching) ||
+      (internalUserIsLoading && internalUserIsFetching)
   }
 
   // Not fully tested logic yet, waiting on API to be available.
   // Assuming self hosted users do not need to sign
   return {
     isFullExperience:
-      !!config.IS_SELF_HOSTED || (!showAgreeToTerms && !showDefaultOrgSelector),
+      !!config.IS_SELF_HOSTED ||
```
RulaKhaled, line 113: you're already checking if it's self hosted here no need to check for self hosted in App.jsx

src/pages/SyncProviderPage/SyncProviderPage.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  // block all requests if flag is false
+  // --
+  // this allows us to also show the sync page within the limited
+  // experience env because if they don't have any providers selected
+  // then we need to have them select one before they move onto onboarding
+  // --
+  // once we allow users to sync multiple providers need change this logic
+  // to allow them to view this page
+  if (!sentryLoginProvider || hasSynced) {
```
RulaKhaled, line 29: so in `useUserAcessGate` -> line 88:

>  if (sentryLoginProvider && !isGuest && !onSyncPage) {
>     // owners array contains a list of the synced providers
>     // if it is zero then they haven't synced any other providers
>     redirectToSyncPage = isEqual(internalUser?.owners?.length, 0)
>   }

you're already checking for this logic there

RulaKhaled (changes requested):
left some comments, i'm mainly wondering why you chose to place sync under app and handle conditions in two places instead of handling things in base layout the way we do for tos and default org selector

src/layouts/BaseLayout/hooks/useUserAccessGate.js
nicholas-codecov: The reason why I have done it like this is because it is expandable in the future and not limited just to the `/sync` route. I think it is better left this way when/if the time comes we can quickly check the path without having any hassle.

src/layouts/BaseLayout/BaseLayout.jsx
nicholas-codecov: This code block here checks to see if you're already on the `/sync` route, and if you are it will set `redirectToSyncPage` to `false`.

```jsx
const onSyncPage = currentRoute.path === '/sync'
if (sentryLoginProvider && !isGuest && !onSyncPage) {
  // owners array contains a list of the synced providers
  // if it is zero then they haven't synced any other providers
  redirectToSyncPage = isEqual(internalUser?.owners?.length, 0)
}
```

src/App.jsx
nicholas-codecov: Not sure I understand what you mean? The sync page is wrapped by the base layout.

nicholas-codecov: Ohh I see what you mean here now, see [here](https://github.com/codecov/gazebo/pull/2235#issuecomment-1718021025)

src/layouts/BaseLayout/hooks/useUserAccessGate.js
nicholas-codecov: Done

src/layouts/BaseLayout/hooks/useUserAccessGate.js
nicholas-codecov: Done

src/layouts/BaseLayout/hooks/useUserAccessGate.js
nicholas-codecov: So we may want to enable some of these features inside `useUserAccessGate` to self hosted users potentially, so if that was removed and also removed in `App.jsx` it could potentially show the page accidentally. I'd prefer to keep it the way it is so it will explicitly not show while running in self-hosted mode.

RulaKhaled: would we ever show sync page for self hosted though?

src/layouts/BaseLayout/BaseLayout.jsx
RulaKhaled: makes sense

src/pages/SyncProviderPage/SyncProviderPage.tsx
RulaKhaled: so doesn't this mean the page will only be rendered if a user hasn't synced any providers? `useUserAccessGate` takes care of the redirect to /sync, will there be ever a case where the sync page is rendered when a user has already synced a provider?

nicholas-codecov: Yes in the future, we are planning to allow users to sync multiple providers

src/layouts/BaseLayout/hooks/useUserAccessGate.js
nicholas-codecov: Yes, I believe the eventual plan is to allow other OAuth providers on self-hosted

src/App.jsx
scott-codecov: @nicholas-codecov FYI - we'll want this available in enterprise environments as well (which I think is what `IS_SELF_HOSTED` means?)

nicholas-codecov: I was told/under the impression that this will be released later, not right away.

nicholas-codecov: It's also currently behind a feature flag so it won't work anyways on self-hosted, even if this was removed I think*

scott-codecov: Looking to release Okta login for a new dedicated cloud deployment which would also need this `/sync` page available.  Is it possible to remove the feature flag as well?  There's nothing other than the API that links to this new page, right?

trent-codecov: Alternatively could the feature flag be passed in as a build time config env or gate it to app_env enterprise?

nicholas-codecov: Once cloud has been tested and working a day or two, I will remove the flag, and then at the same time I'll remove the self hosted checks. 

Also to release Okta work will have to be done to the enterprise login page to even support it, so that has to be done before any release to dedicated cloud is done.

nicholas-codecov: Something else I forgot to mention, is that we also have to fix the URLs that Gazebo currently uses to login/logout users, since they are currently using the base url not api url.

RulaKhaled (approved):

PR #2249, opened 2023-09-07
ref: Update "team bot missing" for users that have GitHub app to "Issue with Github app" 

RulaKhaled:
# Description
Add one more case to missing bot banner to handle users with github app but missing bot. 

# Screenshots
Screenshot 2023-09-06 at 7 08 07 PM

# Legal Boilerplate


src/pages/CommitDetailPage/BotErrorBanner/BotErrorBanner.jsx
```diff
@@ -22,57 +22,90 @@ function useProviderSetting() {
...
+  if (ghWithNoApp) {
+    return (
+      <div className="gap-1 lg:flex lg:w-max">
+        <span>
+          The bot posts the coverage report comment on pull requests. If
+          you&apos;re using GitHub, the best way to integrate with Codecov.io is
+          to Install
+        </span>
+        <A to={{ pageName: 'codecovGithubApp' }}> Codecov&apos;s GitHub App.</A>
+      </div>
```
nicholas-codecov, line 50: Wanna switch this to a `<p>` 👀

nicholas-codecov (approved):
One quick thing that won't require a review. Once that's done feel free to ship away!

src/pages/CommitDetailPage/BotErrorBanner/BotErrorBanner.jsx
nicholas-codecov: Also just noticed the `lg:w-max`, that is not a valid tailwind class the `w-max` specifically. See [here](https://tailwindcss.com/docs/max-width#reading-width).

RulaKhaled: nice catch, i think we can safely get rid of all classes in this div

PR #2250, opened 2023-09-07
Setup a fileviewer list to replace old table

terry-codecov:
# Description
Less documentation then the Tableui, migration/DX is more or less identical, different HTML markup requirements (not using a proper table). I made a fairly polished demo in storybook.

# Notable Changes
* Converted CodeRenderer to typescript
* Tweaked/improved TableUI style/examples if I found something better.
* Code renderer styles were affecting things outside of itself, added a classname to scope it down to only it.
* Added expanded support to table + file list

# Screenshots
image

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/ui/CodeRenderer/CodeRenderer.css
+++ b/src/ui/CodeRenderer/CodeRenderer.css
@@ -1,18 +1,18 @@
-tbody,
-tr,
-td {
+.coderenderer tbody,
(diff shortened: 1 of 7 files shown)


nicholas-codecov (approved):
:shipit:

PR #2261, opened 2023-09-13
ref: Update usePulls to Repository Field and to TypeScript

nicholas-codecov:
# Description

This PR updates the `usePulls` hook to TypeScript and to use the `repository`, and `percentCovered` GQL fields as their old ones are now deprecated.

# Notable Changes

- Update `usePulls` to use `repository` over `repositoryDeprecated`
- Update `usePulls` to use `percentCovered` over `coverage`
- Update `usePulls` to TypeScript
- Update `PullsTable` to use updated `usePulls` hook
- Update `PullsTable/Coverage` to use updated passed data


src/services/pulls/usePulls.tsx
```diff
@@ -0,0 +1,282 @@
...
+          for (const edge of edges) {
+            if (edge?.node) {
+              branches.push(edge.node)
+            }
+
+            if (edge === null) {
+              branches.push(edge)
+            }
+          }
+        }
```
adrian-codecov, line 265: I believe you could use `mapEdges` here, although that implementation doesn't check 261 (which makes me wonder why). It must be that connections cannot have null edges in between 🤔 otherwise, we could probably adjust the mapEdges fn or use a utility to mapEdges that accounts for null edges 🤷

adrian-codecov (approved):
lgtm

PR #2264, opened 2023-09-14
Running GHA tests

trent-codecov:
# Description

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


.github/workflows/ci.yml
```diff
@@ -0,0 +1,275 @@
...
+          path: |
+            node_modules
+          key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-
+
+      - name: Install packages
+        if: ${{ steps.cache-node-modules.outputs.cache-hit != 'true' }}
+        run: |
+          npm ci
```
nicholas-codecov, line 45: I would recommend trying out [this](https://bun.sh/guides/install/cicd) if the cache is missed, should be a lot faster install time over `npm`, and outputs the same `node_modules` directory

.github/workflows/ci.yml
```diff
@@ -0,0 +1,275 @@
...
+          path: |
+            node_modules
+          key: ${{ runner.os }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ env.cache-name }}-
+
+      - name: Install packages
+        if: ${{ steps.cache-node-modules.outputs.cache-hit != 'true' }}
+        run: |
+          npm run build
```
nicholas-codecov, line 71: Think may be a bit of copy paste error here, by this step we should have `node_modules` present so you probably don't need to cache them again, as well you should probably always be good to `npm run build` without a check for a cache miss. As well the last step here probably should be named `Build React App`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,275 @@
...
+            ${{ runner.os }}-${{ env.cache-name }}-
+
+      - name: Install packages
+        if: ${{ steps.cache-node-modules.outputs.cache-hit != 'true' }}
+        run: |
+          npm ci
+  build:
+    name: Build App
+    runs-on: ubuntu-latest
+    needs: [install]
```
nicholas-codecov, line 49: We can probably wait to build until the tests pass, no sense building the application if the tests fail, and the build isn't required for the tests to run.

nicholas-codecov (changes requested):
Few things to take a peak at 🙂

.github/workflows/ci.yml
trent-codecov: I'll test it out, but my understanding is that each of these steps runs in isolation, so for node_modules to exist, I need to call the cache step. I'm using cache less as an ephemeral cache and more as a persistent store between jobs.

PR #2268, opened 2023-09-15
feat: Create Provider Sync Page ... Round 2

nicholas-codecov:
# Description

This PR creates a page that allows user to sync various git providers with their account when signing in with a non-git provider i.e. Sentry, Okta, etc. This PR also includes changes to the `useUserAccessGate` hook which adds in a check to see if the user has synced any providers and if not then redirect them to the `/sync` route.

# Notable Changes

- Create `SyncProviderPage`, `SyncProviderButton` components
- Update `useUserAccessGate` to support checks for synced providers
- Update `BaseLayout` to add check for synced providers after TOS and before selecting default org
- Add `/sync` route to `App`

# Screenshots

Screenshot 2023-08-31 at 9 02 25 AM


src/layouts/BaseLayout/hooks/useUserAccessGate.js
```diff
@@ -31,47 +33,88 @@ function useOnboardingRedirect({ username }) {
...
 
-  useOnboardingRedirect({ username: data?.user.username })
+  const {
+    data: internalUser,
+    isLoading: internalUserIsLoading,
+    isFetching: internalUserIsFetching,
+    isSuccess: internalUserIsSuccess,
+  } = useInternalUser({
+    enabled: !config.IS_SELF_HOSTED,
+    retry: 0,
```
RulaKhaled, line 62: Can we do `retry = false` as listed in https://tanstack.com/query/v4/docs/react/guides/query-retries just to make sure it’s set correctly

PR #2269, opened 2023-09-15
feat: show the global upload token for non-admin folks

adrian-codecov:
# Description
We don't want to show the global upload token tab for non-admin customers. This PR hides that page and handles the redirect for the upload token route

# Notable Changes
- Adjusts the logic in `AccountSettings`/`AccountSettingsSideMenu` to only show this page if you're an admin

# Legal Boilerplate


nicholas-codecov (approved):
PR title is a little confusing .... but after reading the description looks all good to me 👍

PR #2271, opened 2023-09-18
feat: Use API_URL Instead of BASE_URL

nicholas-codecov:
# Description

Currently we're using the `BASE_URL` value for sign in and sign out which isn't compatible with Login w/Sentry so this PR updates Gazebo to use the `API_URL` instead. As well these changes utilizes the changes made to the API to allow a `to` query parameter to be passed to `/logout` and have the user redirect to that value.

# Notable Changes

- Switch over to using `API_URL` over `BASE_URL` for sign in and sign out
- Add query params to sign out
- Add `to` query param to sign out in `Dropdown` so users are redirected to the marketing site (TBC)
- Refactor `useNavLink` tests

### DIFF
--- a/src/layouts/Header/Dropdown.jsx
+++ b/src/layouts/Header/Dropdown.jsx
@@ -14,8 +14,6 @@ function Dropdown({ currentUser }) {
   const { provider } = useParams()
   const isGh = providerToName(provider) === 'Github'
 
-  const to = `${window.location.protocol}//${window.location.host}/${provider}`
-
   const items =
     !config.IS_SELF_HOSTED && isGh
       ? [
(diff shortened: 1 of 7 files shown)


src/layouts/Header/Dropdown.jsx
```diff
@@ -14,6 +14,8 @@ function Dropdown({ currentUser }) {
   const { provider } = useParams()
   const isGh = providerToName(provider) === 'Github'
 
+  const to = `${window.location.protocol}//${window.location.host}/login`
```
terry-codecov, line 17: Is there a reason you're accessing the window directly instead of using useLocation? I'm pretty sure its a proxy of window but it would make testing/mocking this easier then a global in the spec.

terry-codecov (approved):

PR #2282, opened 2023-09-23
Setup pull request page to pass around selected flags

terry-codecov:
# Description
Updates the pull request page to pass around the flags multi select url params through out regular navigation. This maintains the flag selection when traversing the tabbed navigation and files/folders.

# Notable Changes
* Use nav links was missing some tests so I also added them in
* Some of dir/file links dont match the other navigation links, I left them as is to avoid a larger refactor but will likely need to address in the TS refactor cause that's not going to fly with the typings.

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/services/navigation/useNavLinks/useNavLinks.js
```diff
@@ -155,12 +163,21 @@ export function useNavLinks() {
     },
     commit: {
       path: (
-        { provider = p, owner = o, repo = r, commit } = {
+        { provider = p, owner = o, repo = r, commit = c, ...queryParams } = {
           provider: p,
           owner: o,
           repo: r,
+          commit: c,
+          flags: [],
```
nicholas-codecov, line 171: Should this be `queryParams: {}`

src/shared/ContentsTable/TableEntries/PullEntries/PullDirEntry.jsx
```diff
@@ -1,10 +1,15 @@
...
+import { useLocation } from 'react-router-dom'
 
 import { usePrefetchPullDirEntry } from 'services/pathContents/pull/dir'
 
 import DirEntry from '../BaseEntries/DirEntry'
 
 function PullDirEntry({ pullId, urlPath, name, filters }) {
+  const { search } = useLocation()
+  const searchParams = qs.parse(search, { ignoreQueryPrefix: true })
+
```
nicholas-codecov, line 12: The filters here will include the search params, so you don't need to grab them from the URL

src/shared/ContentsTable/TableEntries/PullEntries/PullFileEntry.jsx
```diff
@@ -13,6 +15,9 @@ function PullFileEntry({
   urlPath,
   displayType,
 }) {
+  const { search } = useLocation()
+  const queryParams = qs.parse(search, { ignoreQueryPrefix: true })
+
```
nicholas-codecov, line 20: Same thing here, you can pass a filters prop and have them passed down from the contents table hook that is using these components

nicholas-codecov (commented):
Few more comments to take a peak at

src/shared/ContentsTable/TableEntries/PullEntries/PullFileEntry.jsx
terry-codecov: There is no filters in the PullFileEntry, I'm unclear on how I'd be able to extend this like the one above. Could you clarify?

src/shared/ContentsTable/TableEntries/PullEntries/PullFileEntry.jsx
```diff
@@ -6,16 +6,20 @@ import { displayTypeParameter } from '../../constants'
...
   urlPath,
   displayType,
+  filters,
 }) {
+  const flags = filters?.flags?.length > 0 ? filters?.flags : []
+
   const { runPrefetch } = usePrefetchPullFileEntry({
     path,
-    ref: commitSHA,
+    commitSha,
```
nicholas-codecov, line 21: I don't see the `usePrefetchPullFileEntry` being updated in the PR so you'll still need to alias `ref` to the `commitSha`

nicholas-codecov (changes requested):
One quick thing to peak at.

src/shared/ContentsTable/TableEntries/PullEntries/PullFileEntry.jsx
terry-codecov: Ah shoot good catch, not sure how that happened. Ty

nicholas-codecov (approved):
LGTM 👍

PR #2291, opened 2023-09-27
414 add tier hook

adrian-codecov:
# Description
We're introducing the notion of a tier based on your plan. We want to access that throughout the app, so we're making a generic hook to access in multiple places throughout the app.

# Notable Changes
- Add `useTier` hook
- Add test
- **Typescript**

# Legal Boilerplate


src/services/tier/useTier.ts
```diff
@@ -0,0 +1,48 @@
...
+  LITE: 'lite',
+  PRO: 'pro',
+  ENTERPRISE: 'enterprise',
+} as const
+
+export const TierSchema = z
+  .object({
+    tierName: z.nativeEnum(TierNames),
+  })
+  .nullish()
```
nicholas-codecov, line 17: Can you use this as a child of a `RequestSchema` and parse all the way down from the owner to make sure we're validating the whole response.

src/services/tier/useTier.ts
```diff
@@ -0,0 +1,48 @@
...
+    queryFn: ({ signal }) =>
+      Api.graphql({
+        provider,
+        query,
+        signal,
+        variables: {
+          owner,
+        },
+      }).then((res) => {
+        return TierSchema.parse(res?.data?.owner?.plan) ?? {}
```
nicholas-codecov, line 46: Can you switch this to a safe parse and check to see if the request was successful and if not return a rejected promise so that we can handle that error a little more gracefully, as right now it won't come up as a network error but as a regular JS error that crashed the app.

nicholas-codecov (changes requested):
Few things to take a quick peak at around the use of zod, handling errors, and validating the full response.

src/services/tier/useTier.ts
adrian-codecov: Sureee! When should we use parse vs safeParse? I suppose we prefer on safeParse and handle promises right?

nicholas-codecov: Should always use safeParse ... sorry for my previous indiscretions. It's much better/proper to handle errors rather then throwing them all the time. The reason for `Promise.reject` is because we have lint rules that you can't throw objects, so we use `Promise.reject` in it's place as Tanstack accepts that in the same manner as throwing an error, then will throw it up into the proper ErrorBoundary

adrian-codecov: Makes sense, I assumed we transitioned into safeParse as we discovered more of zod 👌

src/services/tier/useTier.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const TierSchema = z
+  .object({
+    owner: z.object({
+      plan: z
+        .object({
+          tierName: z.nativeEnum(TierNames),
+        })
+        .nullish(),
+    }),
```
nicholas-codecov, line 21: The `owner` field here could also be `null` so needs to have `nullish()` added on, highly recommend pasting your query into playground and checking the types there, if it doesn't have a `!`, that indicates that the value could be returned as `null`.

src/services/tier/useTier.ts
```diff
@@ -0,0 +1,63 @@
...
+        const parsedRes = TierSchema.safeParse(res?.data)
+
+        if (!parsedRes.success) {
+          return Promise.reject({
+            status: 404,
+            data: null,
+          })
+        }
+
+        return parsedRes.data?.owner?.plan?.tierName
```
nicholas-codecov, line 61: You can't return an undefined value in a query function or else tanstack query will yell at you, so could we some how figure out some clean way of returning something here that isn't undefined but doesn't make things too complicated. Could be an object like:

```js
return {
  pro: true | false
  lite: true | false
  free: true | false
}
```

nicholas-codecov (changes requested):
Sorry missed one thing in the bottom, as well as another quick change to the zod schema

nicholas-codecov (approved):
Nice work with that sweet sweet TypeScript

PR #2314, opened 2023-10-17
Remove segment from frontend

terry-codecov:
# Description
Deleted segment from gazebo. Many deletions!

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


public/index.html
```diff
@@ -48,12 +48,6 @@
     -->
     <title>Codecov</title>
     <% if(process.env.REACT_APP_ENV === "production") { %>
-      <script>
```
adrian-codecov, line 51: Lets gooo

adrian-codecov (approved):
Nitpick: found this mention of segment here https://github.com/codecov/gazebo/blob/main/src/services/tracking/utm.js#L9.

Also approved. Introducing segment was one of the first PRs I did in Gazebo... the nostalgia 👋

PR #2317, opened 2023-10-17
feat: add conditional to exclude static/profiling token

adrian-codecov:
# Description
We want to exclude the static/impact analysis tokens for team tier customers. This change does that.

# Notable Changes
- Create new file + hooks for team specific token list
- Adjusts tests
- Adjust GeneralTab to display either tokens or tokensTeam component
- Add hooks folder for specific hooks

# Screenshots
Without
Screenshot 2023-10-17 at 4 20 38 PM

With
Screenshot 2023-10-17 at 4 20 57 PM

# Legal Boilerplate


src/pages/RepoPage/SettingsTab/tabs/GeneralTab/GeneralTab.jsx
```diff
@@ -1,19 +1,30 @@
...
+import { Tokens, TokensTeam } from './Tokens'
 
 function GeneralTab() {
   const { data } = useRepoSettings()
+  const { provider, owner } = useParams()
+  const { data: tierData } = useTier({ provider, owner })
+
+  const { multipleTiers: isTeamTier } = useFlags({
+    multipleTiers: false,
+  })
```
nicholas-codecov, line 18: Left over from a previous PR, but I think this naming alias may be a bit dangerous as it's not descriptive of the flag and can easily be confused with creating a variable based off of a conditional check.

src/pages/RepoPage/SettingsTab/tabs/GeneralTab/Tokens/TokensTeam/hooks/useRepoSettingsTeam.js
nicholas-codecov, line 1: Do you think it would be worth it to move this hook into `services`?

nicholas-codecov: Also any chance we can convert this hook over to TS?

nicholas-codecov (commented):
Few comments to take a peak at

src/pages/RepoPage/SettingsTab/tabs/GeneralTab/GeneralTab.jsx
adrian-codecov: Yeaah sounds good, I'll change it

src/pages/RepoPage/SettingsTab/tabs/GeneralTab/Tokens/TokensTeam/hooks/useRepoSettingsTeam.js
adrian-codecov: Re 1: We previously discussed as a team on having a hook per page, which is why I decided to place this one here. Would it make more sense to move the `useRepoSettings` hook into a new hooks folder next to the tokens component instead?

Re 2: I opted to leave this one as js as it was practically a copy paste from an existing js one, so it would take me a bit of time to move it into TS, and that's the case for any incoming work that follows this logic. I'm prioritizing completing this team-plan initiative as it's time sensitive, but I'm happy to make a new "convert to TS" ticket for this (and every) ticket as a compromise

nicholas-codecov (approved):
Looking good, ship it ✅

PR #2321, opened 2023-10-18
ref: Update TOS to work for service less users.

RulaKhaled:
# Description
To allow users that haven't yet synced a provider to accept tos, we had to refactor the page to fire the updated serviceless mutation/handle internal user. 

# Notable Changes
- Update mutation hook to fire service less request, and to send the marketing consent. 
- Update API to accept `suppportsServiceless`.
- Update base layout to redirect to `/terms` if terms page meant to be displayed. 
-  TOS to live behind a flag + have it's own route + handle the redirects.
- TOS logic 
- Tests 

# Screenshots

https://github.com/codecov/gazebo/assets/91732700/6aa7111b-08e1-4b55-b27c-1910c95d7679

# Legal Boilerplate


src/layouts/BaseLayout/BaseLayout.jsx
```diff
@@ -36,11 +35,7 @@ const OnboardingOrChildren = ({ children }) => {
   const { isImpersonating } = useImpersonate()
 
   if (showAgreeToTerms && !isFullExperience) {
-    return (
-      <Suspense fallback={null}>
-        <TermsOfService />
-      </Suspense>
-    )
+    return <Redirect to="/terms" />
```
terry-codecov, line 38: I think having a dedicated route and using redirects means the original path the user was trying to access will be lost. Example:

User comes from a PR link, hasn't signed TOS, when they sign it they're sent to the `/:provider/:org` not the PR they clicked on.

https://github.com/codecov/gazebo/pull/2321/files#diff-a8fba335305fed7cd0a638ac77db50888a5c94a8c0115d5cd78a00664904dec2R112


Is this a known trade off we're making?

RulaKhaled: I think it’s fine. we require no synced services to render this page, TOS lived under a component that used queries that serve on providers so it made sense to take it out, otherwise we would have to alter HomePageRedirect
and other components to make this happen

RulaKhaled: Ps. Might be worth it to do the same thing for default org? sync page has it’s own dedicated route as well, I think it’s weird to jump from `/terms` -> `/sync` -> `/` But then again, default org works fine with services.

terry-codecov: I've to raised this to AJ. I (unfortunately) think loosing the navigation path is not good UX and that another approach might be needed here.. But I'll let him weigh the trade offs

RulaKhaled: sure. do you mind including me in the conversation please?

terry-codecov: Yup for sure! I directed him to this issue actually how ever if more slacking happens we'll move to the team channel

RulaKhaled: thank you

src/shared/api/api.js
```diff
@@ -55,8 +55,19 @@ function prefillMethod(method) {
...
+  provider,
+  query,
+  variables = {},
+  signal,
+  supportsServiceless = false,
+}) {
+  let uri = `${config.API_URL}/graphql/${provider}`
+
+  if (supportsServiceless && !provider) {
+    uri = `${config.API_URL}/graphql/`
```
terry-codecov, line 68: Is the `/` needed when the original one doesn't have it?

terry-codecov (approved):
We can finally complete this?! 🙌

src/shared/api/api.js
RulaKhaled: yah the endpoint requires`/` unlike when there are providers passed

PR #2332, opened 2023-10-19
Remove free trial expiry banner from plan page

rohitvinnakota-codecov:
# Description

Currently, the `Your trial ends in`  banner is visible across all tabs. This PR makes it such that it does on render on the `Plan` tab. Two approaches were considered. 

1.) Dismiss the banner no matter where the user is after they click **Upgrade** , even if they are on that page already
2.) Do not show the banner on the Plan page, even if they did not get there by clicking the button

This change implements approach 2, keeping in mind that the sales experience on that page is soon to be upgraded. 

# Notable Changes

I also cleaned up some redundancies in `TrialBanner.spec.tsx` in order to add tests for this behaviour. I'm leaning towards creating issues for any refactoring work in the future but prefer to keep these changes for now. 

# Screenshots

https://github.com/codecov/gazebo/assets/148245014/ed440317-cf79-4ab3-a850-01ef85740b30

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/shared/GlobalTopBanners/TrialBanner/TrialBanner.tsx
+++ b/src/shared/GlobalTopBanners/TrialBanner/TrialBanner.tsx
@@ -1,6 +1,6 @@
 import { differenceInCalendarDays } from 'date-fns'
 import isUndefined from 'lodash/isUndefined'
-import { useParams } from 'react-router-dom'
+import { useLocation, useParams } from 'react-router-dom'
 
 import config from 'config'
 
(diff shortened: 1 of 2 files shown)


src/shared/GlobalTopBanners/TrialBanner/TrialBanner.tsx
```diff
@@ -24,13 +24,18 @@ const determineDateDiff = ({
...
+}
+
 interface UrlParams {
   provider?: string
   owner?: string
 }
 
 const TrialBanner: React.FC = () => {
   const { provider, owner } = useParams<UrlParams>()
+  const pathStartsWithPlan = startsWithPlan(useLocation().pathname)
```
nicholas-codecov, line 38: One small thing, can you pull out `useLocation` here and assign it directly to a variable?

nicholas-codecov (approved):
Nice work, thanks for that quick fix.

PR #2335, opened 2023-10-20
feat: Grab flags in IndirectChangesTable and pass along with request

nicholas-codecov:
# Description

This PR updates the indirect changes table on the commit  detail page to pull the flags from the URL and pass them along with the request to filter out files that are not covered with the selected flags.

Closes codecov/engineering-team#348

# Notable Changes

- Update `IndirectChangesTable` to pull flags from URL and pass along to request
- Update tests


src/pages/CommitDetailPage/subRoute/IndirectChangesTab/IndirectChangesTable/IndirectChangesTable.jsx
```diff
@@ -162,6 +181,18 @@ function IndirectChangesTable() {
   }
 
   if (isEmpty(indirectChangedFiles)) {
+    if (
+      isArray(flags) ||
+      (commit?.compareWithParent?.__typename === 'Comparison' &&
```
rohitvinnakota-codecov, line 186: Consider moving these 2 lines into a helper method that takes in commit info and the `impactedFiles?.__typename` to compare against along lines 177-178. This way you can avoid the nested conditionals and also reduce some repeated code.

nicholas-codecov: I'm happy to pull out the conditional. Regarding the repeated code, I personally disagree with the DRY principles quite a bit most of the time, if you're doing something quite a lot then I feel it's alright to pull stuff out into a helper function. However in this case here where I am only checking this condition twice I think it's quite overkill to pull it out into a little helper function. I also find it a lot easier to read the current implementation because it is quite explicit whereas if it's pulled out into a helper function, you as the developer now have to switch back and forth to understand the implementation of the helper function, causing a slightly increased dev cost.

rohitvinnakota-codecov: I agree with what you're saying  👍 . I did see a similar function [here](https://github.com/codecov/gazebo/pull/2334/files#diff-b7be5031d4672655bb63198a1acc000947fb4af44eb9db1688d01075916888afR196) as well so I was curious. Happy to let you make the call. I would recommend getting an approval from someone else as I'm still getting up to speed with all the logic

src/pages/CommitDetailPage/subRoute/IndirectChangesTab/IndirectChangesTable/IndirectChangesTable.jsx
```diff
@@ -144,6 +162,7 @@ function IndirectChangesTable() {
     commitid: commitSha,
     filters: {
       hasUnintendedChanges: true,
+      flags: flags,
```
adrian-codecov, line 165: Nitpick: `flags` instead of `flags:flags`

src/pages/CommitDetailPage/subRoute/IndirectChangesTab/IndirectChangesTable/IndirectChangesTable.jsx
adrian-codecov: Thanks for the input @rohitvinnakota-codecov!

Yeah probably a bit early to pull out into it's own function although hard to tell in retrospect. If we feel we, conceptually, want to handle flag errors the same way regardless, might be good to make into it's own thing as changing it in the future would only imply changing it once (and potentially forgetting to change it everywhere). I'm all-in for what we think scales best over time

adrian-codecov (approved):

PR #2342, opened 2023-10-23
Migrate TextInput to TypeScript

rohitvinnakota-codecov:
# Description
As per https://github.com/codecov/engineering-team/issues/518, migrates the TextInput component, tests, and stories to Typescript. 
There are still references to `dataMarketing` elsewhere in the code so I converted it directly to a string type in the component class. I checked existing usage of text boxes in the app and Storybook to verify the changes are working as intended. 

# Legal Boilerplate


src/ui/Icon/Icon.tsx
```diff
@@ -13,7 +13,7 @@ const iconComponentCollection = {
 type IconCollection = typeof iconComponentCollection
 type Variant = keyof IconCollection
 
-type OutlineIconCollection = typeof svgOutline
+export type OutlineIconCollection = typeof svgOutline
```
rohitvinnakota-codecov, line 16: Exported this type so I can declare a custom type in `TextInput.tsx` based on the keys of this collection. `TextInput` renders an Icon with an `outline` variant, which type checks against keys only in this collection.

terry-codecov (commented):
Looking much better 🙌

terry-codecov (approved):

PR #2343, opened 2023-10-24
Connect flag selector to flags filter on PR details page

terry-codecov:
# Description
Update impacted files resolver for use pull, connect the flag selector to the api.

# Notable Changes
* Moves to using typename for impacted files states
* AJ is on vacation so I added some basic placeholder copy for the unknown flags state.

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/PullRequestPage/subroute/FilesChangedTab/FilesChanged/FilesChanged.jsx
```diff
@@ -57,14 +54,31 @@ function FilesChangedTab() {
...
+    data?.impactedFiles.length > 0
+  ) {
     return (
       <div className="flex flex-col gap-2">
         <Table />
       </div>
     )
   }
 
+  if (data?.impactedFilesType === 'UnknownFlags') {
```
adrian-codecov, line 68: Have we defined an enum w/ `UknownFlags`? Might be nice if we ever rename what UnknownFlags is

src/pages/PullRequestPage/subroute/FilesChangedTab/FilesChanged/hooks/useImpactedFilesTable.js
```diff
@@ -18,56 +19,73 @@ export const orderingParameter = Object.freeze({
...
-      changeCoverage,
-      hasHeadOrPatchCoverage,
-      headName: impactedFile?.headName,
-      fileName: impactedFile?.fileName,
-      isCriticalFile: impactedFile?.isCriticalFile,
-      pullId: pull?.pullId,
+  const mutatedImpactedFiles = compareWithBase?.impactedFiles?.results?.map(
+    (impactedFile) => {
+      const headCoverage = impactedFile?.headCoverage?.percentCovered
+      const missesCount = impactedFile?.missesCount || 0
```
adrian-codecov, line 39: We can get rid of the || 0 here as we always return a number as per https://github.com/codecov/codecov-api/blob/main/graphql_api/types/impacted_file/impacted_file.py#L126C22-L126C22, which follows https://github.com/codecov/codecov-api/blob/main/services/comparison.py#L893. Probably logic copy-pasted from when it wasn't working 👌

adrian-codecov (approved):
Minor comments but lgtm!

src/pages/PullRequestPage/subroute/FilesChangedTab/FilesChanged/FilesChanged.jsx
terry-codecov: There is not one no, but  I can go ahead and make one, I mostly was leaning on not as it was the only instance using it yet. I dont have strong feelings about it.

PR #2347, opened 2023-10-25
Convert Sparkline to typescript

rohitvinnakota-codecov:
# Description
Converts `Sparkline`, spec, and stories to Typescript. Tested by checking relevant pages in app and storybook runs. This closes https://github.com/codecov/engineering-team/issues/519.

# Legal Boilerplate


src/ui/Sparkline/Sparkline.tsx
```diff
@@ -2,32 +2,47 @@ import { extent } from 'd3-array'
...
-const Sparkline = ({
+interface SparklineData {
+  value: NumberOrNullOrUndefined
+  start: NumberOrNullOrUndefined
+  end: NumberOrNullOrUndefined
+  mode: 'empty' | 'normal'
+}
+
+export interface SparklineProps {
+  datum: any[]
```
rohitvinnakota-codecov, line 21: Prefer to avoid `any` but we need to support custom inputs.

terry-codecov: Hmm, could we pass Sparkline a generic? That way internally it could use that instead of any's

Example:
`select?: (data: any) => NumberOrNullOrUndefined`
could be
`select?: (data: T) => NumberOrNullOrUndefined`

rohitvinnakota-codecov: I did a little bit of digging to see if it was possible. If we pass a generic type to `SparklineProps`, we would have to define either a default type definition or specify a type in the component function. For example, something like `export interface SparklineProps<T = any>` or `const Sparkline: React.FC<SparklineProps<any>>`. One other thing to note is `datum` is an array of potentially different types, and data can be any type. 

We could specify a default type that does not require the use of `any` doing something like `T extends {value:...}` but that would restrict the custom types we are able to support by forcing them to adhere to that type. My feeling is that this is a decent solution, but will require a few changes to other files. 

Another path forward is refactoring the file to keep the same behaviour but make it work better with typescript, which I also think is valid but that could be considered as a separate issue. 

I'm keen to hear your thoughts on these suggestions.

terry-codecov: Ok I took a look myself, you're correct I dont think we can properly use Typescript generics yet... 

For now any is ok, long term I think the option to refactor the code is the best option. Properly use generics is important for this sort of data selection pattern. But we can spin that out into another tech debt ticket. Thanks for the investigation 👍

terry-codecov (approved):

PR #2355, opened 2023-10-27
Convert Modal and BaseModal to TypeScript

rohitvinnakota-codecov:
# Description

This PR converts `Modal`, `BaseModal`, related spec and story jsx files to typescript. This PR closes [#Modal](https://github.com/codecov/engineering-team/issues/489)

# Legal Boilerplate


src/ui/Modal/Modal.spec.tsx
```diff
@@ -62,7 +61,7 @@ describe('Modal', () => {
         />
       )
 
-      await user.click(screen.getByLabelText('Close'))
+      await userEvent.click(screen.getByLabelText('Close'))
```
nicholas-codecov, line 64: Can you bring the `userEvent.setup()` we ran into issues with using it directly previously, it's also the preferred style outlined by [testing-library](https://testing-library.com/docs/user-event/setup).

nicholas-codecov (changes requested):
One thing to take a peak at

src/ui/Modal/Modal.spec.tsx
rohitvinnakota-codecov: I see, thanks for the context.

nicholas-codecov (approved):
Looks all good to me 👍

PR #2356, opened 2023-10-28
feat: replace usePlans to use gql instead of REST

adrian-codecov:
# Description
We're moving away from the REST endpoint to surface plans and use GQL instead. Functionality should be the same + in parity with previous endpoint.

# Notable Changes
- Repurpose `usePlans` to `useAvailablePlans` using gql
- Adjust many tests to ensure this still works

# Legal Boilerplate


src/services/account/useAvailablePlans.ts
```diff
@@ -0,0 +1,77 @@
...
+                baseUnitPrice: z.number(),
+                benefits: z.array(z.string()),
+                billingRate: z.string().nullable(),
+                marketingName: z.string(),
+                monthlyUploadLimit: z.number().nullable(),
+                value: z.string(),
+              })
+              .nullish()
+          )
+          .nullish(),
```
nicholas-codecov, line 23: What is the reason for switching from `nullable()` to `nullish()`?

src/services/account/useAvailablePlans.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface UseAvailablePlansArgs {
+  provider: string
+  owner: string
+}
+
+export const useAvailablePlans = ({
+  provider,
+  owner,
+}: UseAvailablePlansArgs) => {
+  // the plans are very static data
```
nicholas-codecov, line 53: If the plans are very static we could probably set a longer lasting `staleTime` to reduce the amount of requests made overall.

nicholas-codecov (changes requested):
Few comments, but also just wondering if the billing util functions will still work all the same, as the new `findTeamPlan` functions differently iirc

src/services/account/useAvailablePlans.ts
adrian-codecov: Copy paste from an early days hook, good reminder of why we use nullable over nullish 👍

nicholas-codecov (approved):
Thank you thank you, looks all good to me

PR #2359, opened 2023-10-30
feat: change planName to value for planData hook

adrian-codecov:
# Description
We're changing the `planName` resolver value for plan data to `value`. Historically, we've used `value`, and although we ultimately want to use `planName`, maintaining both keys/utils/proptypes/revolving logic with different key names can be hard to follow, specially with those files not using TS atm.

# Notable Changes
- Change `usePlanData` hook to use `value` over `planName`
- Change every use of planName to use value throughout the codebase + tests

# Link to Sample Entry

# Legal Boilerplate


rohitvinnakota-codecov (approved):
👍  consistency is good

PR #2367, opened 2023-11-01
feat: refactor and rename paidplancard

adrian-codecov:
# Description
I'm adding some logic to cleanup the `ProPlanCard`, renamed to `PaidPlanCard`. I initially thought it was going to be more logic but ended up being just a minor refactor after all, this one should be fairly easy.

# Notable Changes
- Renamed `ProPlanCard` to `PaidPlanCard` and everywhere it used it
- Refactored test, included test to use team plan information
- Got rid of the props and added independent hooks called inside that component

# Screenshots
Screenshot 2023-10-31 at 9 45 27 PM

# Link to Sample Entry

# Legal Boilerplate


src/pages/PlanPage/subRoutes/CurrentOrgPlan/CurrentPlanCard/PaidPlanCard/PaidPlanCard.jsx
```diff
@@ -1,38 +1,47 @@
...
+    provider,
+    owner,
+  })
+  const scheduledPhase = accountDetails?.scheduleDetail?.scheduledPhase
+  const plan = planData?.plan
+  const marketingName = plan?.marketingName
+  const benefits = plan?.benefits
+  const value = plan?.value
+  const baseUnitPrice = plan?.baseUnitPrice
+  const seats = plan?.planUserCount
```
nicholas-codecov, line 23: Are you able to provide any default values here if these values are possibly null?

nicholas-codecov (commented):
One small question, and seems like some missing test coverage :eyes:

src/pages/PlanPage/subRoutes/CurrentOrgPlan/CurrentPlanCard/PaidPlanCard/PaidPlanCard.jsx
adrian-codecov: So all of these can't be nullable as per our backend/zod schema, so we should be fine here. 

But otherwise, for a scenario that applies, what default value would work here? It almost feels like if we didn't have all of these that we shouldn't show the card, or have designs reflect that state (but to be clear, they're non-nullable here)

Screenshot 2023-11-01 at 9 35 54 AM

PR #2369, opened 2023-11-01
Alert Texts should show dynamic plan names and upload limits

rohitvinnakota-codecov:
# Description

This PR closes https://github.com/codecov/engineering-team/issues/718. It adjusts the text on the alert banners to show the name of the plan and the upload limit for that account. It makes use of the `monthlyUploadLimit` attribute from `usePlanData` to check for a limit. If that attribute is not defined, it considers that there is no upload limit and defaults to not showing any banners. 

No TSX changes for now in the interest of time.

(Preview with custom data)
Screenshot 2023-11-01 at 4 22 42 PM

# Legal Boilerplate


src/pages/OwnerPage/Header/HeaderBanners/ExceededUploadsAlert/ExceededUploadsAlert.jsx
```diff
@@ -1,19 +1,21 @@
+import PropTypes from 'prop-types'
+
 import A from 'ui/A'
 import Banner from 'ui/Banner'
 import BannerContent from 'ui/Banner/BannerContent'
 import BannerHeading from 'ui/Banner/BannerHeading'
 import Button from 'ui/Button'
 
-const ExceededUploadsAlert = () => (
+const ExceededUploadsAlert = ({ planName, monthlyUploadLimit }) => (
```
adrian-codecov, line 9: How do you feel about using hooks in this component that get the information you need rather than providing those as parameters? We've tried to adopt this notion that components can and should be* standalone, so they can used anywhere you need them without it's parent needing a specific interface, leading to prop-drilling. 

My first instinct is then to be "well, I already have the information on my parent, what's the big deal, plus it it's another request anyway". To the 1st point, this leads to coupling and prop drilling, which we try to avoid. To the 2nd point, this approach doesn't lead to any extra requests/latency on our end because useQuery caches requests with a particular queryKey, so it won't fire N requests for N calls from a component and all of it's children, it would instead fire once and on any subsequent requests check if there's a cached response from that. Wdyt?

If you agree, then my suggestion would be to add the `usePlanData` hook and get the planName/monthlyUploadLimit from there (and adjust the tests and whatnot). This applies to the other component below as well.

src/pages/OwnerPage/Header/HeaderBanners/ReachingUploadLimit/ReachingUploadLimit.jsx
```diff
@@ -1,19 +1,21 @@
+import PropTypes from 'prop-types'
+
 import A from 'ui/A'
 import Banner from 'ui/Banner'
 import BannerContent from 'ui/Banner/BannerContent'
 import BannerHeading from 'ui/Banner/BannerHeading'
 import Button from 'ui/Button'
 
-const ExceededUploadsAlert = () => (
+const ExceededUploadsAlert = ({ planName, monthlyUploadLimit }) => (
```
adrian-codecov, line 9: Also this has a different component name than the file, I believe we want to call this component `ReachingUploadLimit` and probably missed it, could you please change it?

src/pages/OwnerPage/Header/HeaderBanners/HeaderBanners.jsx
```diff
@@ -4,40 +4,48 @@ import { useParams } from 'react-router-dom'
...
-    numberOfUploads >= REACHING_UPLOAD_LIMIT
-
+  // If monthlyUploadLimit is not defined, we consider the account can have an
+  // unlimited amount of uploads
+  const monthlyUploadLimit = planData?.plan?.monthlyUploadLimit
+  const isUploadsExceeded = monthlyUploadLimit
+    ? numberOfUploads >= monthlyUploadLimit
+    : false
+  const isUploadsReachingLimit = monthlyUploadLimit
+    ? !isUploadsExceeded && numberOfUploads >= 0.9 * monthlyUploadLimit
```
adrian-codecov, line 25: Ah smart, I like the 90% trick, I think this is a fair assumption and we can tweak it later. In fact, could you ask product how one determines approaching upload limit? It could be a percentage of your total, or a flat value. I prefer percentages but might be worth confirming w/ the law 😂

src/pages/OwnerPage/Header/HeaderBanners/HeaderBanners.jsx
```diff
@@ -4,40 +4,48 @@ import { useParams } from 'react-router-dom'
...
-    numberOfUploads < MAX_UPLOADS_NUMBER &&
-    numberOfUploads >= REACHING_UPLOAD_LIMIT
-
+  // If monthlyUploadLimit is not defined, we consider the account can have an
+  // unlimited amount of uploads
+  const monthlyUploadLimit = planData?.plan?.monthlyUploadLimit
+  const isUploadsExceeded = monthlyUploadLimit
+    ? numberOfUploads >= monthlyUploadLimit
+    : false
+  const isUploadsReachingLimit = monthlyUploadLimit
```
adrian-codecov, line 24: Out of curiosity, does this read weird? `isUploadsReachingLimit`? It sounds more like a `isApproachingUploadLimit` to me

src/pages/OwnerPage/Header/HeaderBanners/ExceededUploadsAlert/ExceededUploadsAlert.jsx
rohitvinnakota-codecov: That makes sense to me. Thanks for the explanation.

src/pages/OwnerPage/Header/HeaderBanners/HeaderBanners.jsx
rohitvinnakota-codecov: Changed name

src/pages/OwnerPage/Header/HeaderBanners/HeaderBanners.jsx
rohitvinnakota-codecov: Confirmed with product

adrian-codecov (approved):

PR #2372, opened 2023-11-02
fix: Drop Transactions from Blocked User Agents

nicholas-codecov:
# Description

Just some small tweaks to our Sentry config to clean up our issues as well as stopping the use of deprecated options.

# Notable Changes

- Create function to check for blocked user agents
- Set Sentry Replay rates

### DIFF
--- a/src/sentry.ts
+++ b/src/sentry.ts
@@ -51,6 +51,14 @@ const deClutterConfig = {
 
 export const SentryRoute = Sentry.withSentryRouting(Route)
 
+const checkForBlockedUserAgents = () => {
+  const userAgents = ['Bytespider']
+
+  return userAgents.some((agent) =>
+    window?.navigator?.userAgent.includes(agent)
+  )
+}
+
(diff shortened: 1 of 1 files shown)


src/sentry.ts
```diff
@@ -51,6 +51,14 @@ const deClutterConfig = {
 
 export const SentryRoute = Sentry.withSentryRouting(Route)
 
+const checkForBlockedUserAgents = () => {
```
adrian-codecov, line 54: Funny! When I first reviewed this I was going to suggest doing an approach like this to add a list of agents we wanted blocked, but then I thought it was a bit overkill, so funny we arrived to this on our own

adrian-codecov (approved):

src/sentry.ts
nicholas-codecov: I pulled it out cause we're using it in two locations now ... so I thought it best they stay in sync with one another

PR #2377, opened 2023-11-03
Enable/disable tabs and flag select based on team plan and private repo status

terry-codecov:
# Description
Updates the tabs and flag selector based on plan type and repo status

Addresses:
https://github.com/codecov/engineering-team/issues/624
https://github.com/codecov/engineering-team/issues/686

# Code Example

# Notable Changes
* Added private repo resolve to the pull tabs query
* The tests are long/complicated due to there being two related feature flags 👎 

# Screenshots
Team Plan Private repo
Screenshot 2023-11-03 at 10 28 34 AM
Non Team plan / public
Screenshot 2023-11-03 at 10 29 26 AM

# Link to Sample Entry

# Legal Boilerplate


src/pages/PullRequestPage/PullRequestPageTabs/PullRequestPageTabs.jsx
```diff
@@ -18,13 +19,16 @@ function PullRequestPageTabs() {
     indirectChangesCount,
     directChangedFilesCount,
     commitsCount,
+    isPrivateRepo,
```
adrian-codecov, line 22: How do you feel about using `useRepoSettingsTeam` and getting the private variable from there, instead of adding it to this hook? It feels a little weird that we're getting that from a hook called `useTabsCounts` and that the only reason we're fetching it from there is cause we already had access to the repository resolver so adding the key is inconvenient. On the flip side we had to change 6 files total to make this happen, while only the change affects this file on the component side. I mainly feel the hook starts to drift from being what it's meant to do, returning counts 😬

src/pages/PullRequestPage/PullRequestPageTabs/PullRequestPageTabs.spec.jsx
```diff
@@ -82,22 +107,40 @@ afterAll(() => {
...
       graphql.query('GetCommits', (req, res, ctx) =>
         res(ctx.status(200), ctx.data(mockCommits))
       ),
       graphql.query('BackfillFlagMemberships', (req, res, ctx) =>
         res(ctx.status(200), ctx.data({}))
+      ),
+      graphql.query('OwnerTier', (req, res, ctx) =>
+        res(
+          ctx.status(200),
+          ctx.data({ owner: { plan: { tierName: plan.toLowerCase() } } })
```
adrian-codecov, line 142: Nitpick: could we rename `plan` to `tier` ?

adrian-codecov (commented):
Mainly the isPrivate comment, otherwise lgtm!

src/pages/PullRequestPage/PullRequestPageTabs/PullRequestPageTabs.jsx
terry-codecov: It's definitely drifting, I was mostly trying to avoid adding another network request and this was all ready being used in the code, that being said it's definitely a short cut. To be honest I wasnt aware of the `useRepoSettingsTeam` hook, I think it makes sense to refactor to use that one instead.

src/pages/PullRequestPage/PullRequestPageTabs/PullRequestPageTabs.spec.jsx
terry-codecov: I could at the query level, but it is called plan in the actual resolver so maybe we should also update it there and deprecate the plan resolver?

adrian-codecov: Ah I should have been bit more specific, meant `plan: { tierName: tier.toLowerCase()}}`, cause tierName is a tier, not a plan itself

terry-codecov: Oh I see! Just to make sure I understand I need to replace the tiers with plans, so it needs to be

`plan: { planName: plan.toLowerCase()}}` ?

src/pages/PullRequestPage/PullRequestPageTabs/hooks/useTabsCounts.spec.js
```diff
@@ -36,7 +36,6 @@ const mockPullData = {
     isCurrentUserPartOfOrg: true,
     repository: {
       __typename: 'Repository',
-      private: true,
```
adrian-codecov, line 39: Huh, not sure why this registered as a change but 🤷

adrian-codecov (approved):

PR #2381, opened 2023-11-03
Analytics Page: Only Surface Public Repos for Team Tier

rohitvinnakota-codecov:
# Description

On the analytics page, if the user is in team tier, only show public repos. This change makes use of the newly added `isPublic` field from https://github.com/codecov/codecov-api/pull/224. This PR closes https://github.com/codecov/engineering-team/issues/723

Behaviour of isPublic
```
If isPublic is null it will return both public and private repos
If isPublic=true it will only return public repos
If isPublic=false it will only return private repos

```

# Legal Boilerplate


src/shared/ListRepo/ReposTable/ReposTable.jsx
```diff
@@ -134,6 +136,9 @@ function ReposTable({ searchValue, owner, sortItem, filterValues = [] }) {
   const { data: ownerData } = useOwner({
     username: owner || userData?.user?.username,
   })
+  const { provider } = useParams()
+  const { data: tierName } = useTier({ provider, owner })
+  const isPublic = tierName === TierNames.TEAM ? true : null
```
JerrySentry, line 141: - Which page does this change correspond to on the UI?
- Does this mean once a basic user upgrades to team plan all their private repos will just disappear from this page, is it better to show all the repos and mark the private ones as "unavailable with team plan" or something?

rohitvinnakota-codecov: 1. This is only on the analytics page. The `ReposTable` component is rendered there. It is used in another view but is overwritten by a `ReposTableTeam` component if the user is on a team plan in [that component](https://github.com/codecov/gazebo/blob/main/src/shared/ListRepo/ListRepo.jsx). 

2. I agree. I was following spec here and I am open to input from @adrian-codecov and product. It is interesting that private repos are no longer visible if the user "_upgrades_" to a team plan.

adrian-codecov: Yeap, expected logic is when you are team customer, you only show the public ones. I also feel it's a bit of a weird flow, specially cause as a basic customer you can see your public and private repos as of today, but I believe that's yet to change in the future (so the basic plan would only see public as well). It currently does feel like a weird experience, but this is the ask for the team plan atm.

adrian-codecov, line 141: I would rename this variable a little bit, it is weird to equate if something is public to a tiername 😓. Maybe like `shouldDisplayPublicReposOnly` or something like that?

adrian-codecov (approved):
I'd just change the variable name, otherwise gtg!

PR #2388, opened 2023-11-07
fix: Remove default org feature flag and adjust `useUserAccessGate`

RulaKhaled:
# Description
 While removing default org feature flag i noticed that we are handling missing user weirdly, in a way that if the request fails, we still update the fields e.g showDefaultOrg which was causing some weird behaviors in use user access gate hook. 

# Notable Changes
-> Fix missing user and handle update values if user data exists 
-> Decouple internal user and user 
-> Remove default org flag 
-> Update tests 

# Screenshots
nothing visual changed 

closes: https://github.com/codecov/engineering-team/issues/483 

# Legal Boilerplate


nicholas-codecov (approved):
Looks awesome, thank you thank you 🎉

PR #2389, opened 2023-11-07
ref: Ensure All Queries Use Object Syntax

nicholas-codecov:
# Description

This PR updates a handful of queries to use the object syntax in preparation for the migration to TanStack v5.

Setup work for codecov/engineering-team#678

# Notable Changes

- Update a handful of hooks to object syntax
- Update `MemberTable` and how it invalidates queries


src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.jsx
```diff
@@ -96,12 +96,7 @@ function MemberTable() {
     onSuccess: () => {
       queryClient.invalidateQueries(['SelfHostedSettings'])
       queryClient.invalidateQueries(['Seats'])
-      queryClient.invalidateQueries([
-        'SelfHostedUserList',
-        params?.activated,
-        params?.search,
-        params?.isAdmin,
-      ])
+      queryClient.invalidateQueries(['SelfHostedUserList'])
```
adrian-codecov, line 99: You don't need the whole sets of params to invalidate the query right? Like if you had a query key with

`['SelfHostedUserList', 'gh']`
`['SelfHostedUserList', 'gh', 'codecov']`
`['SelfHostedUserList']`

That prompt will invalidate any of them right?

adrian-codecov (approved):

src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.jsx
nicholas-codecov: I believe so, they're not 100% explicit, in the docs they say:

> The invalidateQueries method can be used to invalidate and refetch single or multiple queries in the cache based on their query keys or any other functionally accessible property/state of the query. By default, all matching queries are immediately marked as invalid and active queries are refetched in the background.

PR #2391, opened 2023-11-09
fix: Redirect on Flags Tab for Public Repos

nicholas-codecov:
# Description

This PR updates the `FlagsTab` to only redirect while an org is on a team plan and the repo is also public.

Closes codecov/engineering-team#781

# Notable Changes

- Update `FlagsTab.jsx`
- Update tests

### DIFF
--- a/src/pages/RepoPage/FlagsTab/FlagsTab.spec.jsx
+++ b/src/pages/RepoPage/FlagsTab/FlagsTab.spec.jsx
@@ -36,6 +36,21 @@ const flagsData = [
   },
 ]
 
+const mockRepoSettings = (isPrivate = false) => ({
+  owner: {
+    repository: {
+      defaultBranch: 'master',
+      private: isPrivate,
+      uploadToken: 'token',
+      graphToken: 'token',
+      yaml: 'yaml',
+      bot: {
+        username: 'test',
+      },
+    },
+  },
+})
+
 const server = setupServer()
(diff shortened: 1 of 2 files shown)


src/pages/RepoPage/FlagsTab/FlagsTab.spec.jsx
```diff
@@ -84,21 +108,43 @@ describe('Flags Tab', () => {
...
   }
 
   describe('when user has a team tier', () => {
-    beforeEach(() => {
-      setup({ tierValue: TierNames.TEAM })
+    describe('the repo is public', () => {
+      it('renders the flags tab', async () => {
+        setup({
+          tierValue: TierNames.TEAM,
+          isPrivate: false,
```
rohitvinnakota-codecov, line 123: mega nit: do you need this?

rohitvinnakota-codecov (approved):

src/pages/RepoPage/FlagsTab/FlagsTab.spec.jsx
nicholas-codecov: So because this is testing that the repo is not private I like to be entirely explicit in the setup for things that are related to the test. This means that you don't have to go up and check the setup function and the arguments default value to see if it aligns with what's being tested.

If one day for whatever reason we switch the default value then this test will also still continue to work properly.

PR #2399, opened 2023-11-10
Update the rest of the pull request tabs to filter by flags

terry-codecov:
# Description
Fixed not all tabs receiving flags filter. Adds filtering to Indirect Files tab, File Explorer, 

# Code Example

# Notable Changes
- Refactored some old tests to use the newer style. Setup returns mock functions instead of a "global" mock and not mocking location / api services.
- Updated the flag tab even though no api support yet so it shares the query cache with other tabs.

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/PullRequestPage/subroute/IndirectChangesTab/IndirectChangedFiles/hooks/useIndirectChangedFilesTable.spec.js
```diff
@@ -1,122 +1,135 @@
...
+  server.listen()
+  console.error = () => {}
+})
+afterEach(() => {
+  queryClient.clear()
+  server.resetHandlers()
+})
+afterAll(() => server.close())
+
 describe('useRepoContentsTable', () => {
```
nicholas-codecov, line 83: nit: this isn't the function that's being tested here, seems to be in a couple of other places in here 👀

src/pages/PullRequestPage/subroute/FileExplorer/FileExplorer.jsx
```diff
@@ -49,7 +58,7 @@ function FileExplorer() {
       />
       {isLoading && <Loader />}
       {data?.length === 0 && !isLoading && (
-        <MissingFileData isSearching={isSearching} />
+        <MissingFileData isSearching={isSearching} hasFlagsSelected={!!flags} />
```
nicholas-codecov, line 61: Hrrm, thoughts on swapping this to `flags?.length > 0` currently it will return `true` if the list exists and nothing is selected.

nicholas-codecov (commented):
Few things to take a peak at, as well seems like a test is failing

src/pages/PullRequestPage/subroute/IndirectChangesTab/IndirectChangedFiles/hooks/useIndirectChangedFilesTable.spec.js
terry-codecov: Might be an old typo that slipped though in previous PRs, I could give it the once over. Do you have the other files you noticed?

nicholas-codecov: It's all in the same file

nicholas-codecov (approved):
Thanks for fixing up that erroneous test name!

Looks all good to me, ship it! 🚀 🌔

PR #2402, opened 2023-11-14
[Client] Add logic to return measurements data based on specifying public and private repo

rohitvinnakota-codecov:
# Description

Make the line chart on the Analytics page filter on only public repos if the user is on a team plan. This makes use of the new `isPublic` parameter on the measurements GQL field as added in https://github.com/codecov/engineering-team/issues/786. This PR closes https://github.com/codecov/engineering-team/issues/787.

# Legal Boilerplate

### DIFF
--- a/src/pages/AnalyticsPage/Chart/useCoverage.spec.js
+++ b/src/pages/AnalyticsPage/Chart/useCoverage.spec.js
@@ -2,11 +2,17 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { renderHook, waitFor } from '@testing-library/react'
 import { graphql } from 'msw'
 import { setupServer } from 'msw/node'
-import { MemoryRouter, Route } from 'react-router-dom'
+import { MemoryRouter, Route , useParams } from 'react-router-dom'
+
+import { TierNames } from 'services/tier'
 
 import { useCoverage } from './useCoverage'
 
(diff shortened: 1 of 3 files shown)


src/pages/AnalyticsPage/Chart/useCoverage.spec.js
```diff
@@ -2,11 +2,17 @@ import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
...
 import { setupServer } from 'msw/node'
-import { MemoryRouter, Route } from 'react-router-dom'
+import { MemoryRouter, Route , useParams } from 'react-router-dom'
+
+import { TierNames } from 'services/tier'
 
 import { useCoverage } from './useCoverage'
 
 jest.mock('services/charts')
+jest.mock('react-router-dom', () => ({
```
adrian-codecov, line 12: This is an antipattern we used to do back in the day we're trying to replace. Effectively this is mocking useParams so that jest doesn't complain, but we can also rightfully provide it to emulate the router, take a look at this test for example, https://github.com/codecov/gazebo/blob/main/src/services/trial/useStartTrial.spec.tsx#L23. Could you get rid of this mock and instead provide the route parameters, seems like owner and provider, to the request instead? Lmk if you have a question

adrian-codecov (approved):

PR #2405, opened 2023-11-15
fix: Move team special offer out of plan default offer

RulaKhaled:
# Description
Problems in new team plan cancel offer:
- it is tied to a discount 
- it shows for team plan users 

# Notable Changes
- Took out the team plan cancel offer out of the default offer and displayed the expected team plan card in the top level page, it is tied to the existence of the team plan and only if user is not on team plan. 
- New tests 
- Typescript refactors 

# Screenshots
nothing visual changed 

# Legal Boilerplate


src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/TeamPlanSpecialOffer/TeamPlanSpecialOffer.tsx
```diff
@@ -0,0 +1,45 @@
...
+      },
+    ])
+  }, [setCrumbs])
+
+  return (
+    <div className="flex w-3/5 flex-col gap-4">
+      <div>
+        <h2 className="text-lg font-semibold">Alternative plan offer</h2>
+        <p>
+          We&apos;d like to introduce you to our lighter alternative - the Team
```
adrian-codecov, line 26: Not sure if you saw in the bug ticket, we want to adjust this copy a bit too, see 2) here https://github.com/codecov/engineering-team/issues/790

src/pages/PlanPage/subRoutes/CancelPlanPage/CancelPlanPage.tsx
```diff
@@ -37,11 +56,16 @@ function CancelPlanPage() {
 
   const discountNotApplied =
     !accountDetailsData?.subscriptionDetail?.customer?.discount
-  const showDiscount =
+  const showSpecialOffer =
     discountNotApplied && isMonthlyPlan(accountDetailsData?.plan?.value)
+  const showTeamSpecialOffer =
+    multipleTiers &&
+    shouldDisplayTeamCard({ plans }) &&
+    !isTeamPlan(accountDetailsData?.plan?.value)
```
adrian-codecov, line 64: Would this work if we had enterprise/basic/gh marketplace (`users`) plans here?

I wonder if we could do something instead of `!isTeamPlan` to `isProPlan` and do something like this


``` javascript
export function isProPlan(plan) {
  if (isString(plan)) {
    return !isUndefined(isSentryPlan(plan)) || !isUndefined(isCodecovProPlan(plan))
  }
  return false
}

```
 (and we'd also need to make the isCodecovProPlan method)

src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/TeamPlanSpecialOffer/TeamPlanSpecialOffer.tsx
RulaKhaled: oops

src/pages/PlanPage/subRoutes/CancelPlanPage/CancelPlanPage.tsx
RulaKhaled: so plan should be showing for:
users-pr-inappm
users-pr-inappy
users-sentrym
users-sentryy

any other plans?

adrian-codecov: For those yup!

src/shared/utils/billing.js
```diff
@@ -93,6 +93,20 @@ export function isSentryPlan(plan) {
...
 }
 
+export function isCodecovProPlan(plan) {
+  if (isString(plan)) {
+    return plan === Plans.USERS_PR_INAPPM || plan === Plans.USERS_PR_INAPPY
+  }
+  return false
+}
+
+export function isProPlan(plan) {
```
adrian-codecov, line 103: nitpick: these new methods need tests

adrian-codecov (approved):

PR #2411, opened 2023-11-17
Auto logout users if their github session expires

rohitvinnakota-codecov:
# Description

This PR closes https://github.com/codecov/engineering-team/issues/642. If the user has a session expiry set by Github Oauth, they will be asked to sign in again once they reach the expiry time. 

Session expiry data is sent via codecov-api as per this [change](https://github.com/codecov/codecov-api/pull/237/files)

When the session expiry is reached, auto logout the user and display a banner indicating their session expired. 

A couple of points

- There is a possibility of adding a warning 15 mins before expiry as a separate modal.
- There might also be a way to auto-refresh the user's token if they are still browsing

Both of these are additional feature adds. The plan for now is to monitor customer feedback and check back on adding those features. 

# Screenshots
Screenshot 2023-11-20 at 1 59 09 PM

# Link to Sample Entry

# Legal Boilerplate


src/ui/SessionExpiryModal/SessionExpiryModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+    const setupSessionIntervalCheck = (sessionExpiryTime: Date) => {
+      checkSession(sessionExpiryTime)
+      return setInterval(
+        () => checkSession(sessionExpiryTime),
+        ONE_MINUTE_MILLIS
+      )
+    }
+
+    let intervalId: NodeJS.Timeout
```
nicholas-codecov, line 41: This will be running in the browser so it shouldn't use a node related type. Timeout interval ids are just numbers so you should be able to just set it as a type `number`

src/ui/SessionExpiryModal/SessionExpiryModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+const SessionExpiryModal: React.FC = () => {
+  const [showModal, setShowModal] = useState(false)
+  const sessionExpiryTimeString = Cookies.get('session_expiry')
+
+  const checkSession = (sessionExpiryTime: Date) => {
+    const currentTime = new Date()
+    const timeLeft = sessionExpiryTime.getTime() - currentTime.getTime()
+    if (timeLeft <= TWO_MINUTES_MILLIS) {
+      setShowModal(true)
+    }
```
nicholas-codecov, line 20: Can you add in the conditional check `setShowModal`'s current state, this will block any potential infinite re-renders.

src/ui/SessionExpiryModal/SessionExpiryModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
```
nicholas-codecov, line 21: Can you wrap this function in a `useCallback` so that it's memoized between re-renders

src/ui/SessionExpiryModal/SessionExpiryModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+    const delayBeforeStart =
+      timeLeft > THIRTY_MINUTES_MILLIS ? timeLeft - THIRTY_MINUTES_MILLIS : 0
+
+    const setupSessionIntervalCheck = (sessionExpiryTime: Date) => {
+      checkSession(sessionExpiryTime)
+      return setInterval(
+        () => checkSession(sessionExpiryTime),
+        ONE_MINUTE_MILLIS
+      )
+    }
```
nicholas-codecov, line 39: Can this function be moved outside of the `useEffect`, as well as wrapping it in a `useCallback` to keep it memoized in between re-renders.

src/ui/SessionExpiryModal/SessionExpiryModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+        <Button
+          variant="primary"
+          hook=""
+          disabled={false}
+          to={{ pageName: 'signOut' }}
+        >
+          Please log in again to continue using Codecov
+        </Button>
+      }
+    />
```
nicholas-codecov, line 73: Just wondering if you have consulted design on the modal? There's a few changes on the design I think could optimize it a bit more.

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
```diff
@@ -0,0 +1,75 @@
...
+    let intervalId: number
+    const delayBeforeStart = getCheckDelay(sessionExpiryTime)
+    const timeoutId = setTimeout(() => {
+      intervalId = setupSessionIntervalCheck(sessionExpiryTime)
+      checkSession(sessionExpiryTime)
+    }, delayBeforeStart)
+
+    return () => {
+      clearTimeout(timeoutId)
+      clearInterval(intervalId)
```
rohitvinnakota-codecov, line 59: Not really sure what's happening with some of these annotations as I'm fairly sure I've ensured these lines have been covered in the spec file.

src/ui/TopBanner/TopBanner.tsx
```diff
@@ -158,7 +162,7 @@ const TopBannerRoot: React.FC<React.PropsWithChildren<TopBannerProps>> = ({
 
 TopBannerRoot.propTypes = {
   variant: PropTypes.oneOf(['default', 'warning']),
-  localStorageKey: PropTypes.string.isRequired,
+  localStorageKey: PropTypes.string,
```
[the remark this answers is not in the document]
nicholas-codecov: You can actually just remove these `propTypes` as they're no longer required for TypeScript files as we can just use the types themselves.

src/ui/TopBanner/TopBanner.tsx
```diff
@@ -26,7 +26,7 @@ type Variants = keyof typeof variants
 
 const topBannerContext = z.object({
   variant: z.union([z.literal('default'), z.literal('warning')]),
-  localStorageKey: z.string(),
+  localStorageKey: z.string().nullish(),
```
nicholas-codecov, line 29: If it can only be set to a `string` or `undefined`, can this be swapped from `nullish` to `optional`?

src/ui/TopBanner/TopBanner.tsx
```diff
@@ -115,19 +119,19 @@ const IconSymbol: React.FC = () => {
...
 interface TopBannerProps {
   variant?: Variants
-  localStorageKey: string
+  localStorageKey?: string
 }
 
 const TopBannerRoot: React.FC<React.PropsWithChildren<TopBannerProps>> = ({
   variant = 'default',
-  localStorageKey,
+  localStorageKey = null,
```
nicholas-codecov, line 127: This default value of `null` conflicts with the types defined in the props interface. Adding in `?` makes the field either it's value in this case a `string` or `undefined`, so you can just skip assigning a default value here.

src/layouts/Header/Dropdown.spec.jsx
```diff
@@ -91,6 +96,31 @@ describe('Dropdown', () => {
...
       })
 
+      it('removes session expiry tracking key on sign out', async () => {
+        const { user } = setup()
+
+        jest.spyOn(console, 'error').mockImplementation()
+        const mockRemoveItem = jest.spyOn(
+          window.localStorage.__proto__,
+          'removeItem'
+        )
```
nicholas-codecov, line 106: Can you move this mock into the setup function, in case any new tests in the future they may end up accidentally calling it are already mocked out.

src/layouts/LoginLayout/LoginLayout.spec.tsx
```diff
@@ -85,4 +85,17 @@ describe('LoginLayout', () => {
       expect(child).toBeInTheDocument()
     })
   })
+
+  describe('when session is expired', () => {
+    it('renders the expiry banner', async () => {
+      setup()
+      jest
+        .spyOn(window.localStorage.__proto__, 'getItem')
+        .mockReturnValue('true')
```
nicholas-codecov, line 94: Same thing with this mock here, can you quickly move it into the `setup` function.

src/ui/SessionExpiryTracker/SessionExpiryTracker.spec.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+  it('shows expired modal only 2 minutes before expiry or later', async () => {
+    const expiryTime = new Date()
+    expiryTime.setMinutes(expiryTime.getMinutes() + 15)
+    Cookies.get = jest.fn().mockImplementation(() => expiryTime.toString())
+    const mockRemoveItem = jest.spyOn(
+      window.localStorage.__proto__,
+      'removeItem'
+    )
+    const mockSetItem = jest.spyOn(window.localStorage.__proto__, 'setItem')
```
nicholas-codecov, line 35: Can you move this mocks into a setup function.

src/ui/SessionExpiryTracker/SessionExpiryTracker.spec.tsx
nicholas-codecov, line 1: Can you move the mocks in this file into a re-usable `setup` function

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
nicholas-codecov: Hrrm, sometimes this is an issue with GitHub caching, but looking in Codecov, it does say these lines are missing 🤔

nicholas-codecov: Okay there seems to be an issue with our test splitting, not including all of the UI files

nicholas-codecov: I've pushed a fix for this in #2418

nicholas-codecov (approved):
Looks all good to me 👍

katia-sentry:
Working on improvements to this upcoming sprint https://github.com/codecov/engineering-team/issues/1892

PR #2414, opened 2023-11-21
Convert some more of pull request details page to TS

terry-codecov:
# Description
Not much to say, just converting over more JS files. Looks like #145 is actually API work so there will be another gazebo follow up pr passing that setting when it exists.

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/PullRequestPage/subroute/CommitsTab/shared/CIStatus/CIStatus.tsx
```diff
@@ -29,9 +32,15 @@ Status.propTypes = {
...
-function CIStatus({ commitid, coverage, ciPassed }) {
+interface CIStatusProps {
+  commitid?: string
+  coverage?: number
+  ciPassed?: boolean | null
+}
+function CIStatus({ commitid, coverage, ciPassed = false }: CIStatusProps) {
   return (
     <div className="flex-1 flex-col items-end lg:flex">
+      {/* @ts-ignore */}
```
RulaKhaled, line 43: we can just do `hook="commitid-link" isExternal={false}` instead of ignoring ts warnings

RulaKhaled (approved):
lgtm, left a comment

PR #2416, opened 2023-11-22
Create team plan query for pull request view

rohitvinnakota-codecov:
# Description

This issue closes https://github.com/codecov/engineering-team/issues/774. There are no visible changes but it fetches less GQL fields if the user is on the team plan.  I went with modifying the query directly as an approach instead of creating a new hook since we cannot conditionally call hooks, so the logic was pushed to the query itself. Users under the team plan cannot view components, flags, or indirect files changed so those are skipped if they're on that plan. 

# Legal Boilerplate


nicholas-codecov (approved):
Looks all good to me 👍

PR #2422, opened 2023-11-27
feat: Create a new hook for the components selector hook 

RulaKhaled:
# Description
We need to fetch name of components to display for the components selector

# Notable Changes

- Create new hook 
- Tests

# Legal Boilerplate

### DIFF
--- a/src/services/pull/usePullComponents.tsx
+++ b/src/services/pull/usePullComponents.tsx
@@ -30,8 +30,8 @@ const query = `
         ... on Repository {
           pull(id: $pullId) {
             compareWithBase {
+              __typename
               ... on Comparison {
-                __typename
                 componentComparisons(filters: $filters) {
                   name
                 }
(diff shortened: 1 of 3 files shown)


src/services/pull/usePullComponents.tsx
```diff
@@ -0,0 +1,207 @@
...
+    $filters: ComponentsFilters
+  ) {
+    owner(username: $owner) {
+      repository(name: $repo) {
+        __typename
+        ... on Repository {
+          pull(id: $pullId) {
+            compareWithBase {
+              ... on Comparison {
+                __typename
```
nicholas-codecov, line 34: Because the typename is under `... on Comparison` I don't believe it will show up for the other types, I would lift it up one level just in case.

nicholas-codecov (commented):
Just one quick thing to take a peak at

PR #2424, opened 2023-11-27
Adjust Uploads Card to Not Display Flag Related Data for team plan

rohitvinnakota-codecov:
# Description

This PR closes https://github.com/codecov/engineering-team/issues/815. 

This is a temporary solution to prevent fetching any flag related data in the uploads pane for team tier users. There are currently no team tier users with flags configured so this will be an invisible change. We will need robust handling on the backend as well as proper Pro -> Team downgrade logic in the future. 

**Before**

Screenshot 2023-12-01 at 12 46 49 PM

**After**

Screenshot 2023-12-01 at 12 47 09 PM

# Legal Boilerplate


adrian-codecov (commented):
Lgtm, but do you have a screenshot/visual of it working with local/staging data? Just want to make sure there's nothing we accidentally miss through code only

src/services/commit/useCommit.spec.tsx
```diff
@@ -135,6 +135,70 @@ const dataReturned = {
...
+              node: {
+                id: 0,
+                state: 'PROCESSED',
+                provider: 'travis',
+                createdAt: '2020-08-25T16:36:19.55947400:00',
+                updatedAt: '2020-08-25T16:36:19.67986800:00',
+                downloadUrl:
+                  '/api/gh/febg/repo-test/download/build?path=v4/raw/2020-08-25/F84D6D9A7F883055E40E3B380280BC44/f00162848a3cebc0728d915763c2fd9e92132408/30582d33-de37-4272-ad50-c4dc805802fb.txt',
+                ciUrl: 'https://travis-ci.com/febg/repo-test/jobs/721065746',
+                uploadType: 'UPLOADED',
```
adrian-codecov, line 169: Another note I just realized, if the session comes with uploadType `carryforward`, will we show that session altogether as well? I forget specifically what we want to do in the interim

rohitvinnakota-codecov: Yeah it will be shown(without any mention of flags in the uploads pane). However there are no customers in the team plan with that configuration. 

When we implement more foundational changes, I expect fetching those to be skipped at the API level.

adrian-codecov (approved):

PR #2426, opened 2023-11-27
feat: Create PR selector components

RulaKhaled:
# Description
Create the new components selector, this shall be used in most of PR tabs.  

Ticket: https://github.com/codecov/engineering-team/issues/841

# Notable Changes
- Create the new selector 
- New tests

# Screenshots
It's not placed anywhere yet but here's how it'll look like:
https://github.com/codecov/gazebo/assets/91732700/3d3113f3-7fa4-4be6-916a-230ac83f992a

# Legal Boilerplate


src/pages/PullRequestPage/subroute/ComponentsSelector/ComponentsSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+    filters: {
+      components: componentSearch ? [componentSearch] : undefined,
+    },
+    options: {
+      suspense: false,
+    },
+  })
+
+  const components =
+    data?.pull?.compareWithBase.__typename === 'Comparison'
```
rohitvinnakota-codecov, line 35: Curious what this means. Do components only exist in this scenario?

src/pages/PullRequestPage/subroute/ComponentsSelector/ComponentsSelector.tsx
```diff
@@ -0,0 +1,89 @@
...
+      : []
+
+  if (
+    (!components?.length && !isLoading && !componentSearch) ||
+    !componentsSelectFlag
+  ) {
+    return null
+  }
+
+  const componentsNames = new Set()
```
rohitvinnakota-codecov, line 46: I think you can do 
`const componentsNames = new Set(params?.components)` and avoid the ts error

rohitvinnakota-codecov (approved):

src/pages/PullRequestPage/subroute/ComponentsSelector/ComponentsSelector.tsx
RulaKhaled: The ts error is coming from the fact that `useLocation` params is not typed, so we would still see the error anywhere we use params :cry:

src/pages/PullRequestPage/subroute/ComponentsSelector/ComponentsSelector.tsx
RulaKhaled: Yes! We only have results if compareWithBase is of comparison type, we have other union types surfaced as errors like `MissingBaseCommit` and `MissingHeadCommit` but these are already handled in the hook level: https://github.com/codecov/gazebo/blob/main/src/services/pull/usePullComponents.tsx

More on union types [here](https://graphql.org/learn/schema/#union-types)

PR #2429, opened 2023-11-29
feat add ProController component to the 2nd form component

adrian-codecov:
# Description
We're adding the `ProController` isolated component that handles form data. This is to improve the general flow for the UpgradeForm component

# Notable Changes
- Added the `ProController`, a component that encapsulates the BillingOptions, UserCount and PriceCallout components
- Made use of `ProController` in the UpgradeForm2 component
- Some variable renaming: `setValue` to `setFormValue`, `planString` to `newPlan`
- Created some types to standardize the values
- Changed a bit the UpdateButton component to remove unnecessary prop drilling and adjusted test as well

* there are 2 tests that are skipped, I need a hand w/ those 🙃 

# Screenshots
Screenshot 2023-11-28 at 3 40 14 PM

(Still only working for Pro. I'm abstaining from doing a full-blown check in every of these PRs as the code isn't running and I'll do one when all the 3 services are in)

# Legal Boilerplate

### DIFF
--- a/src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm2/Controllers/ProPlanController/ProPlanController.spec.tsx
+++ b/src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm2/Controllers/ProPlanController/ProPlanController.spec.tsx
@@ -254,14 +254,6 @@ describe('ProPlanController', () => {
         expect(optionBtn).toHaveClass('bg-ds-primary-base')
       })
 
-      it.skip('renders the seat input with 10 seats (existing subscription)', async () => {
-        setup({ planValue: Plans.USERS_PR_INAPPM })
-        render(<ProPlanController {...props} />, { wrapper: wrapper() })
-
-        const seatCount = await screen.findByRole('spinbutton')
-        expect(seatCount).toHaveValue(10)
-      })
-
       it('has the price for the year', async () => {
         setup({ planValue: Plans.USERS_PR_INAPPM })
         render(<ProPlanController {...props} />, { wrapper: wrapper() })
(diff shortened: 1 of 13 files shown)


src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm2/Controllers/ProPlanController/ProPlanController.spec.tsx
```diff
@@ -0,0 +1,325 @@
...
+        const optionBtn = await screen.findByRole('button', { name: 'Monthly' })
+        expect(optionBtn).toBeInTheDocument()
+        expect(optionBtn).toHaveClass('bg-ds-primary-base')
+      })
+
+      it.skip('renders the seat input with 10 seats (existing subscription)', async () => {
+        setup({ planValue: Plans.USERS_PR_INAPPM })
+        render(<ProPlanController {...props} />, { wrapper: wrapper() })
+
+        const seatCount = await screen.findByRole('spinbutton')
```
rohitvinnakota-codecov, line 261: This relates to the `input` tag in `TextInput`. With the way `<input>` is used, you can only check for `min` and `size` attributes on the tag. You could use those to assert on these tests by passing 10 as the `min` attribute and checking the input's min. 

Testing the value in the input box in TS is a bit [trickier](https://stackoverflow.com/questions/53071851/getting-the-value-from-input-element-in-typescript). You could try creating a state function that explicitly sets the value to access it easier if you feel strongly about testing it.

adrian-codecov: Mmm issue isn't quite that, we can target the input's value, but that is determined by the `(...register('seats')` variable, which I'm unsure how to give a return value for

adrian-codecov: Thinking about it, there's a test at the parent level that also does this and we don't need to mock this value, so I'll just move the test there

adrian-codecov: The tests for that are already in the `UpgradeForm2` file, see this as an example, https://github.com/codecov/gazebo/blob/main/src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm2/UpgradeForm2.spec.jsx#L655

rohitvinnakota-codecov (approved):

PR #2447, opened 2023-12-08
[feat] Add useCommitComponents hook

rohitvinnakota-codecov:
# Description

This PR closes https://github.com/codecov/engineering-team/issues/856. Following up on the API changes from https://github.com/codecov/codecov-api/pull/268, this hook fetches all components for a given commit. 

In addition to this, for a given list of component names, it should return all components associated with that commit.

Screenshot 2023-12-08 at 2 13 45 PM
 

# Legal Boilerplate


RulaKhaled (approved):
lgtm just left one quick comment

PR #2460, opened 2023-12-14
feat: Update fileViewers with the right components selector

RulaKhaled:
# Description
Place the new multi selectors in PR file viewer, in commit details file viewer, and in coverage tab file viewer. 

# Notable Changes
- Test 
- Add each page's selector to its file viewer 

# Screenshots
Screenshot 2023-12-14 at 5 22 43 PM
Screenshot 2023-12-14 at 5 25 24 PM

https://github.com/codecov/gazebo/assets/91732700/6ba53665-dd46-45f5-931c-ca0649a885be

Note: the gray line will hold flags selector and other legends in upcoming issues, i'm trying not to bind it only to components rendering
 

# Legal Boilerplate


nicholas-codecov (commented):
Wondering @codecovdesign thoughts on adding this to the sticky header already in place so the user can select the components even if they've scrolled down, just to keep it inline with how the flag selector currently works on the file viewer.

RulaKhaled:
> Just taking an initial peak at the preview deploy, something seems up with the selector as it's see through with the labels:

Which page is that can we create a separate ticket for this please? Maybe we can look deeper into components selector itself 

as for the sticky header, we plan to do this when we move the other legends and flags filter to the div, for now it'll look kinda cluttered when we have more than two sticky rows:
Screenshot 2023-12-14 at 10 42 45 PM

RulaKhaled:
> The issue is on the coverage tab file viewer. This should only be a single addition so I don't think it's worth creating a ticket and waiting for. (unless the fix doesn't work)

done ty

rohitvinnakota-codecov (approved):

nicholas-codecov (approved):
Looks good 👍

PR #2469, opened 2023-12-20
fix: Take out components selector from ui + rename it

RulaKhaled:
# Description
The components selector for the coverage tab should live in a shared folder that is not the UI. UI is for custom reusable components which should follow some rules, for more context: https://www.notion.so/sentry/UI-Best-Practices-Plan-b59f7d7ef5774600837d40d1bbdb4b8e 

also renamed it so we don't get confused with the components selector that lives under commit details page 

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


nicholas-codecov (approved):
Just playing around with the name a little thoughts something more like `ComponentsMultiSelect` just more clear as to what the component actually is imo.

PR #2472, opened 2023-12-23
677 remove trials from all public repos

adrian-codecov:
# Description
We're adding a check to surface trial starting information only when an org has at least 1 private repo.

closes https://github.com/codecov/engineering-team/issues/677

# Notable Changes
- Small changes in a couple of components trigger a start to the trial
- Add hasPrivateRepos resolver
- Adjust tests

# Link to Sample Entry

# Legal Boilerplate


rohitvinnakota-codecov (approved):

src/services/account/usePlanData.ts
```diff
@@ -12,6 +12,7 @@ export const TrialStatuses = {
 
 const PlanDataSchema = z
   .object({
+    hasPrivateRepos: z.boolean().nullish(),
```
RulaKhaled, line 15: looking at https://github.com/codecov/codecov-api/blob/main/graphql_api/types/owner/owner.graphql#L18
it's not nullish

adrian-codecov: Good catch! Gonna adjust this

adrian-codecov: Since this is a required field now, had to adjust other tests to add it, hence why the files changed blew up a bit, just letting you know 👌

PR #2480, opened 2024-01-02
 Add app install banner + modal to the UI

rohitvinnakota-codecov:
# Description

This PR adds a banner when a user attempts to install an app from Github of which they are not the owner of. This PR adds the following changes. 

- Check if `setup_action` is part of the URL params(this is a param that Github adds automatically after attempted installation).
- Update `HomePageRedirect` so that we preserve `setup_action` if it is on the URL. We were overwriting it previously to strip params.  
- Conditionally render the banner. Note that I used the existing `RequestInstallBanner` component as the UI element. 
- Convert `RequestInstallBanner` to typescript

Design linked here: 
https://github.com/codecov/engineering-team/issues/764

# Screenshots

https://github.com/codecov/gazebo/assets/148245014/edbe9b8d-cac3-4b27-9bd2-6be8271cc966

# Legal Boilerplate


src/App.jsx
nicholas-codecov, line 1: Do you know why we were removing them before, and why we're keeping them now?

rohitvinnakota-codecov: Discussed offline. This is the intended behaviour.

nicholas-codecov (approved):
Awesome work!

PR #2481, opened 2024-01-03
[feat] Add useRepoATS hook

rohitvinnakota-codecov:
# Description

Now that we have `primaryLanguage` and `isATSConfigured` supported on the [backend](https://github.com/codecov/codecov-api/pull/323), this change adds a [useRepoATS](url) hook to be consumed on the main repo view(to be added in a follow-up PR). 

The idea is to call this hook and use the response to conditionally render the ATS tab and its contents as per the [design](https://github.com/codecov/engineering-team/issues/811). 

This PR closes https://github.com/codecov/engineering-team/issues/931.

# Legal Boilerplate

### DIFF
--- a/src/services/ats/useRepoATS.tsx
+++ b/src/services/ats/useRepoATS.tsx
@@ -10,8 +10,8 @@ import A from 'ui/A'
 
 const RepositorySchema = z.object({
   __typename: z.literal('Repository'),
-  isATSConfigured: z.boolean().nullish(),
-  primaryLanguage: z.string().nullish(),
+  isATSConfigured: z.boolean().nullable(),
+  primaryLanguage: z.string().nullable(),
 })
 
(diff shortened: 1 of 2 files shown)


src/services/ats/useRepoATS.tsx
```diff
@@ -0,0 +1,114 @@
...
+import {
+  RepoNotFoundErrorSchema,
+  RepoOwnerNotActivatedErrorSchema,
+} from 'services/repo'
+import Api from 'shared/api'
+import A from 'ui/A'
+
+const RepositorySchema = z.object({
+  __typename: z.literal('Repository'),
+  isATSConfigured: z.boolean().nullish(),
```
adrian-codecov, line 13: Can these be undefined or just null? If so might be better to use `nullable`() instead

adrian-codecov (approved):

PR #2488, opened 2024-01-06
fix: Fix BB bug

RulaKhaled:
# Description
we counted on the fact that users would have at least email when they login to Codecov, but we realised BB users can have both business email and email null, this is changing Zod schema to accommodate such use case .

# Screenshots
https://github.com/codecov/gazebo/assets/91732700/f7302a66-f472-4571-8c69-4ebb93220caf

https://github.com/codecov/gazebo/assets/91732700/d06d7709-9c51-4474-b96a-2701bc68606f

# Link to Sample Entry

# Legal Boilerplate


nicholas-codecov (approved):
Send it

PR #2494, opened 2024-01-08
feat: Create hook to fetch bundle analysis summary for a pull

nicholas-codecov:
# Description

This PR adds in a new hook that will be used to get the information required for the bundle analysis dropdown that will be used when both Coverage and a Bundle analysis data is present inside a repo for a given pull.

GH codecov/engineering-team#992

# Notable Changes

- Create new `usePullBADropdownSummary` hook
- Write up tests for `usePullBADropdownSummary`


src/services/pull/usePullBADropdownSummary.tsx
```diff
@@ -0,0 +1,170 @@
...
+}
+
+export function usePullBADropdownSummary({
+  provider,
+  owner,
+  repo,
+  pullId,
+}: UseCommitDropdownSummaryArgs) {
+  return useQuery({
+    queryKey: ['CommitSummary', provider, owner, repo, pullId],
```
rohitvinnakota-codecov, line 121: Think we want `PullBADropdownSummary`. 

I might have missed this in the previous PR so that might need an update.

nicholas-codecov: Nice catch, all fixed up.

PR #2499, opened 2024-01-10
feat: Create commit page table for bundle analysis

nicholas-codecov:
# Description

This PR adds in an initial implementation of a table displaying bundle analysis information. The error handling is a little light here, but will be fleshed out down the road a bit more when I integrate the table to the dropdown and the whole page itself.

GH codecov/engineering-team#994

# Notable Changes

- Create new `CommitBundleAnalysisTable` component
- Create parent `CommitBundleAnalysis` component
- Fix up `handlers.js`
- Few tweaks to `useCommitBundleList`
- Add/Update tests

# Screenshots

Screenshot 2024-01-10 at 13 44 38


src/pages/CommitDetailPage/CommitBundleAnalysis/CommitBundleAnalysisTable/CommitBundleAnalysisTable.tsx
```diff
@@ -0,0 +1,186 @@
...
+  const { data } = useCommitBundleList({
+    provider,
+    owner,
+    repo,
+    commitid: commitSha,
+  })
+
+  if (
+    data?.commit?.bundleAnalysisCompareWithParent?.__typename !==
+    'BundleAnalysisComparison'
```
RulaKhaled, line 178: `l`: can you turn this into an enum and update line 74 too

src/pages/CommitDetailPage/CommitBundleAnalysis/CommitBundleAnalysis.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const Loader = () => (
+  <div className="flex flex-1 justify-center">
+    <Spinner />
+  </div>
+)
+
+const CommitBundleAnalysis: React.FC = () => {
+  return (
+    <>
```
RulaKhaled, line 17: is there a need for this extra fragment?

nicholas-codecov: This is really just a placeholder component until it's fully fleshed out, just gonna leave as is.

src/pages/CommitDetailPage/CommitBundleAnalysis/CommitBundleAnalysisTable/CommitBundleAnalysisTable.tsx
nicholas-codecov: Abstracting this out I think muddies the water as to what is being checked here when you take a peak at the condition, because of the way we're defining our zod schemas TypeScript gives us only potential string literals instead of just the type string, blocking us from having any incorrect string values here. Abstracting this out would then require you to go take a peak at the enum to check the value whereas here you can see directly what is being checked.

I think we reach for these enums too much because we're so used to the protections they provide in JavaScript where TypeScript automatically handles this for us and will tell us when we're wrong.

src/pages/CommitDetailPage/CommitBundleAnalysis/CommitBundleAnalysis.tsx
RulaKhaled: do you mean we'll add a component inside of this fragment outside of suspense?

nicholas-codecov: This component is going to drastically change in the very very near future.

RulaKhaled: i still think we should remove the extra fragment though as we're viewing the current code, approving to unblock

RulaKhaled (approved):

PR #2518, opened 2024-01-16
Refactor Repos table in owner page and analytics page 

rohitvinnakota-codecov:
# Description

Updates the `ReposTable` component to use tanstack. This table is rendered on the owner page and the analytics page.  This PR also includes typescript conversions to `ReposTable` as well as `useRepos`. This closes https://github.com/codecov/engineering-team/issues/583.

# Screenshots

https://github.com/codecov/gazebo/assets/148245014/abd94831-885d-48f8-b7f8-f7cfe9b768e2

# Legal Boilerplate


src/shared/ListRepo/ReposTable/NoRepoCoverage.jsx
```diff
@@ -26,11 +26,11 @@ function NoRepoCoverage({
 }
 
 NoRepoCoverage.propTypes = {
-  owner: PropTypes.string.isRequired,
-  repoName: PropTypes.string.isRequired,
-  isCurrentUserPartOfOrg: PropTypes.bool.isRequired,
-  activated: PropTypes.bool.isRequired,
-  active: PropTypes.bool.isRequired,
+  owner: PropTypes.string,
```
rohitvinnakota-codecov, line 29: Long term fix here is to convert this file to typescript

src/shared/ListRepo/ReposTable/ReposTable.tsx
```diff
@@ -0,0 +1,212 @@
...
+    isFetching,
+    isFetchingNextPage,
+  } = useRepos({
+    activated,
+    sortItem: getOrderingDirection(sorting),
+    term: searchValue,
+    repoNames: filterValues,
+    owner,
+    isPublic: shouldDisplayPublicReposOnly,
+  })
```
nicholas-codecov, line 110: You'll need to disable suspense in this `useRepos` call, and grab the loading state and render a spinner in the table's body if loading, that way the table headers don't disappear when sorting

src/shared/ListRepo/ReposTable/ReposTable.tsx
```diff
@@ -0,0 +1,212 @@
...
+            hook="load-more"
+            isLoading={isFetchingNextPage}
+            onClick={fetchNextPage}
+            to={undefined}
+            disabled={false}
+          >
+            Load More
+          </Button>
+        </div>
+      )}
```
nicholas-codecov, line 207: Can you remove this and swap to using infinite scrolling? It's something we've been trying to move towards to enhance the users experience, you can see an example here: [load more trigger](https://github.com/codecov/gazebo/blob/main/src/pages/RepoPage/PullsTab/PullsTableTeam/PullsTableTeam.tsx#L28-L37), [loaders](https://github.com/codecov/gazebo/blob/main/src/pages/RepoPage/PullsTab/PullsTableTeam/PullsTableTeam.tsx#L178-L179).

nicholas-codecov (changes requested):
Two things to look at

src/services/repos/useRepos.tsx
```diff
@@ -0,0 +1,271 @@
...
+            pageInfo {
+                hasNextPage
+                endCursor
+            }
+          }
+        }
+      }
+
+      ${repositoryFragment}
+  `
```
nicholas-codecov, line 193: `l`: This is me being just a little picky, but can you update these with their formatted versions so they're easier to read:

```graphql
query ReposForOwner(
  $filters: RepositorySetFilters!
  $owner: String!
  $ordering: RepositoryOrdering!
  $direction: OrderingDirection!
  $after: String
  $first: Int
) {
  owner(username: $owner) {
    repositories(
      filters: $filters
      ordering: $ordering
      orderingDirection: $direction
      first: $first
      after: $after
    ) {
      edges {
        node {
          ...RepoForList
        }
      }
      pageInfo {
        hasNextPage
        endCursor
      }
    }
  }
}

${repositoryFragment}
```

src/services/repos/useRepos.tsx
```diff
@@ -0,0 +1,271 @@
...
```
nicholas-codecov, line 135: `l`: Another quick format change:

```graphql
query MyRepos(
  $filters: RepositorySetFilters!
  $ordering: RepositoryOrdering!
  $direction: OrderingDirection!
  $after: String
) {
  me {
    viewableRepositories(
      filters: $filters
      ordering: $ordering
      orderingDirection: $direction
      first: $first
      after: $after
    ) {
      edges {
        node {
          ...RepoForList
        }
      }
      pageInfo {
        hasNextPage
        endCursor
      }
    }
  }
}

${repositoryFragment}
```

src/services/repos/useRepos.tsx
```diff
@@ -0,0 +1,271 @@
...
+function fetchMyRepos({
+  provider,
+  variables,
+  after,
+  signal,
+}: FetchMyReposArgs) {
+  const query = `
+    query MyRepos($filters: RepositorySetFilters!, $ordering: RepositoryOrdering!, $direction: OrderingDirection!, $after: String) {
+        me {
+          viewableRepositories(filters: $filters, ordering: $ordering, orderingDirection: $direction, first: $first, after: $after) {
```
nicholas-codecov, line 120: `m`: There seems to be an error between the variables being declared and the ones being used. You seem to be missing `$first: Int` in your query operation.

src/shared/ListRepo/ReposTable/ReposTable.spec.tsx
```diff
@@ -228,8 +252,71 @@ describe('ReposTable', () => {
...
+        let tableElements = await screen.findAllByText(/Repo name/)
+        expect(tableElements[0]).toHaveTextContent('Repo name 1')
+        expect(tableElements[1]).toHaveTextContent('Repo name 2')
+        expect(tableElements[2]).toHaveTextContent('Repo name 3')
+
+        await user.click(header)
+        tableElements = await screen.findAllByText(/Repo name/)
+        expect(tableElements[0]).toHaveTextContent('Repo name 3')
+        expect(tableElements[1]).toHaveTextContent('Repo name 2')
+        expect(tableElements[2]).toHaveTextContent('Repo name 1')
```
nicholas-codecov, line 274: `m`: I would recommend not checking the table sort order and rather checking that the correct variables are passed. Since the sorting is being done on the API side these tests are really only checking that your mocked out response is sending the correct data. What I would suggest switching to is create a mock in the `setup` function and call it with the `req.variables` (i think that's right) inside the mocked GQL request and check that the the mock function was called with the right variables.

Same goes for the other sorting tests.

> This may be going against something I have said previously (at some point in time), but I think this is a better approach.

nicholas-codecov (changes requested):
Few things to take a peak at.

nicholas-codecov (approved):
Thanks for those changes, approved ✅

PR #2531, opened 2024-01-23
Update flag select in pull request view to only show relevant flags 

rohitvinnakota-codecov:
# Description

This PR closes https://github.com/codecov/engineering-team/issues/823. Following up from https://github.com/codecov/codecov-api/pull/350, we now only see relevant flags to the PR in the pull view. This uses the same query as in the flags tab. 

The query used in the flags tab was modified to accept search terms to maintain behaviour in the Flags Select component

# Screenshots

Screenshot 2024-01-26 at 2 48 44 PM

# Legal Boilerplate


nicholas-codecov (approved):
LGTM

PR #2532, opened 2024-01-23
[feat] Add metric to track app install shares

rohitvinnakota-codecov:
# Description

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/shared/GlobalTopBanners/RequestInstallBanner/RequestInstallBanner.spec.tsx
+++ b/src/shared/GlobalTopBanners/RequestInstallBanner/RequestInstallBanner.spec.tsx
@@ -1,3 +1,4 @@
+import * as Sentry from '@sentry/react'
 import { render, screen, waitFor } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { MemoryRouter, Route } from 'react-router-dom'
@@ -11,6 +12,16 @@ import RequestInstallBanner from './RequestInstallBanner'
 jest.mock('services/navigation')
 jest.mock('config')
 jest.mock('shared/featureFlags')
+jest.mock('@sentry/react', () => {
+  const originalModule = jest.requireActual('@sentry/react')
+  return {
+    ...originalModule,
+    metrics: {
+      ...originalModule.metrics,
+      increment: jest.fn(),
+    },
+  }
+})
 
 console.error = () => {}
 
@@ -29,6 +40,8 @@ type SetupArgs = {
 }
 
 describe('RequestInstallBanner', () => {
+  afterAll(() => jest.restoreAllMocks())
+
   function setup({ setUpAction = 'request', isSelfHosted = false }: SetupArgs) {
     const user = userEvent.setup()
     const mockSetItem = jest.spyOn(window.localStorage.__proto__, 'setItem')
@@ -85,6 +98,21 @@ describe('RequestInstallBanner', () => {
               ' admin or owner to assist.'
           )
           expect(modalText).toBeInTheDocument()
+        })
+
+        it('should call capture the share request metrics', async () => {
+          const { user, mockGetItem } = setup({})
+          render(<RequestInstallBanner />, { wrapper: wrapper('/gh/codecov') })
+
+          mockGetItem.mockReturnValue(null)
+
+          const btn = screen.getByRole('button', { name: /Share Request/ })
+          expect(btn).toBeInTheDocument()
+          await user.click(btn)
+
+          expect(Sentry.metrics.increment).toHaveBeenCalledWith(
+            'user_shared_request'
+          )
         })
       })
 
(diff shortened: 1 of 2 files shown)


src/shared/GlobalTopBanners/RequestInstallBanner/RequestInstallBanner.tsx
```diff
@@ -14,6 +15,7 @@ import TopBanner, { saveToLocalStorage } from 'ui/TopBanner'
 const APP_INSTALL_BANNER_KEY = 'request-install-banner'
 const COPY_APP_INSTALL_STRING =
   "Hello, could you help approve the installation of the Codecov app on GitHub for our organization? Here's the link: https://github.com/apps/codecov/installations/select_target"
+const SHARE_REQUEST_METRICS_KEY = 'user_shared_request'
```
AbhiPrasad, line 18: if we have this more widespread might make sense to colocate all metrics keys in a shared file somewhere so no risk of duplicates at all.

Also in Sentry we typically use `.` for the delimiter, and `_` to represent compound words, and use plural for metrics, so `user.shared.requests` in this case.

image

Do we also want to add `request.install` prefix or something similar?

`request_install.user.shared.requests`

AbhiPrasad (commented):
Thanks for giving Sentry Metrics a try! For reference docs can be found here: https://github.com/getsentry/sentry-javascript/discussions/9938

src/shared/GlobalTopBanners/RequestInstallBanner/RequestInstallBanner.tsx
AbhiPrasad: actually we don't use plural I think I just hallucinated that - sorry for the fake news lol

`request_install.user.shared.request` makes the most sense to me

nicholas-codecov: @AbhiPrasad, thoughts on moving this to our `sentry.ts` file and creating an `as const` object that contain our keys, we could also use functions to create a base key and extend the remainder of the key, something like:

```ts
 // sentry.ts
const METRIC_KEYS = {
  REQUEST_INSTALL: (key: string) => `request_install.${key}`
} as const

// random file
const handleClick = () => {
  metrics.increment(METRIC_KEYS.REQUEST_INSTALL('user.shared.request'))
}
```

AbhiPrasad: I think a function abstraction introduces a lot of complexity that we don't need right now, let's start simple with just plain strings? We can move to introduce abstraction when there are more metrics and we feel like we need to enforce naming.

```ts
// sentry/metricKey.ts
export const REQUEST_INSTALL_USER_SHARED_REQUEST = 'request_install.user.shared.request';
```

```
// random file
import { REQUEST_INSTALL_USER_SHARED_REQUEST } from 'blah/blah/sentry/metricKey.ts';

metrics.increment(REQUEST_INSTALL_USER_SHARED_REQUEST);
```

You can also write a wrapper function around `metrics.increment` and try to enforce the key names via a string union or even TS template literal types.

rohitvinnakota-codecov: I can document the current process(it's fairly straightforward). 

It's hard to define an abstraction that makes sense without seeing more of these in the codebase so I am willing to wait till then and use this approach for now.

src/shared/GlobalTopBanners/RequestInstallBanner/RequestInstallBanner.spec.tsx
```diff
@@ -86,6 +99,21 @@ describe('RequestInstallBanner', () => {
...
+          mockGetItem.mockReturnValue(null)
+
+          const btn = screen.getByRole('button', { name: /Share Request/ })
+          expect(btn).toBeInTheDocument()
+          await user.click(btn)
+
+          expect(Sentry.metrics.increment).toHaveBeenCalledWith(
+            'request_install.user.shared.request'
+          )
+        })
```
nicholas-codecov, line 116: This is more of a philosophical question, do we want to be explicitly testing that this metric gets picked up, since we're even mocking it out we're not testing any of the functionality of it? The big question is that do we gain anything meaningful?

rohitvinnakota-codecov: I'm treating `Sentry.metrics` as a 3rd party library relative to the gazebo repo, and think it's not necessary to test the internals of the metrics functions. 

It is useful to have the tests for metrics still since we are testing whether we attempt to capture something. 

If there are refactors to components in the future, the tests and descriptions should provide a way to maintain this behaviour, and also that we are capturing the right kind of metric.

PR #2533, opened 2024-01-23
Fix default billing control behaviour 

rohitvinnakota-codecov:
# Description
This closes https://github.com/codecov/internal-issues/issues/197. 

- [x] Test with Team account on monthly plan 
- [x] Test with Pro account on monthly plan 
- [x] Test with Sentry account on monthly plan

### What this PR aims to fix: 

When changing the plan (team to pro or pro to team) we always snap back to annual as the selected billing cadence

When arriving on the manage plan page we always default to annual as the selection and don’t automatically select the current billing cadence

The solution in both cases would be to show what has already been selected instead.
For 1, if you have selected monthly in the past we surface it by default when you land on this page. If you have selected it while on the page and you then change the plan, we don't change the billing cadence.
For 1, show the current billing cadence by default when first navigating to the plan page.

# Screenshots

https://github.com/codecov/gazebo/assets/148245014/69981577-d5c6-4611-97da-29c74e19bee2

# Legal Boilerplate


src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeDetails/UpgradeDetails.tsx
```diff
@@ -12,9 +12,15 @@ function UpgradeDetails({
 }: {
   selectedPlan: z.infer<typeof IndividualPlanSchema>
 }) {
-  if (selectedPlan?.value === Plans.USERS_SENTRYY) {
+  if (
+    selectedPlan?.value === Plans.USERS_SENTRYY ||
```
adrian-codecov, line 16: There's utils for this that we could use, like the `isTeamPlan` or `isSentryPlan`

src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/UpgradeForm.jsx
```diff
@@ -78,6 +79,7 @@ function UpgradeForm({ selectedPlan, setSelectedPlan }) {
       </div>
       <PlanTypeOptions
         setFormValue={setFormValue}
+        getFormValues={getFormValues}
```
adrian-codecov, line 82: Might not be needed after this https://github.com/codecov/gazebo/pull/2533/files#r1468071697

adrian-codecov (commented):
Overall lgtm, just some quick changes and then I'm gtg

adrian-codecov (approved):

PR #2554, opened 2024-01-31
feat: Create hook to fetch branch summary for a branch

nicholas-codecov:
# Description

This PR adds in a new hook to fetch summary info and a the list of bundles for a given branch. The reason why we're fetching these in a single hook is because the list of bundles is not paginated nor sortable via the API, as well this is a stopgap hook until the bundle tab is fully built out, and not the final implementation.

GH codecov/engineering-team#1024

# Notable Changes

- Create new `useBranchBundles` hook
- Create tests for new hook

### DIFF
--- a/src/services/branches/useBranchBundles.tsx
+++ b/src/services/branches/useBranchBundles.tsx
@@ -10,10 +10,17 @@ import {
 import Api from 'shared/api'
 import A from 'ui/A'
 
+const BundleSchema = z.object({
+  name: z.string(),
+  sizeTotal: z.number(),
+  loadTimeTotal: z.number(),
+})
+
 const BundleAnalysisReportSchema = z.object({
   __typename: z.literal('BundleAnalysisReport'),
   sizeTotal: z.number(),
   loadTimeTotal: z.number(),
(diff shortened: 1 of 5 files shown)


src/services/branches/useBranchBundles.tsx
```diff
@@ -0,0 +1,172 @@
...
+  repo,
+  branch: branchArg,
+}: UseBranchBundleSummaryArgs) => {
+  const { data: repoOverview, isLoading } = useRepoOverview({
+    provider,
+    repo,
+    owner,
+  })
+
+  const branch = branchArg ?? repoOverview?.defaultBranch
```
RulaKhaled, line 116: what if repo doesn't have a default branch?

nicholas-codecov: According to the DB the default branch field on the repos table is not null, and incase the branch doesn't have one it looks like it will set a default value to that field, but just incase I'll add in a check that will continue to disable the query if there isn't a branch.

RulaKhaled (approved):

PR #2565, opened 2024-02-06
[fix] Disable update on initial render on billing page 

rohitvinnakota-codecov:
# Description

We currently have a bug where the `Update` button is not disabled when a user is on a team plan and goes to the billing page for the first time. This PR fixes the issue by defaulting the form values to the user's current plan. 

**Before**

https://github.com/codecov/gazebo/assets/148245014/6404c261-7cc4-43bf-808e-c05b4ae85f9d

https://github.com/codecov/gazebo/assets/148245014/174b982c-8aec-4fd9-ba65-516aa2fd921d

**After**

# Legal Boilerplate


src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/UpdateButton/UpdateButton.tsx
```diff
@@ -20,14 +20,12 @@ const UpdateButton: React.FC<BillingControlsProps> = ({
 }) => {
   const { provider, owner } = useParams<{ provider: string; owner: string }>()
   const { data: accountDetails } = useAccountDetails({ provider, owner })
-
```
RulaKhaled, line 23: what's up with the empty lines

RulaKhaled (approved):

PR #2588, opened 2024-02-14
fix: Remove redirect from coverage onboarding

nicholas-codecov:
# Description

We shouldn't redirect from the new repo page component if the repo is active.


RulaKhaled (approved):
Lets seeee

PR #2600, opened 2024-02-16
feat: Add in new formatTimeToString function handling milliseconds 

nicholas-codecov:
# Description

Add in a new util function that formats times to string using milliseconds instead of seconds, as we're moving to use milliseconds in the API in some upcoming fields.

Closes codecov/engineering-team#1200

# Notable Changes

- Deprecate older `formatTimeToString`
- Introduce new `formatTimeToString`


src/shared/utils/bundleAnalysis.ts
```diff
@@ -1,6 +1,6 @@
-const KILOBYTE = 1000
-const MEGABYTE = 1000000
-const GIGABYTE = 1000000000
+const KILOBYTE = 1_000
```
RulaKhaled, line 1: But why the underscore?

PR #2602, opened 2024-02-16
feat: #235 - Org page CTA Button Update

ajay-sentry:
# Description

Minor change for the CTA button from "setup repo" to "Configure"

[Related Ticket](https://github.com/orgs/codecov/projects/31/views/12?pane=issue&itemId=33571451)

# Code Example

# Notable Changes

Converts the InactiveRepo File to TS

# Screenshots

Screenshot 2024-02-16 at 1 02 18 PM

# Legal Boilerplate


src/shared/ListRepo/InactiveRepo/InactiveRepo.tsx
```diff
@@ -0,0 +1,32 @@
...
+  owner: string
+  isActive: boolean
+  repoName?: string
+  isCurrentUserPartOfOrg?: boolean
+}) {
+  if (isActive) return <>Deactivated</>
+  if (!isCurrentUserPartOfOrg) return <>Inactive</>
+
+  return (
+    // @ts-ignore
```
ajay-sentry, line 18: Added this to avoid making a bunch of changes to the AppLink component. Lemme know if this isn't the right way we go about doing this though

rohitvinnakota-codecov: It might be best to also convert `AppLink` + the spec file to typescript. At minimum, we should create a ticket for that and link the issue as comment `TODO`. 

If you can do it in this PR, that would be great! Let me know if it will require a lot of changes though.

rohitvinnakota-codecov (approved):

PR #2605, opened 2024-02-20
feat: Create useBundleBranchSelector

nicholas-codecov:
# Description

This PR creates the new `useBundleBranchSelector` to be used in the upcoming branch context selector for the updated bundles tab.

GH codecov/engineering-team#1201

# Notable Changes

- Create `useBundleBranchSelector`
- Add in tests


src/pages/RepoPage/BundlesTab/BundleContent/BundleSummary/hooks/useBundleBranchSelector.ts
```diff
@@ -0,0 +1,59 @@
...
+
+interface URLParams {
+  provider: string
+  owner: string
+  repo: string
+  branch?: string
+  ref?: string
+}
+
+const getDecodedBranch = (branch?: string) =>
```
ajay-sentry, line 13: nit: if branch and ref don't exist, should we be explicit about the "falsey" return value we want from the ternary? Like '' or similar

src/pages/RepoPage/BundlesTab/BundleContent/BundleSummary/hooks/useBundleBranchSelector.ts
```diff
@@ -0,0 +1,59 @@
...
+  !!branch ? decodeURIComponent(branch) : branch
+
+const getDecodedRef = (ref?: string) => (!!ref ? decodeURIComponent(ref) : ref)
+
+interface UseBundleBranchSelectorProps {
+  branches: Branch[] | null
+  defaultBranch: string
+}
+
+export function useBundleBranchSelector({
```
ajay-sentry, line 23: Is this essentially the same as src/pages/RepoPage/CoverageTab/hooks/useBranchSelector.js?

nicholas-codecov: Yes it is pretty much the same

src/pages/RepoPage/BundlesTab/BundleContent/BundleSummary/hooks/useBundleBranchSelector.ts
nicholas-codecov: We can't use `""`, but we can return `undefined` instead

ajay-sentry: '' is what would have been returned from the conditional if branch was ''. Which is what spawned the question in the first place :) 

If that's what we want then I'm aligned

Screenshot 2024-02-20 at 2 07 12 PM

nicholas-codecov: So one small thing that I did also change is to swap to using the `??` nullish coalescing operator which only fails the left hand side if `null` or `undefined` as it is a bit safer to use over the typical `||` logical OR.

rohitvinnakota-codecov (approved):

src/pages/RepoPage/BundlesTab/BundleContent/BundleSummary/hooks/useBundleBranchSelector.ts
ajay-sentry: Yep yep totally makes sense!

PR #2608, opened 2024-02-20
After trial start action display to org > repos

rohitvinnakota-codecov:
# Description

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/PlanPage/subRoutes/CurrentOrgPlan/CurrentPlanCard/shared/ActionsBilling/ActionsBilling.jsx
+++ b/src/pages/PlanPage/subRoutes/CurrentOrgPlan/CurrentPlanCard/shared/ActionsBilling/ActionsBilling.jsx
@@ -9,6 +9,7 @@ import {
   usePlanData,
 } from 'services/account'
 import { useStartTrial } from 'services/trial'
+import { useRedirect } from 'shared/useRedirect'
 import {
   canApplySentryUpgrade,
   isFreePlan,
@@ -28,6 +29,7 @@ function PlansActionsBilling({ plan }) {
   })
 
   const { mutate, isLoading } = useStartTrial()
+  const { hardRedirect } = useRedirect({ href: `/${provider}/${owner}` })
 
   const canStartTrial =
     planData?.plan?.trialStatus === TrialStatuses.NOT_STARTED &&
@@ -40,6 +42,7 @@ function PlansActionsBilling({ plan }) {
         <Button
           onClick={() => {
             mutate({ owner })
+            hardRedirect()
           }}
           hook="start-trial"
           variant="primary"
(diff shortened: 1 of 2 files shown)


src/services/trial/useStartTrial.ts
```diff
@@ -75,16 +78,8 @@ export const useStartTrial = () => {
...
-      renderToast({
-        type: 'generic',
-        title: '14 day trial has started',
-        content: '',
-        options: {
-          duration: 5000,
-          position: 'bottom-left',
-        },
-      })
+      localStorage.setItem(LOCAL_STORAGE_USER_STARTED_TRIAL_KEY, 'true')
```
nicholas-codecov, line 81: Rather than storing this in `localStorage` what do you think about just adding a query param to the redirect, something like `?trial_started=true`, that way we should be able to avoid using a useEffect, which we should be doing at all costs.

src/pages/OwnerPage/OwnerPage.jsx
```diff
@@ -39,6 +43,24 @@ function OwnerPage() {
...
+        title: '14 day trial has started',
+        content: '',
+        options: {
+          duration: 5000,
+          position: 'bottom-left',
+        },
+      })
+      localStorage.removeItem(LOCAL_STORAGE_USER_STARTED_TRIAL_KEY)
+    }
+  }, [userStartedTrial])
```
nicholas-codecov, line 63: Just an aside, do we really need a useEffect here? What specifically are we using the features of useEffect for that we can't get with just the `if` statement?

nicholas-codecov (commented):
Somethings to take a peak at

src/pages/OwnerPage/OwnerPage.jsx
rohitvinnakota-codecov: The toast was rendering multiple times on load. With `useEffect`, we can make sure it only runs once after the initial render.

nicholas-codecov: Could you not assign the value accessed from local storage with `let` then set the value to false in your conditional right after the first render?

src/services/trial/useStartTrial.ts
rohitvinnakota-codecov: I did this to avoid making changes in `App.jsx` around handling redirects. I'm not sure if we can avoid `useEffect` in that scenario either if there are multiple re-renders so both solutions could work.

nicholas-codecov: What would you need to change in `App` if you were to use a query param?

rohitvinnakota-codecov: Adding some additional logic to `HomePageRedirect` + tests and all of that

src/pages/OwnerPage/OwnerPage.jsx
rohitvinnakota-codecov: I get mixed behaviour without `useEffect. Either I get the toast rendering many times, or not at all.

nicholas-codecov: Grr okay

src/services/trial/useStartTrial.ts
nicholas-codecov: Yea homepage redirect is not setup at all to handle multiple query params, we should take a peak at fixing that.

nicholas-codecov (approved):
LGTM

PR #2612, opened 2024-02-21
ref: Convert TokensTable to tanstack table

spalmurray-codecov:
# Description

Refactor TokensTable and the hooks it depends on: `useSessions` and `useRevokeUserToken`. Also, update the column layout of TokensTable to better accommodate long names. Since the 'Token' column has fixed width, allowing the name column to take up more width seems like good change. See change in screenshot.

Closes codecov/engineering-team#601

# Notable Changes

- Convert `TokensTable` to TS/tanstack table
- Update column layout of TokensTable
- Convert `useSessions` to TS
- Convert `useRevokeUserToken` to TS

# Screenshots
layoutchange


src/services/access/useRevokeUserToken.ts
```diff
@@ -15,15 +15,15 @@ export function useRevokeUserToken({ provider }) {
...
         provider,
         query,
-        variables,
+        variables: {
+          input: { tokenid },
+        },
         mutationPath: 'revokeUserToken',
       }).then((res) => {
-        queryClient.invalidateQueries('sessions')
+        queryClient.invalidateQueries(['sessions'])
```
ajay-sentry, line 26: jw, was this needed?

src/services/access/useSessions.ts
```diff
@@ -0,0 +1,115 @@
...
+      Api.graphql({
+        provider,
+        query,
+        signal,
+      }).then((res) => {
+        const parsedRes = RequestSchema.safeParse(res?.data)
+
+        if (!parsedRes.success) {
+          return Promise.reject({
+            status: 404,
```
ajay-sentry, line 97: wrote about this in one of Nick's reviews, but what are your thoughts on making this a 400 vs 404?

ajay-sentry (approved):
looks good to me!

src/services/access/useSessions.ts
spalmurray-codecov: I agree 404 seems off, but I think this would probably be a 5XX kind of problem most of the time it occurs. Maybe I'm wrong on that feeling though.

src/services/access/useRevokeUserToken.ts
nicholas-codecov: (it probably didn't work fully properly before 😬)

nicholas-codecov (approved):
Looks all good to me ✅

PR #2613, opened 2024-02-21
Feat: #1057 - Clarify prerequisite of coverage reports

ajay-sentry:
# Description

Adds a little box to the Github, CircleCI and other CI "new" repo pages explaining to check our quick start docs confirming they have coverage reports set up already, along with a link to the docs.

# Notable Changes

Uncommented one of the book icons for this. There was a separate one on the [figma](https://www.figma.com/file/VQy5zqPaQ5Vwv7cAeOnQKX/GH-1049?type=design&node-id=3-2199&mode=design&t=9t9MMiZIi2Xuvj1G-0), but I figured it'd probably be better to use one we already have in the repo rather than importing a fresh one.

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/17359bbf-376c-43aa-956b-52889fe96638

# Link to Sample Entry

# Legal Boilerplate


nicholas-codecov (changes requested):
This seems to only update the repo token versions of the onboarding where we need to also support the global upload tokens as well. Since this is being reused all over the place in six different components I think this is a good time to pull this out into it's own component.

nicholas-codecov (approved):
Ty for those quick changes 🎉

PR #2615, opened 2024-02-22
feat: Create hook to fetch list of bundles for a repo

nicholas-codecov:
# Description

This PR creates a hook that fetches the list of bundle names and returns an array of the names. This hook will be used with the bundle select in the upcoming update to the bundles tab.

Closes codecov/engineering-team#1202

# Notable Changes

- Create new hook to fetch list of bundle names
- Add in tests


src/services/branches/useBranchBundlesNames.tsx
```diff
@@ -0,0 +1,177 @@
...
+        ) {
+          bundles =
+            data.owner.repository.branch.head.bundleAnalysisReport.bundles?.map(
+              ({ name }) => name
+            )
+        }
+
+        return { bundles }
+      }),
+    enabled: (isSuccess && isString(branch)) || opts?.enabled,
```
RulaKhaled, line 175: Can you explain the first condition please? this looks to me like it will be resolved AFTER the hook was resolved

nicholas-codecov: So this is to disable the query if no branch was passed, because we first need to make a call to grab the default branch from repo overview, which is why we're checking to see if the request was successful from repo overview and we have a valid branch, if we don't then we shouldn't run this query.

RulaKhaled: ohhhh isSuccess is coming from repoOverview, makes sense

RulaKhaled (approved):

PR #2619, opened 2024-02-23
feat: Create hook to fetch bundle tab summary data

nicholas-codecov:
# Description

This PR creates a new hook to fetch bundle summary information required for the new bundle summary section on the updated bundles tab.

Closes codecov/engineering-team#1206

# Notable Changes

- Move branch related bundle analysis hooks into their own directory
- Introduce a new `useBundleSummary` hook
- Tests


src/services/bundleAnalysis/useBundleSummary.tsx
```diff
@@ -0,0 +1,190 @@
...
+import { MissingHeadReportSchema } from 'services/comparison'
+import {
+  RepoNotFoundErrorSchema,
+  RepoOwnerNotActivatedErrorSchema,
+} from 'services/repo'
+import Api from 'shared/api/api'
+import type { NetworkErrorObject } from 'shared/api/helpers'
+import A from 'ui/A'
+
+const BundleDataSchema = z.object({
```
ajay-sentry, line 13: just want to confirm, none of these params are required? Similarly with the BundleAnalysisReportSchema

src/services/bundleAnalysis/useBundleSummary.tsx
```diff
@@ -0,0 +1,190 @@
...
+        provider,
+        query,
+        signal,
+        variables: { owner, repo, branch, bundle },
+      }).then((res) => {
+        const parsedData = RequestSchema.safeParse(res?.data)
+
+        if (!parsedData.success) {
+          return Promise.reject({
+            status: 404,
```
ajay-sentry, line 145: nit: would this be better as a 400? Signaling a bad request instead of not found

src/services/bundleAnalysis/useBundleSummary.tsx
nicholas-codecov: Yes they are non-null in the schema, everything was non-null by default beforehand and we're doing a much better job ensuring things aren't non-null when we're able to now.

src/services/bundleAnalysis/useBundleSummary.tsx
```diff
@@ -0,0 +1,190 @@
...
+        }
+
+        if (data?.owner?.repository?.__typename === 'OwnerNotActivatedError') {
+          return Promise.reject({
+            status: 403,
+            data: {
+              detail: (
+                <p>
+                  Activation is required to view this repo, please{' '}
+                  {/* @ts-expect-error */}
```
ajay-sentry, line 168: curious about this ts-expect-error

src/services/bundleAnalysis/useBundleSummary.tsx
ajay-sentry: Got it, thanks for the clarification

src/services/bundleAnalysis/useBundleSummary.tsx
nicholas-codecov: This is just a copy paste, ignoring the type issues with the A tag

ajay-sentry: Fair enough, plan to swap to applink in a future PR?

src/services/bundleAnalysis/useBundleSummary.tsx
```diff
@@ -0,0 +1,190 @@
...
+          return Promise.reject({
+            status: 404,
+            data: {},
+            dev: 'useBundleSummary - 404 Not found error',
+          } satisfies NetworkErrorObject)
+        }
+
+        if (data?.owner?.repository?.__typename === 'OwnerNotActivatedError') {
+          return Promise.reject({
+            status: 403,
```
ajay-sentry, line 163: nit: this one might be better as a 401 instead of 403 as well

ajay-sentry (approved):

src/services/bundleAnalysis/useBundleSummary.tsx
nicholas-codecov: No we don't use the AppLink for in components directly for links like this, the `A` is built on top of it for that reason

ajay-sentry: oh right, brain fart there lol

src/services/bundleAnalysis/useBundleSummary.tsx
ajay-sentry: For sure, the detail to get the user to Activate made me think it was more in line with unauthorized than forbidden though since the user _can_ fix it. I can see either way

src/services/bundleAnalysis/useBundleSummary.tsx
nicholas-codecov: Hrrm we're pretty standard across the board right now setting this as 404, if we're going to change to 400, we should also move all of the other ones at the same time. Are you able to create a ticket to address this?

PR #2622, opened 2024-02-24
feat: Create bundles tab summary section

nicholas-codecov:
# Description

This PR adds in a new `BundlesDetail` component that is used in the top right of the updated bundles tab.

Closes codecov/engineering-team#1207

# Notable Changes

- Create new `BundlesDetail` component
- Few tweaks to `useBundleSummary` hook
- Few small tweaks here and there
- Update to tests

# Screenshots

Screenshot 2024-02-23 at 20 45 28

### DIFF
--- a/src/services/repo/useRepoOverview.tsx
+++ b/src/services/repo/useRepoOverview.tsx
@@ -139,6 +139,6 @@ export function useRepoOverview({
         }
       })
     },
-    enabled: opts?.enabled,
+    enabled: !!opts?.enabled,
   })
 }
--- a/src/services/repo/useRepoOverview.tsx
+++ b/src/services/repo/useRepoOverview.tsx
@@ -57,12 +57,16 @@ interface UseRepoOverviewArgs {
   provider: string
   owner: string
   repo: string
+  opts?: {
+    enabled?: boolean
+  }
 }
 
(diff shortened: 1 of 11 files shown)


src/services/repo/useRepoOverview.tsx
```diff
@@ -135,5 +144,6 @@ export function useRepoOverview({
         }
       })
     },
+    enabled,
```
spalmurray-codecov, line 147: What is this for?

nicholas-codecov: From [TanStack Query Docs](https://tanstack.com/query/v4/docs/framework/react/reference/useQuery):

Screenshot 2024-02-26 at 14 06 34

nicholas-codecov: It allows us to stop a query (and skip a network request) in this instance I've added it in, because in the hook where I'm calling it I use it to grab the default branch if there is no passed branch. But if there is a passed branch I don't need to grab the default branch so I can just skip the request altogether.

spalmurray-codecov: Ahhh okay that makes sense! Thanks @nicholas-codecov

PR #2627, opened 2024-02-27
Chore: add Lodash import rule

ajay-sentry:
# Description

Aim of this PR is to fix instances where we are importing the full lodash module rather than isolated modules from it. This will help keep our bundle size a little bit lower!

# Code Example

Check screenshot for example code that fails the rule

# Notable Changes

Adds a new ESLint rule that checks to make sure lodash isn't imported blankly across the repo

# Screenshots

Screenshot 2024-02-27 at 10 29 56 AM

# Link to Sample Entry

# Legal Boilerplate


rohitvinnakota-codecov (approved):

nicholas-codecov (approved):
LGTM ✅

PR #2635, opened 2024-02-29
feat: Update assets table to have modules table expander

nicholas-codecov:
# Description

This PR adds in a new `ModulesTable` that is used as the sub-component for what the asset table is expanded. It shows all of the modules that make up that given asset, their type, size, and estimated download time.

Closes codecov/engineering-team#1211

# Notable Changes

- Create new `ModulesTable` component
- Update `AssetsTable`
- Update tests

# Screenshots

Screenshot 2024-02-29 at 08 34 57

Screenshot 2024-02-29 at 08 35 09


src/services/bundleAnalysis/useBundleAssetModules.tsx
```diff
@@ -212,5 +221,6 @@ export const useBundleAssetModules = ({
 
         return { modules }
       }),
+    enabled: enabled,
```
spalmurray-codecov, line 224: Do we need a test for this?

spalmurray-codecov (approved):
code looks good

src/services/bundleAnalysis/useBundleAssetModules.tsx
spalmurray-codecov: I guess codecov says we're covered actually lol

nicholas-codecov: So the reason why I tested the other one, and not this, is that this is much more straightforward. Whereas the other one we were also checking to see what the repo overview hook was also saying so it was a little more complicated scenario.

PR #2645, opened 2024-03-04
Update select placement in Fileviewer

rohitvinnakota-codecov:
# Description

Moves placements of flags/components selectors to be in-line with key. This closes https://github.com/codecov/engineering-team/issues/1282.

# Screenshots
**Before**
Screenshot 2024-03-04 at 11 20 53 AM

**After**
Screenshot 2024-03-04 at 11 20 34 AM

https://github.com/codecov/gazebo/assets/148245014/9865dfde-1d79-4619-8818-48344f323693

# Legal Boilerplate


ajay-sentry (approved):
C L E A N

nicholas-codecov (changes requested):
`h`: Can you take a look at the responsiveness of these changes, they seem to break the dropdowns atm

Screenshot 2024-03-05 at 07 00 53

nicholas-codecov (approved):
LGTM

src/pages/RepoPage/CoverageTab/subroute/Fileviewer/Fileviewer.jsx
```diff
@@ -34,6 +29,7 @@ function FileView() {
         sticky
         stickyPadding={215}
         showFlagsSelect={showFlagSelector}
+        showComponentsSelect
```
thomasrockhu-codecov, line 32: Just curious, my frontend is lacking. Why doesn't this also accept a var like `{showComponentsSelector}`?

src/ui/FileViewer/ToggleHeader/ToggleHeader.jsx
```diff
@@ -1,23 +1,27 @@
...
 }) {
   return (
-    <Title title={title} sticky={sticky}>
-      <TitleCoverage coverage={LINE_STATE.UNCOVERED} />
-      <TitleCoverage coverage={LINE_STATE.PARTIAL} />
-      <TitleCoverage coverage={LINE_STATE.COVERED} />
-      <TitleHitCount showHitCount={showHitCount} />
-      {showFlagsSelect && <TitleFlags />}
-    </Title>
+    <div>
```
thomasrockhu-codecov, line 15: Also just for my knowledge, why does this need to be `div` wrapped?

src/pages/RepoPage/CoverageTab/subroute/Fileviewer/Fileviewer.jsx
nicholas-codecov: Oh shoot you're right, this needs to be fixed @rohitvinnakota-codecov I totally missed this. Unless you're passing the value directly as @thomasrockhu-codecov is suggesting it will always be true, which is not always the case.

src/ui/FileViewer/ToggleHeader/ToggleHeader.jsx
nicholas-codecov: So wrapping it in a `div` means that the styles from the parent are only applied to the outer element, whereas the inner elements are unchanged by the outer styles. If you want a big longer of an explanation, let me know!

thomasrockhu-codecov: @nicholas-codecov oh absolutely! I've always just shoved things into a div to make it work, but I never knew the real reasoning

PR #2649, opened 2024-03-04
ref: Update pull bundle head list hook

nicholas-codecov:
# Description

Quick PR to update the pull head table data fetching hook so that we're using the latest fields to grab bundle information from.

# Notable Changes

- Update hook to grab data from new fields
- Update table to use new fields from API response
- Update tests


ajay-sentry (approved):
looks good

PR #2652, opened 2024-03-04
build: fix upload test results secrets accessing

joseph-sentry:
there was a mistake with the accessing of secrets here where I put `secrets.secrets`


nicholas-codecov (approved):
Whoops

PR #2657, opened 2024-03-06
don't get planData if user is not in org

JerrySentry:
# Description

TrialBanner is called by anyone visiting a org page, this change make it so that we don't retrieve the plan data of the org for users not belonging in the org. This change is necessary because in the API there is a change that hides plan / planData / hasPrivateRepos.

# Legal Boilerplate


ajay-sentry (approved):
lgtm, discussed on code review meeting

PR #2660, opened 2024-03-06
ref: #597 TS conversions for some flag table related components

ajay-sentry:
# Description

yak shaving around cleaning up some of the flag table component "innards" prior to converting the table itself to TS and tanstack table.

I find working from "out to in" helps me when doing refactors like this because I can see how a lot of stuff is building together to create these larger components

# Code Example

# Notable Changes

Nothing major, in the process of doing some of the conversions I did move things over from "Object.freeze" to generic consts, as well as explicitly declaring props in the areas we did some prop drilling; like for the progress and coverage progress components.

# Screenshots

Screenshot 2024-03-06 at 3 12 38 PM

# Manual Testing

- Navigated to flags table and confirmed no regression
- Clicked through the app and aside from my one accountDetails mis-step I think app looks to be functioning similarly

# Legal Boilerplate


src/services/account/useAccountDetails.ts
```diff
@@ -77,7 +77,7 @@ export const SubscriptionDetailSchema = z
     currentPeriodEnd: z.number(),
     customer: z
       .object({
-        discount: z.number(),
```
ajay-sentry, line 80: this is unrelated but is a regression I need to merge with one of my upcoming PRs

src/services/repo/useRepoBackfilled.ts
```diff
@@ -0,0 +1,73 @@
...
+        if (!parsedData.success) {
+          return Promise.reject({
+            status: 404,
+            data: {},
+            dev: 'useRepoBackfilled - 404 Not Found Error',
+          })
+        }
+
+        return {
+          ...parsedData.data.config,
```
rohitvinnakota-codecov, line 68: Should we do `data?.config`?

src/shared/utils/determineProgressColor.spec.ts
```diff
@@ -40,6 +40,7 @@ describe('determineProgressColor', () => {
   describe('coverage value is not a number', () => {
     it('returns default', () => {
       const result = determineProgressColor({
+        // @ts-expect-error
```
rohitvinnakota-codecov, line 43: what is this for?

src/ui/CoverageProgress/CoverageProgress.tsx
```diff
@@ -0,0 +1,20 @@
+import Progress from 'ui/Progress'
+import { progressColors, progressVariants } from 'ui/Progress/Progress'
+
+const CoverageProgress = ({
+  amount,
+  color = 'primary',
+  variant = 'default',
+}: {
+  amount: number
```
rohitvinnakota-codecov, line 9: This might break when the hooks return null on a type validation. Maybe we can do amount: number | null | undefined. Then you don't need the  /* @ts-expect-error */ in the spec file either

src/ui/Progress/Progress.tsx
```diff
@@ -1,33 +1,42 @@
...
-function Progress({ amount, label, color = 'default', variant = 'default' }) {
+} as const
+
+function Progress({
+  amount,
+  label,
+  color = 'default',
+  variant = 'default',
+}: {
+  amount: number
```
rohitvinnakota-codecov, line 24: same here

src/ui/CoverageProgress/CoverageProgress.tsx
rohitvinnakota-codecov: ex: useRepoCommitContents

src/ui/TotalsNumber/TotalsNumber.spec.tsx
```diff
@@ -29,6 +30,7 @@ describe('TotalsNumber', () => {
     })
 
     it('renders - when there is an invalid value', () => {
+      // @ts-expect-error
```
rohitvinnakota-codecov, line 33: Can we let value be undefined on the type in setup? Prefer avoiding heavy use of expect-error

src/shared/utils/determineProgressColor.spec.ts
ajay-sentry: this was for the sake of testing a non-number val, I can swap to null though based off the other changes

src/services/repo/useRepoBackfilled.ts
ajay-sentry: good catch, updated

rohitvinnakota-codecov (approved):

PR #2665, opened 2024-03-07
feat: Add metrics for bundle analysis

nicholas-codecov:
# Description

This PR adds in a type-safe wrapper function for Sentry Metrics that allows us to dynamically define metric name strings via a set of types for different metric types. All you need to do is modify the types, and the metrics helper function will automatically have the metric name string for you. As well in this PR I'm also adding in some basic metrics for different parts of bundle analysis such as onboarding and expanding dropdowns.

The reason why I've gone with this approach of wrapping Sentry's metric utility is because it for two primary reasons, dynamically creating template strings for consistency, as well as using plain strings instead of complex objects for better GZIP results. On the first point, if we created a big object with many fields and enums, we would have to manage all of the strings ourselves potentially introducing small typos between similar strings. With the current implementation all parent fields are joined together with a `.` (following Sentry metric guidelines). The second reason is for better optimizations with GZIP, using plain strings GZIP can better optimize compression of the assets we send to the user.

Closes codecov/engineering-team#998

# Notable Changes

- Create `metrics` wrapper for Sentry metrics in `shared/utils/metrics.ts`
- Add in bundle onboarding metrics
- Add metrics around different bundle aspects in different components


src/shared/utils/metrics.ts
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable camelcase */
+import { metrics as sentryMetrics } from '@sentry/react'
+
+type LeafKeyUnion<T, K extends string = ''> = {
```
rohitvinnakota-codecov, line 4: Can we call this `MetricKeyNameUnion` or something similar so it's more descriptive?

src/shared/utils/metrics.ts
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }
+}
+
+type SetKeys = {
+  // TODO: Remove this key when adding in non-test keys
+  testKey: string
+}
+
+type DistributionValue = Parameters<typeof sentryMetrics.distribution>['1']
```
rohitvinnakota-codecov, line 74: curious as to why we are using strings as indices for collections?

src/pages/CommitDetailPage/CommitCoverage/CommitCoverage.jsx
```diff
@@ -153,17 +152,23 @@ function CommitCoverageRoutes() {
...
   })
 
-  const showCommitSummary = !(
-    repoData?.repository?.private && tierName === TierNames.TEAM
-  )
+  useEffect(() => {
+    if (overview.bundleAnalysisEnabled && overview.coverageEnabled) {
+      metrics.increment('commit_detail_page.coverage_dropdown.opened', 1)
+    } else if (overview.coverageEnabled) {
+      metrics.increment('commit_detail_page.coverage_page.visited_page', 1)
```
rohitvinnakota-codecov, line 167: Sentry metrics increments by 1 by default. Maybe we can do the same in our abstraction?

src/pages/CommitDetailPage/CommitCoverage/CommitCoverage.spec.jsx
```diff
@@ -615,4 +649,44 @@ describe('CommitCoverage', () => {
...
+            'commit_detail_page.coverage_page.visited_page',
+            1,
+            undefined
+          )
+        )
+      })
+    })
+
+    describe('when coverage and bundle analysis are enabled', () => {
+      it('send correct metrics', async () => {
```
rohitvinnakota-codecov, line 674: hyper nit: `sends`

src/pages/CommitDetailPage/CommitBundleAnalysis/CommitBundleAnalysis.tsx
```diff
@@ -59,13 +61,22 @@ const BundleContent: React.FC<BundleContentProps> = ({ bundleCompareType }) => {
...
   const { data: commitPageData } = useCommitPageData({
     provider,
     owner,
     repo,
     commitId: commitSha,
   })
 
+  useEffect(() => {
+    if (overview?.bundleAnalysisEnabled && overview?.coverageEnabled) {
+      metrics.increment('commit_detail_page.bundle_dropdown.opened', 1)
```
rohitvinnakota-codecov, line 74: Does the user actually open a dropdown in this case? It looks like we are just checking repo overview to see if they have features enabled.

src/shared/utils/metrics.ts
nicholas-codecov: The type doesn't really matter in this instance, we're only looking at the key.

src/pages/CommitDetailPage/CommitBundleAnalysis/CommitBundleAnalysis.tsx
nicholas-codecov: So if both of these are true, and this component is rendered we know for a fact that the dropdown was opened. There's not real great way that the dropdown exposes itself to check and see if the dropdown was opened or closed.

src/pages/CommitDetailPage/CommitCoverage/CommitCoverage.jsx
nicholas-codecov: I'd rather leave the wrapper to be as close to the Sentry metric implementation, incase anyone reads their docs it will match what they're reading (minus the limited set of names)

src/shared/utils/metrics.ts
nicholas-codecov: Oh sorry, if you're talking about for grabbing the value from the `Parameters` TS utility, it's just the way you do it in TS it seems

src/shared/utils/metrics.ts
nicholas-codecov: Ehhhhhhhhh yea I guess that's okay

rohitvinnakota-codecov (approved):

PR #2666, opened 2024-03-07
fix: update discount schema with proper zod object

ajay-sentry:
# Description

updates the discount schema used in the `account-details` rest endpoint with the proper keys

# Code Example

# Notable Changes

N/A, copied directly from the serializer here: https://github.com/codecov/codecov-api/blob/daacd21fe6982c5f53f941669299dbe086c58a75/api/internal/owner/serializers.py#L83-L102

Additionally updated the mocks for the UT to account for this update

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


nicholas-codecov (approved):
Alrighty, send it

PR #2668, opened 2024-03-07
feat: add Sentry metrics to network error boundary

nicholas-codecov:
# Description

After todays eng retro I threw together this quick PR so we can start collecting some basic metrics from the network error boundary so we can setup some alerts for when things catch on fire.

# Notable Changes

- Add in new metrics keys to `metrics.ts`
- Add in helper function
- Update `NetworkErrorBoundary` to send the metrics

### DIFF
--- a/src/layouts/shared/NetworkErrorBoundary/networkErrorMetrics.ts
+++ b/src/layouts/shared/NetworkErrorBoundary/networkErrorMetrics.ts
@@ -1,0 +1,47 @@
+import { metrics } from 'shared/utils/metrics'
+
+type SupportNetworkStatusCodes = 401 | 403 | 404 | 500
+
+export const sendNetworkErrorMetrics = (status: SupportNetworkStatusCodes) => {
+  if (status === 401) {
+    metrics.increment('network_errors.401', 1)
+    return
+  }
+
+  if (status === 403) {
+    metrics.increment('network_errors.403', 1)
+    return
+  }
+
+  if (status === 404) {
+    metrics.increment('network_errors.404', 1)
+    return
+  }
+
+  if (status === 500) {
+    metrics.increment('network_errors.500', 1)
+    return
+  }
+}
+
+type GraphQLErrorCodes =
+  | 'UnauthenticatedError'
+  | 'UnauthorizedError'
+  | 'NotFoundError'
+
+export const sendGraphQLErrorMetrics = (status: GraphQLErrorCodes) => {
+  if (status === 'UnauthenticatedError') {
+    metrics.increment('network_errors.unauthenticated_error', 1)
+    return
+  }
+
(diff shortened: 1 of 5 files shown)


src/layouts/shared/NetworkErrorBoundary/networkErrorMetrics.ts
```diff
@@ -0,0 +1,47 @@
+import { metrics } from 'shared/utils/metrics'
+
+type SupportNetworkStatusCodes = 401 | 403 | 404 | 500
+
+export const sendNetworkErrorMetrics = (status: SupportNetworkStatusCodes) => {
+  if (status === 401) {
```
ajay-sentry, line 6: idea: I think we can simplify this code quite a bit (and future-proof ourselves) with something like

return metrics.increment('network_errors.${status}`,1)

src/layouts/shared/NetworkErrorBoundary/networkErrorMetrics.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+type GraphQLErrorCodes =
+  | 'UnauthenticatedError'
+  | 'UnauthorizedError'
+  | 'NotFoundError'
+
+export const sendGraphQLErrorMetrics = (status: GraphQLErrorCodes) => {
+  if (status === 'UnauthenticatedError') {
+    metrics.increment('network_errors.unauthenticated_error', 1)
```
ajay-sentry, line 34: idea: what are your thoughts about explicitly naming these graphql_errors vs. network errors, for easier parsing later on?

nicholas-codecov: I like

src/layouts/shared/NetworkErrorBoundary/networkErrorMetrics.ts
nicholas-codecov: Hrrm ... this has some type-safety implications ...

nicholas-codecov: currently if a network error code doesn't fall into line with any of this ... it does get lifted to a different error boundary

ajay-sentry: hmmmmm, yeah I see what you mean.
I think the use case I was thinking was lets say someone accidentally forgets to add a conditional for a new error code we get it for "free" without having to do the extra leg work.

We could also use a switch case here with a default case to achieve the same if you'd be open to that?

nicholas-codecov: We sorta already have a catch all here

ajay-sentry (approved):

PR #2686, opened 2024-03-12
ref: TS conversions for more flag table innards

ajay-sentry:
# Description

Another set of TS conversions for the flag table inner components. I wanted to split this out as a separate chunk so when I do the actual table refactor it's fairly isolated with changes. 

This set of changes **does not** touch any of the hooks, just components.

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/b51d4d6b-713d-4a3a-b863-64a9b7ad3d71

# Link to Sample Entry

# Legal Boilerplate


spalmurray-codecov (approved):
lgtm

I never thought I'd see a PR title with the word 'innards' in it, but here we are 😆

PR #2688, opened 2024-03-12
ref: Convert useRepo to TS

ajay-sentry:
# Description

Converts the useRepos hook to TS and updates the respective test mocks accordingly

Used the graphQL schemas from codecov-api as a guide.
https://github.com/codecov/codecov-api/blob/06b40b804131fef5a6962522d55f0e4baf1977b4/graphql_api/types/owner/owner.graphql and https://github.com/codecov/codecov-api/blob/6be2a744f32168bd621f375164c0073671d6f340/graphql_api/types/repository/repository.graphql

# Manual Testing

- Ran through the app clicking around to confirm there were no side effects from improper zod schema type casting
- Confirmed UTs passed after mock updates

# Link to Sample Entry

# Legal Boilerplate


src/services/repo/useRepo.ts
```diff
@@ -0,0 +1,90 @@
...
+          yaml
+          activated
+          oldestCommitAt
+          active
+        }
+      }
+    }
+`
+
+export interface UseRepoArgs {
```
spalmurray-codecov, line 47: Don't think this needs to be exported

spalmurray-codecov (approved):
1 nit - looks good otherwise :)

src/services/repo/useRepo.ts
ajay-sentry: good point! Fixed in [e2bac12](https://github.com/codecov/gazebo/pull/2688/commits/e2bac126d6a254940ad10e6ec1b1b016777f7c66)

spalmurray-codecov (approved):

PR #2689, opened 2024-03-12
feat: Show upload count on team plan card

spalmurray-codecov:
# Description

Adds an upload count indicator to the team plan card in the current plan page. This was already a thing for free plans, so this just brings the team plan in line with that.

Closes https://github.com/codecov/internal-issues/issues/232

# Notable Changes

- Add upload count indicator to the team plan card
- Convert related stuff to ts

# Screenshots

Before:
Screenshot 2024-03-12 at 17 44 39

After:
Screenshot 2024-03-12 at 17 41 44


src/pages/PlanPage/hooks/usePlanPageData.ts
```diff
@@ -0,0 +1,61 @@
...
+  const { owner, provider } = useParams<URLParams>()
+  const query = `
+      query PlanPageData($username: String!) {
+        owner(username: $username) {
+          username
+          isCurrentUserPartOfOrg
+          numberOfUploads
+        }
+      }
+    `
```
nicholas-codecov, line 33: `l`: can we move this outside of the hook, that way we it's not being declared upon calling the hook, which is called every time a component re-renders.

src/pages/PlanPage/hooks/usePlanPageData.spec.tsx
```diff
@@ -15,7 +16,7 @@ const mockOwner = {
 const queryClient = new QueryClient({
   defaultOptions: { queries: { retry: false } },
 })
-const wrapper = ({ children }) => (
+const wrapper = ({ children }: { children: ReactNode }) => (
```
nicholas-codecov, line 19: `l`: `React.FC<React.PropsWithChildren>` 👀

nicholas-codecov (commented):
Two quick things to peak at

src/pages/PlanPage/hooks/usePlanPageData.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export function usePlanPageData() {
+  const { owner, provider } = useParams<URLParams>()
+
+  const variables = {
+    username: owner,
+  }
+
+  return useQuery({
+    queryKey: ['PlanPageData', variables, provider, query],
```
nicholas-codecov, line 41: You should be able to remove `query` from here, cache keys should only be variables that either declared in the hook, or passed via args.

src/pages/PlanPage/hooks/usePlanPageData.ts
```diff
@@ -0,0 +1,62 @@
...
+    owner(username: $username) {
+      username
+      isCurrentUserPartOfOrg
+      numberOfUploads
+    }
+  }
+`
+
+export function usePlanPageData() {
+  const { owner, provider } = useParams<URLParams>()
```
nicholas-codecov, line 34: I'm a little hesitant to use `useParams` in these hooks, and prefer to get this info via args. The reason behind that is when you pass these values as args you can require them through TS and ensure that you actually have the values. Whereas when you just call `useParams` in the hook, you can place it anywheres without realizing that you require these values to be used.

nicholas-codecov (commented):
Sorry two super quick things I missed during the last review

nicholas-codecov (approved):
LGTM

PR #2698, opened 2024-03-13
ref: Convert file entry stuff to ts

spalmurray-codecov:
Part of https://github.com/codecov/engineering-team/issues/588

PR #2699, opened 2024-03-13
ref: Convert pull dir entry stuff to ts

spalmurray-codecov:
part of https://github.com/codecov/engineering-team/issues/588

depends on https://github.com/codecov/gazebo/pull/2698


hootener:
## [Bundle](https://app.codecov.io/gh/codecov/gazebo/pull/2699?dropdown=bundle) Report

Changes will increase total bundle size by 2.76kB :arrow_up:

| gazebo-staging-array-push | 5.96MB | 2.76kB :arrow_up: |

src/services/pathContents/pull/dir/constants.ts
```diff
@@ -0,0 +1,148 @@
...
+import { UnknownFlagsSchema } from 'services/impactedFiles/schemas'
+
+const MissingCoverageSchema = z.object({
+  __typename: z.literal('MissingCoverage'),
+  message: z.string(),
+})
+
+const UnknownPathSchema = z.object({
+  __typename: z.literal('UnknownPath'),
+  message: z.string(),
```
rohitvinnakota-codecov, line 13: Would recommend adding `nullish` here. had some issues locally where there was no message being sent with the error types.

src/services/pathContents/pull/dir/constants.ts
```diff
@@ -0,0 +1,148 @@
...
+
+const PathContentFileSchema = z.object({
+  __typename: z.literal('PathContentFile'),
+  name: z.string(),
+  path: z.string().nullable(),
+  hits: z.number(),
+  misses: z.number(),
+  partials: z.number(),
+  lines: z.number(),
+  percentCovered: z.number(),
```
rohitvinnakota-codecov, line 33: think all of these except name are nullable, unless you've verified they're not.

src/services/pathContents/pull/dir/constants.ts
```diff
@@ -0,0 +1,148 @@
...
+  isCriticalFile: z.boolean(),
+})
+
+export type PathContentFile = z.infer<typeof PathContentFileSchema>
+
+const PathContentDirSchema = z.object({
```
rohitvinnakota-codecov, line 39: This is also fine since it is more explicit. 

You could do something like [this](https://github.com/codecov/gazebo/blob/01f688c7cbf6f27017f8f02ab44c944111c3c328/src/services/pathContents/branch/dir/useRepoBranchContents.tsx#L19) and combine the definitions if you wanted in the future. Up to you.

src/shared/ContentsTable/TableEntries/PullEntries/PullDirEntry.spec.tsx
```diff
@@ -125,18 +137,20 @@ describe('PullDirEntry', () => {
     const dir = screen.getByText('dir')
     await user.hover(dir)
 
+    const queryKey = queryClient.getQueriesData({})?.at(0)?.at(0) as Array<any>
```
rohitvinnakota-codecov, line 140: Not really sure what's happening here but you can do as `Array<string>` and avoid any

src/services/pathContents/pull/dir/constants.ts
spalmurray-codecov: They are not nullable. Double checked API. Will fallback to 0.

rohitvinnakota-codecov (approved):

PR #2690, opened 2024-03-12
chore: Remove deprecated jest matchers for native aliases

ajay-sentry:
# Description

As of jest 29, aliased matchers like "toBeCalled" "toBeCalledTimes" etc. have been deprecated in favor of their native counterparts "toHaveBeenCalled" and others.

This PR aims to get rid of all deprecated matchers

[jest changelog](https://github.com/jestjs/jest/blob/main/CHANGELOG.md)
- At the top of the list you can see the breaking change
- Related issue: https://github.com/jestjs/jest/issues/13164

closes: https://github.com/codecov/engineering-team/issues/1404

# Manual Testing

Chore, UT's passing locally should suffice here since this only touches jest.

# Link to Sample Entry

# Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. In 2022 this entity acquired Codecov and as result Sentry is going to need some rights from me in order to utilize my contributions in this PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #2701, opened 2024-03-14
fix: Fix flaky test in useMyOrganizations.spec.tsx

spalmurray-codecov:
Seems like this was introduced by https://github.com/codecov/gazebo/pull/2690


ajay-sentry:
nice catch

PR #2703, opened 2024-03-14
ref: Convert Use Repo Flags Table to TS

ajay-sentry:
# Description

- refactors the useRepoFlagsTable hook used in the FlagsTable to TS. Aiming to convert this to aid with the table refactor for the FlagsTable
- Regression test screenshot below

# Screenshots

Screenshot 2024-03-14 at 3 25 24 PM

# Link to Sample Entry

# Legal Boilerplate


src/pages/RepoPage/FlagsTab/constants.ts
```diff
@@ -13,6 +13,7 @@ export const MEASUREMENT_TIME_INTERVALS = {
 } as const
 
 export const AFTER_DATE_FORMAT_OPTIONS = {
+  ALL_TIME: { months: 6 },
```
thomasrockhu-codecov, line 16: do we only have 6 months worth of data? or is this just wip

ajay-sentry: There's some logic in useMeasurementVariables where if we select "All Time" it uses the oldestCommitAt timestamp, so that 6 month value shouldn't matter ideally. I had to add _some_ value though in the event oldestCommitAt doesn't exist (it's typed out as potentially undefined or null so there's a chance), so I wanted to fall back to the next "highest" value which in this case was 6 months.

This is due to us keying off AFTER_DATE_FORMAT_OPTIONS if oldestCommitAt doesn't exist, where before it likely would've returned undefined, now we're returning a legitimate value.

I can def add a comment up here for clarity if you think it'd be worthwhile

src/services/navigation/useLocationParams.ts
```diff
@@ -1,18 +1,23 @@
...
 
-export function useLocationParams(defaultParams = {}) {
-  const { push } = useHistory()
-  const { pathname, search, state } = useLocation()
+interface LocationParams {
+  [key: string]: string | number | boolean | any[]
+}
+
+export function useLocationParams(defaultParams: LocationParams = {}) {
+  const { push }: { push: History['push'] } = useHistory()
```
nicholas-codecov, line 11: Is this really not typed from ReactRouter?!

src/services/navigation/useLocationParams.ts
```diff
@@ -1,18 +1,23 @@
+import { History, Location } from 'history'
 import omitBy from 'lodash/omitBy'
 import qs from 'qs'
 import { useHistory, useLocation } from 'react-router-dom'
 
-export function useLocationParams(defaultParams = {}) {
-  const { push } = useHistory()
-  const { pathname, search, state } = useLocation()
+interface LocationParams {
+  [key: string]: string | number | boolean | any[]
```
nicholas-codecov, line 7: There may potentially be an object being set somewhere in the app, we may want to account for that

nicholas-codecov (commented):
Few comments to take a peak at

PR #2705, opened 2024-03-15
fix: Fix team plan seat count error showing when it should not

spalmurray-codecov:
# Description

Removes the logic that resets the seat count back to 10 whenever something higher is entered, opting to show an error banner and disable the upgrade button instead.

Closes https://github.com/codecov/engineering-team/issues/1305

# Notable Changes

- Remove reset to 10 seats logic on team plan upgrade form


nicholas-codecov (approved):
Look's all good to me 👍

PR #2708, opened 2024-03-15
[ref] Refactor CodeTree and FileList table on repo page

rohitvinnakota-codecov:
# Description

Refactors the CodeTree and FileList table on the coverage tab and related hooks to use the new tanstack tables. 
`useRepoBranchContentsTable` and the underlying hook `useRepoBranchContents` are used to populate both tables. The tables are conditionally rendered depending on location params. 

https://github.com/codecov/gazebo/assets/148245014/261fcd5f-5f0a-4ce6-9feb-818c5930ef91

https://github.com/codecov/gazebo/assets/148245014/e69f4955-35ce-4a89-a44d-5377b0d44c8e

# Legal Boilerplate


src/pages/RepoPage/CoverageTab/subroute/FileExplorer/CodeTreeTable/CodeTreeTable.spec.tsx
```diff
@@ -123,12 +138,14 @@ const mockOverview = {
...
-  (initialEntries = '/gh/codecov/cool-repo/tree/main/a/b/c') =>
+  (
+    initialEntries = '/gh/codecov/cool-repo/tree/main/'
+  ): React.FC<React.PropsWithChildren> =>
   ({ children }) => {
     return (
       <QueryClientProvider client={queryClient}>
         <MemoryRouter initialEntries={[initialEntries]}>
-          <Route path="/:provider/:owner/:repo/tree/:branch/:path+">
+          <Route path="/:provider/:owner/:repo/tree/:branch/:path*">
```
nicholas-codecov, line 148: We probably want to leave `+` here to align with the paths defined in parent components. As well, `react-router` uses regex to match these paths, and we want the `+` here so that it requires at minimum "one" or more paths, whereas `*` will match 0 or more.

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/CodeTreeTable/CodeTreeTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+    return { direction, ordering }
+  }
+
+  return undefined
+}
+
+export const getBaseColumns = () => {
+  const baseColumns = [
+    columnHelper.accessor('name', {
+      id: 'name',
```
nicholas-codecov, line 60: Just a note for the future, no action required here, but with the column helper you shouldn't be required to assign an id as the helper should do it for you iirc.

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/CodeTreeTable/CodeTreeTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+      ordering = 'LINES'
+    }
+
+    return { direction, ordering }
+  }
+
+  return undefined
+}
+
+export const getBaseColumns = () => {
```
nicholas-codecov, line 57: This function doesn't seem to be doing anything special, can we remove it?

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FileExplorer.jsx
```diff
@@ -39,7 +39,7 @@ function FileExplorer() {
 
   return (
     <>
-      <div className="sticky top-[4.5rem] flex justify-between gap-2 bg-white pt-2">
+      <div className="sticky top-[4.5rem] z-50 flex justify-between gap-2 bg-white pt-2">
```
nicholas-codecov, line 42: Is this the lowest possible `z` index we can use?

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FileListTable/FileListTable.spec.tsx
```diff
@@ -114,12 +113,14 @@ const mockOverview = {
...
-  (initialEntries = '/gh/codecov/cool-repo/tree/main/a/b/c') =>
+  (
+    initialEntries = '/gh/codecov/cool-repo/tree/main/a/b/c'
+  ): React.FC<React.PropsWithChildren> =>
   ({ children }) => {
     return (
       <QueryClientProvider client={queryClient}>
         <MemoryRouter initialEntries={[initialEntries]}>
-          <Route path="/:provider/:owner/:repo/tree/:branch/:path+">
+          <Route path="/:provider/:owner/:repo/tree/:branch/:path*">
```
nicholas-codecov, line 123: Same thing here as previous comment about routes 👀

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FileListTable/FileListTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+      ordering = 'LINES'
+    }
+
+    return { direction, ordering }
+  }
+
+  return undefined
+}
+
+export const getBaseColumns = () => {
```
nicholas-codecov, line 58: Can we remove this function as well?

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FileListTable/FileListTable.tsx
```diff
@@ -0,0 +1,217 @@
...
+      cell: ({ renderValue }) => renderValue(),
+    }),
+  ]
+
+  return baseColumns
+}
+
+function FileListTable() {
+  const [sorting, setSorting] = useState([{ id: 'misses', desc: true }])
+  const ordering = useMemo(() => getOrderingDirection(sorting), [sorting])
```
nicholas-codecov, line 97: This is such a low cost function call you can probably skip wrapping it in a `useMemo` and just call it again on each re-render.

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FlagMultiSelect.jsx
```diff
@@ -65,10 +65,12 @@ function FlagMultiSelect() {
     }
   }
 
+  console.log(flagsMeasurementsActive, isTimescaleEnabled)
```
nicholas-codecov, line 68: 🚨 🚨 🚨

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FlagMultiSelect.jsx
```diff
@@ -65,10 +65,12 @@ function FlagMultiSelect() {
...
+
   return (
     <div className="w-full sm:w-60">
       <MultiSelect
-        disabled={!flagsMeasurementsActive || !isTimescaleEnabled}
+        disabled={false}
```
nicholas-codecov, line 73: This may be a leftover from some testing.

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/hooks/useRepoBranchContentsTable.tsx
```diff
@@ -0,0 +1,206 @@
...
+        if (result?.__typename === 'PathContentDir') {
+          name = (
+            <BranchDirEntry
+              name={result.name}
+              branch={branch}
+              urlPath={urlPath}
+            />
+          )
+        }
+        if (result?.__typename === 'PathContentFile') {
```
nicholas-codecov, line 132: Can we do an else if here, just to save on the extra computation if the first condition is already meet

src/shared/ContentsTable/utils/adjustListIfUpDir.tsx
```diff
@@ -15,13 +15,13 @@ type TreePath = {
...
-  lines: string
-  hits: string
-  misses: string
-  partials: string
-  coverage: ReactNode
+  lines?: string | number | null
+  hits?: string | number | null
+  misses?: string | number | null
+  partials?: string | number | null
+  coverage?: ReactNode | number | null
```
nicholas-codecov, line 24: You shouldn't need to add in `number` and `null` here for `coverage` as `ReactNode` should already account for those types.

src/shared/treePaths/useTreePaths.js
```diff
@@ -1,20 +1,24 @@
...
 }
 
 export function useTreePaths(passedPath) {
   const location = useLocation()
-  const params = qs.parse(location.search, {
-    ignoreQueryPrefix: true,
-    depth: 1,
-  })
+
+  const params = useMemo(() => {
```
nicholas-codecov, line 16: Is this memo required? if not can we remove it

src/shared/treePaths/useTreePaths.js
```diff
@@ -34,37 +38,39 @@ export function useTreePaths(passedPath) {
     { suspense: false }
   )
 
-  const branch = urlBranch && decodeURIComponent(urlBranch)
-  const ref = urlRef && decodeURIComponent(urlRef)
-  const path = urlPath && decodeURIComponent(urlPath)
-  const filePaths = getFilePathParts(passedPath || path)
-  const defaultBranch = repoOverview?.defaultBranch
+  const treePaths = useMemo(() => {
```
nicholas-codecov, line 41: Same with this memo, should refrain from using `useMemo` unless it's absolutely required to create a stable reference OR there are massive performance implications, and for this function there's not a lot of complexity imo that would have a major performance impact.

nicholas-codecov (changes requested):
Few comments to take a peak at. As well I ran into some issues with navigating around, it seems like a suspense somewhere isn't getting turned off correctly causing the whole header to disappear, as well with the suspense being activated, we're losing the benefits of prefetching the repo contents.

Current table:

https://github.com/codecov/gazebo/assets/105234307/542eb54e-8b84-40ce-bc93-22411509aedc

Updated table:

https://github.com/codecov/gazebo/assets/105234307/5c24b0d9-abff-4540-9d83-c43e287cf3c4

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/CodeTreeTable/CodeTreeTable.spec.tsx
rohitvinnakota-codecov: I think that is fine. The path may or may not be defined on the app as well. 
Example: https://app.codecov.io/gh/codecov/gazebo

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/CodeTreeTable/CodeTreeTable.tsx
rohitvinnakota-codecov: Prefer this to avoid over crowding the component. Can we send over the columns in lesser lines?

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FileListTable/FileListTable.spec.tsx
rohitvinnakota-codecov: Same response, the tests also don't work with the change.

src/shared/treePaths/useTreePaths.js
rohitvinnakota-codecov: Also required to prevent infinite re-renders. I spent many hours dealing with this guy.

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/CodeTreeTable/CodeTreeTable.tsx
nicholas-codecov: You don't need to declare it inside the component, you can just declare it above.

src/shared/treePaths/useTreePaths.js
nicholas-codecov: hrrm alrighty

nicholas-codecov (commented):
LGTM will approve once conflicts are resolved

Edit: sorry one quick fix

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/FileListTable/FileListTable.spec.tsx
spalmurray-codecov: I think it's fine with `*`. Testing on prod, `https://app.codecov.io/gh/codecov/gazebo/tree/main` does resolve properly to the root directory of the repo, so 0 path segments or more should be ok.

src/shared/ContentsTable/utils/adjustListIfUpDir.tsx
spalmurray-codecov: @rohitvinnakota-codecov comparing this to my stuff in the pull file explorer, it seems these are required because your zod schema states that lines, hits, misses, partials, and coverage are nullable. The graphql schema states these are all non-null and having looked on the API side, they seem to fall back to 0. Do you have an example of a PathContents query where these are null? If so I'll need to adjust my schema.

src/shared/utils/determineProgressColor.ts
```diff
@@ -6,17 +6,16 @@ export const determineProgressColor = ({
   lowerRange,
 }: {
   coverage: number | null
-  upperRange: number
-  lowerRange: number
+  upperRange?: number | null
```
spalmurray-codecov, line 9: IMO a better way to deal with null inputs is to fall back to codecov's default indication range rather than just `return 'primary'`. The default indication range is lowerRange = 60, upperRange = 80. Lmk what you think.

spalmurray-codecov: and coverage may be non-null based on what we decide in that other conversation

spalmurray-codecov (commented):
some comments lmk what you think

src/shared/ContentsTable/utils/adjustListIfUpDir.tsx
rohitvinnakota-codecov: You're right about the zod schema. updated.

spalmurray-codecov (approved):
Looks good!

nicholas-codecov (approved):
Looks good to me 👍

PR #2709, opened 2024-03-18
fix: Handle guest feature error, no nested catch

RulaKhaled:
# Description
- No need to throw and catch within the same function
- Redirect to `/login` on denied access error, following this change https://github.com/codecov/codecov-api/pull/458

# Link to Sample Entry

# Legal Boilerplate


src/shared/api/api.js
```diff
@@ -89,25 +89,27 @@ function graphql({
...
-        ? await res.json()
-        : Promise.reject({
-            status: res.status,
-            data: await res.json(),
-          })
-    )
-    .catch((err) => {
-      if (err.status === 401 && config.IS_SELF_HOSTED) {
+  }).then(async (res) => {
+    const data = await res.json()
```
nicholas-codecov, line 93: `h`: What happens if the call to `res.json()` throws? I believe that was also being handled by the catch.

nicholas-codecov (changes requested):
I don't think we should be removing the catch. It looks like it was also there to catch other potential rejections that aren't being handled anymore

src/shared/api/api.js
RulaKhaled: fair enough, we used to have: if res.ok => res.json(), otherwise throw an error, this was recently changed when i added the catch block.

src/shared/api/api.js
```diff
@@ -90,24 +90,30 @@ function graphql({
...
+      const data = await res.json()
+
+      if (data?.errors) {
+        if (
+          data?.errors?.[0]?.extensions?.status === 403 &&
+          config.IS_SELF_HOSTED
+        ) {
+          window.location.href = '/login'
+        }
+      }
```
nicholas-codecov, line 103: I just want to double confirm here, users who are logged in, but maybe not have access to something let's say the admin panel in self hosted (which users REST endpoints) won't get redirect to the login page right?

nicholas-codecov: Sorry I totally missed that this change is only occurring in the `graphqil` method 🤦

nicholas-codecov (approved):

PR #2710, opened 2024-03-18
feat: Improve component searching

spalmurray-codecov:
# Description

Improves the component searching experience on the coverage/pull/commit tabs by:
1. handling search in the browser
1. matching substrings

Obviously this is not as robust a search experience as is possible, but is an improvement over the current behaviour (see below)!

Closes https://github.com/codecov/feedback/issues/294

# Notable Changes

- Handle searching client side
- Match substrings instead of only exact matches.

# Screenshots

Before:
Recording 2024-03-18 at 14 16 47

After:
Recording 2024-03-18 at 14 07 57


src/pages/CommitDetailPage/CommitCoverage/routes/ComponentsSelector/ComponentsSelector.tsx
```diff
@@ -19,39 +19,41 @@ function ComponentsSelector() {
...
       suspense: false,
     },
   })
 
-  const components = data?.components
+  const components = useMemo(() => {
+    return data?.components
+  }, [data])
+
+  const componentNames = useMemo(() => {
```
ajay-sentry, line 31: Just curious if you already tried out the regExp class and then regExp.test('',param2) against it. I haven't on the Frontend personally but maybe it'd be another slightly more robust option that isn't too much of a pain to implement

ajay-sentry: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test

ajay-sentry (approved):
looks good!

PR #2712, opened 2024-03-18
feat: support setting `GH_APP` in docker

### LINKED ISSUE #2711: Set `GH_APP` similar to other environment vars in `start-nginx.sh`
# 📣 Feedback / 🐛 Bugs

When setting up the codecov frontend, `GH_APP` isn't easy to specify.  This means that users with non-`codecov` github apps won't be able to add new organizations.

https://github.com/codecov/self-hosted/issues/22

Either we can do the string replace patching in `start-nginx` somehow (A little tricker since `codecov` as a string is pretty ubiquitous) or find a way to inject `window.configEnv` during `start-nginx`.

Is there a preferable path for anyone else?

imnotjames:
# Description

adds support for `CODECOV_GITHUB_APP` docker environment variable

fixes #2711

# Code Example

```
docker -e CODECOV_GITHUB_APP=codecov-example some-image-reference
```

# Notable Changes
When setting the `CODECOV_GITHUB_APP`, the github app linked out to when adding organizations is correct

# Screenshots
N/A

# Link to Sample Entry
N/A

# Legal Boilerplate

### DIFF
--- a/docker/start-nginx.sh
+++ b/docker/start-nginx.sh
@@ -16,6 +16,7 @@ then
   GHE_SCHEME_BASE=${CODECOV_GHE_SCHEME:=https}
   GLE_SCHEME_BASE=${CODECOV_GLE_SCHEME:=https}
   BBS_SCHEME_BASE=${CODECOV_BBS_SCHEME:=https}
+  GITHUB_APP_BASE=${CODECOV_GITHUB_APP_SEARCH:=codecov}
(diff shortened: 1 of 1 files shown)


rfox-codecov (approved):
LGTM textually but what's GH_APP for?

imnotjames:
It sets the name of the app when you're connecting a new repository.  Without this set, you'll be directed to connecting codecov which is not the GitHub app for a self hosted instance

PR #2714, opened 2024-03-18
ref: #597 - Flags Table Refactor Part 3

ajay-sentry:
# Description

This PR covers refactoring the FlagsTable from the old-ui table to the new pattern. One tricky thing with this particular table was it has a nested table, and after working with @nicholas-codecov, as well as spending ~3 days myself triaging some ghost rerendering issues, we were able to workaround it via Memoizing the table creation using `const Table = memo(Component)`

~~Outside of this, I removed the sorting mechanism within the useRepoFlagsTable hook since I _think_ we get it for free without needing to do the additional network calls via the inline sorting functionality from the tanstack table. At least, when working locally that looked to be the case.~~
- This change set has been reverted in [fbd6d15](https://github.com/codecov/gazebo/pull/2714/commits/fbd6d1522f2fcd1ba7be7bb0ba169bf5601491c2), as well as adding infinite scroll functionality

This file along with the spec has also been converted to TS, with the tableData type being the only missing type, until we find some time to go back and refactor that one which will give us the tableData type for free.

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/ba1476a9-2686-43a9-9fe3-61dd0325211d

# Link to Sample Entry

# Legal Boilerplate


src/pages/RepoPage/FlagsTab/subroute/FlagsTable/FlagsTable.spec.tsx
```diff
@@ -43,20 +43,28 @@ const mockFlagMeasurements = {
         edges: [
           {
             node: {
-              name: 'flag1',
+              name: 'flagA',
```
ajay-sentry, line 46: renamed these to test the sorting functionality

src/pages/RepoPage/FlagsTab/subroute/FlagsTable/hooks/useRepoFlagsTable.js
```diff
@@ -11,12 +10,7 @@ import { useLocationParams } from 'services/navigation'
 import { useRepo } from 'services/repo'
 import { useRepoFlags } from 'services/repo/useRepoFlags'
 
-const getSortByDirection = (sortBy) =>
-  sortBy.length > 0 && sortBy[0].desc
-    ? SortingDirection.DESC
-    : SortingDirection.ASC
-
-const useMeasurementVariables = (historicalTrend, oldestCommitAt) => {
+const createMeasurementVariables = (historicalTrend, oldestCommitAt) => {
```
ajay-sentry, line 13: renamed from "use" since it's not actually a hook

src/ui/Sparkline/sparkline.css
```diff
@@ -1,5 +1,24 @@
-td {
+.tableui table {
```
ajay-sentry, line 1: Some style overrides I needed to do in order to get the styles to play nicely for the nested table

src/pages/RepoPage/FlagsTab/subroute/FlagsTable/FlagsTable.tsx
```diff
@@ -0,0 +1,319 @@
...
+import useRepoFlagsTable from './hooks'
+import TableSparkline from './TableEntries/TableSparkline'
+
+import 'ui/Table/Table.css'
+
+interface FlagsTableHelper {
+  name: JSX.Element
+  coverage: JSX.Element
+  trend: JSX.Element
+  delete: JSX.Element | null
```
nicholas-codecov, line 33: Can you swap these to `React.ReactNode` as `JSX.Element` does not allow valid returns such as `null`, `number`, etc.

nicholas-codecov (commented):
One thing to look at here, and the previous comment

nicholas-codecov (approved):
Awesome work, thanks for those changes!

PR #2724, opened 2024-03-22
ref: Update Self Hosted Admin Members Table

nicholas-codecov:
# Description

This PR updates the members table that lives in the admin settings when running in self hosted mode, as well as refactoring some of the pieces over to TS.

Closes codecov/engineering-team#1123

# Notable Changes

- Update `useSelfHostedUserList` to TS
- Update `useSelfHostedSettings` to TS
- Update `MembersTable` to TS and new Table setup
- Update `AutoActivateMembers` tests

# Screenshots

Screenshot 2024-03-22 at 07 34 20


src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+  type: string
+  email: React.ReactNode
+  activationStatus: React.ReactNode
+}
+
+const columnHelper = createColumnHelper<MembersColumns>()
+
+const columns = [
+  columnHelper.accessor('username', {
+    header: 'User Name',
```
ajay-sentry, line 40: While we're here, should we update this to Username? there's one more copy reference in AdminTable.jsx too. 

Thought here is github calls it a Username, we should keep the same semantics

src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+      const disableToggle = maxSeats && !activated
+
+      return {
+        username: <p>{name || username}</p>,
+        type: isAdmin ? 'Admin' : 'Developer',
+        email: <p>{email}</p>,
+        activationStatus: (
+          <Toggle
+            dataMarketing="handle-members-activation"
+            label={activated ? 'Activated' : 'Non-Active'}
```
ajay-sentry, line 81: Kinda similar thing here with Non-Active vs. Inactive vs. Not Activated. Maybe something to bring up to design / prod

src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.tsx
```diff
@@ -0,0 +1,237 @@
...
+  })
+
+  useEffect(() => {
+    if (inView && hasNextPage) {
+      fetchNextPage()
+    }
+  }, [fetchNextPage, inView, hasNextPage])
+
+  const tableContent = useMemo(() => {
+    if (!data || data.length === 0 || !seatData) {
```
ajay-sentry, line 146: nit: I think the first 2 clauses can be combined into !data?.length, assuming that doesn't break a TS type

src/services/selfHosted/useSelfHostedUserList.spec.tsx
```diff
@@ -192,4 +193,25 @@ describe('useSelfHostedUserList', () => {
       })
     })
   })
+
+  describe('endpoint returns invalid data', () => {
+    beforeEach(() => {
```
ajay-sentry, line 198: jw, do we need this console.error() thing?

src/services/selfHosted/useSelfHostedUserList.ts
```diff
@@ -0,0 +1,82 @@
...
+  })
+
+  const memoedData = useMemo(
+    () => data?.pages?.flatMap((page) => page.results) ?? [],
+    [data]
+  )
+
+  return {
+    data: memoedData,
+    ...rest,
```
ajay-sentry, line 80: any chance we can get rid of the ...rest? Personally, think this would be the best opportunity to remove as much ambiguity as possible and from what I can see in other parts of the code base, we look to be destructuring isFetching, hasNextPage, fetchNextPage from there.

src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.tsx
nicholas-codecov: I believe @spalmurray-codecov was just updating the `AdminTable`, if he wants to make the update there 👀

src/services/selfHosted/useSelfHostedUserList.ts
nicholas-codecov: I prefer to keep this as is, as it makes using the hook feel very similar to when using a query directly, also if someone is going to access values from it, they don't need to change the hook at all, they already have all the fields that they need.

src/services/selfHosted/useSelfHostedUserList.spec.tsx
nicholas-codecov: We throw it in for these invalid responses because it will log out that there was an error which is expected, and we don't want to clutter the console output with false positives.

ajay-sentry: Got it, thanks for clarifying that

src/services/selfHosted/useSelfHostedUserList.ts
ajay-sentry: Yeah I gotcha there, makes sense. I can get behind that

ajay-sentry (approved):

PR #2726, opened 2024-03-22
ref: Refactor the hook to TS / accept other measurements types

RulaKhaled:
# Description
Refactor useActivateFlagMeasurements to be generic that takes in measurement type 

closes: https://github.com/codecov/engineering-team/issues/1374

# Link to Sample Entry

# Legal Boilerplate


src/services/repo/useActivateFlagMeasurements.tsx
```diff
@@ -0,0 +1,111 @@
...
+export const MEASUREMENT_TYPE = {
+  FLAG_COVERAGE: 'FLAG_COVERAGE',
+  COMPONENT_COVERAGE: 'COMPONENT_COVERAGE',
+} as const
+
+interface ActivateFlagMeasurementsProps {
+  provider: string
+  owner: string
+  repo: string
+  measurementType: (typeof MEASUREMENT_TYPE)[keyof typeof MEASUREMENT_TYPE]
```
ajay-sentry, line 17: nit: maybe we could pull this out into a type separately as `export type MeasurementType = keyof typeof MEASUREMENT_TYPE` so if we need to use it later we can. Also aids a bit with readability here imo

src/services/repo/useActivateFlagMeasurements.tsx
```diff
@@ -0,0 +1,111 @@
...
+export function useActivateFlagMeasurements({
+  provider,
+  owner,
+  repo,
+  measurementType,
+}: ActivateFlagMeasurementsProps) {
+  const queryClient = useQueryClient()
+  return useMutation({
+    mutationFn: () => {
+      const query = `
```
ajay-sentry, line 46: Nit: maybe we can pull the query const out of the hook so it's only declared the one time rather than every iteration of the hook?

src/services/repo/useActivateFlagMeasurements.tsx
```diff
@@ -0,0 +1,111 @@
...
+        return Promise.reject({
+          status: 404,
+          data: {},
+          dev: 'useActivateFlagMeasurements - 404 failed to parse',
+        } satisfies NetworkErrorObject)
+      }
+
+      const error = parsedData.data.activateMeasurements?.error
+      if (error) {
+        throw new Error(error.message)
```
ajay-sentry, line 89: Just want to confirm that we want the user to see this error message client-side

src/services/repo/hooks.spec.js
```diff
@@ -272,7 +273,12 @@ describe('useActivateFlagMeasurements', () => {
   function setup() {
     server.use(
       graphql.mutation('ActivateMeasurements', (req, res, ctx) => {
-        return res(ctx.status(200), ctx.data())
+        return res(
```
ajay-sentry, line 276: Could we add a test case for when this 404's if we don't have one already?

src/services/repo/useActivateFlagMeasurements.tsx
RulaKhaled: we do, if the hook failed we would want them to know it's the reason they're nor seeing flags measurements

src/services/repo/useActivateFlagMeasurements.tsx
RulaKhaled: why not

src/services/repo/hooks.spec.js
RulaKhaled: yah it's missing some coverage 😿

src/services/repo/useActivateFlagMeasurements.tsx
ajay-sentry: It's a slight perf optimization, essentially the overhead of defining once and using multiple times vs. defining + cleaning up at each iteration.

Another slight benefit would be readability, since we can condense the stuff within the hook to just be business logic vs. consts + business logic

src/services/repo/useActivateFlagMeasurements.tsx
ajay-sentry: Got it! Thanks for confirming

src/services/repo/useActivateFlagMeasurements.tsx
```diff
@@ -0,0 +1,111 @@
...
+        owner,
+        repo,
+      ])
+    },
+    onError: () => {
+      renderToast({
+        type: 'error',
+        title: 'Error activating flag measurements',
+        content:
+          'Please try again. If the error persists please contact support',
```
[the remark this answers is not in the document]
RulaKhaled: I agree, we can handle this separately though to cover other places, might as well refactor renderToastto accept html children instead of plain strings, maybe draft up a ticket for this?

src/services/repo/useActivateFlagMeasurements.tsx
RulaKhaled: Of course! i really meant "why not let's do it" 😆

ajay-sentry: oop 🙈

src/services/repo/useActivateFlagMeasurements.tsx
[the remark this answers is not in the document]
ajay-sentry: That's a good idea, let me post to product/design and see if they're already working on an answer for this

ajay-sentry (approved):
lgtm :)

PR #2729, opened 2024-03-25
ref: Refactor usRepoFlags to ts

RulaKhaled:
# Description
closes: https://github.com/codecov/gazebo/issues/2579

# Legal Boilerplate


src/services/repo/useRepoFlags.spec.tsx
```diff
@@ -14,7 +16,7 @@ const queryClient = new QueryClient({
   },
 })
 
-const wrapper = ({ children }) => (
+const wrapper = ({ children }: { children: React.ReactNode }) => (
```
spalmurray-codecov, line 19:
```suggestion
const wrapper: React.FC<React.PropsWithChildren> = { children }) => (
```

spalmurray-codecov (approved):
lgtm other than little type change suggested!

PR #2741, opened 2024-04-03
fix: Don't set default org if update default org is in progress 

RulaKhaled:
# Description
Upon sign up and when user selects personal, it then momentarily redirects to an org selector view. This gives user perception of missing part of flow or broken.

closes: https://github.com/codecov/engineering-team/issues/1385

# Notable Changes
i can't reproduce locally, so nothing visual changed, this is a pure attempt. 

# Link to Sample Entry

# Legal Boilerplate


ajay-sentry (approved):
let's try it!

PR #2743, opened 2024-04-04
fix: Handle case where branch or bundle is not selected

nicholas-codecov:
# Description

This PR adds in a new `InfoBanner` component that will render when there is no branch or bundle selected. There are some interesting cases where when initially setting up bundle analysis it seems like the branch entry for the default branch hasn't finished and we would end up with a "Unknown error" and they would just have to select the branch. As well this PR adds in an info banner to persuade the user to select a bundle if one has not already been selected.

Closes codecov/engineering-team#1409

# Notable Changes

- Add in new `InfoBanner` component
- Update `BranchSelector` to redirect to URL with branch set as the param when default branch is selected
- Update `BundleContent` to use the new `InfoBanner` component
- Fix up the text spacing in bundle tab `ErrorBanner`

# Screenshots

Screenshot 2024-04-04 at 10 22 59

Screenshot 2024-04-04 at 10 20 19


ajay-sentry (approved):
solid 💯

PR #2753, opened 2024-04-08
feat: Make email field required when submitting feedback

nicholas-codecov:
# Description

This PR makes it so that users are required to submit an email before submitting feedback through Sentry's feedback utility.


RulaKhaled (approved):

ryan953 (commented):
If people are logged into this site then `Sentry.setUser()` might be useful too, so folks do t have to fill out more fields. 

The feedback config docs link out to it:
https://docs.sentry.io/platforms/javascript/user-feedback/configuration/#user-and-form

PR #2755, opened 2024-04-08
feat: Invalidate queries on poll when syncing repos

ajay-sentry:
# Description

This originally stemmed from discussion with @adrian-codecov w.r.t. https://github.com/codecov/engineering-team/issues/1477. Refer to that link for some context on how sync repos currently works.

What we _want to do_ now though is give the user a sense that something is happening. We can trivially do this, at the cost of some extra GQL requests to our BE, by invalidating the query for fetching repos at the current polling interval.

This PR accomplishes the above by simply adding a query invalidation at the time of getting the current sync status. If we find that we are making too many requests to the BE, we can always modify the polling interval but for now it will effectively be `# requests = polling interval * number of polls`

There is also an optimization built in to prevent excess query invalidates by checking the current repo number in the cache and comparing it to our page size. We only want to do the invalidation if the number of repos is less than our page size!

**Benefits:**
- upserts happen in separate transactions, so each poll / refetch will generate some new repos
- we can monitor this change separate from turning on the generator 
- The polling interval is trivial enough to modify ad-hoc should we need to
- **Users receive some sense of feedback that something is happening**
  - We already have infinite scroll baked into this table as well

# Code Example

# Notable Changes

When we are polling for the current sync status, we are now also invalidating the "repos" query and effectively refetching them.

# Screenshots

**Some of the testing weirdness I was seeing. Even though the numbers are "off" the behavior looks to be correct**
Screenshot 2024-04-08 at 9 32 57 AM
Screenshot 2024-04-08 at 9 33 09 AM

# Link to Sample Entry

# Legal Boilerplate


adrian-codecov (approved):
lgtm!

PR #2757, opened 2024-04-08
fix: Reflect new components schema

RulaKhaled:
# Description
We're adding termId for the search value for components, also rename to components yaml as it's a duplicate name under repository query. 

# Link to Sample Entry

# Legal Boilerplate


src/services/repo/useRepoComponentsSelect.spec.tsx
```diff
@@ -34,7 +34,7 @@ const dataReturned = {
   owner: {
     repository: {
       __typename: 'Repository',
-      components: [
+      componentsYaml: [
```
JerrySentry, line 37: should add the `id` in the mock as well right?

JerrySentry (approved):

src/services/repo/useRepoComponentsSelect.spec.tsx
RulaKhaled: yes! good catch

PR #2773, opened 2024-04-12
fix: Not showing select bundle when there is no bundle selected

nicholas-codecov:
# Description

Currently missing the case where there's no bundle selected and we get a bad response from the API, this PR solves that by adding in a new check for those cases.

# Notable Changes

- Update conditional rendering of the `BundleContent` component


hootener:
## [Bundle](https://stage-app.codecov.dev/gh/codecov/gazebo/pull/2773?dropdown=bundle) Report

Changes will increase total bundle size by 3.48kB :arrow_up:

| gazebo-staging-array-push | 6.46MB | 3.48kB :arrow_up: |

spalmurray-codecov (approved):
👍

PR #2780, opened 2024-04-17
chore: Update Sentry to V8 Beta

nicholas-codecov:
# Description

As Sentry helps us dogfood our beta products, we should reciprocate and help dog food theirs. After chatting w/@AbhiPrasad the current beta is stable enough to make this move, and no planned breaking changes will be made between now, and the full release of V8. 

This PR updates the React SDK up to beta 1 of their V8 release. This update allows us to remove some of the packages that are now included magically inside of the SDK reducing the amount of deps we have. Also included in this PR are some changes around the metrics being used in the `UpdateButton` component on the plan page. As well as some TS file conversions.

Closes codecov/engineering-team#1580

# Notable Changes

- Update Sentry to V8 Beta 1
- Update `src/index.js` to `src/index.tsx`
- Update Sentry Config
- Use internal metrics helper for `UpdateButton`


hootener:
## [Bundle](https://stage-app.codecov.dev/gh/codecov/gazebo/pull/2780?dropdown=bundle) Report

Changes will increase total bundle size by 102.31kB :arrow_up:

| gazebo-staging-array-push | 6.57MB | 102.31kB :arrow_up: |

AbhiPrasad (approved):
Thank you!

Migration guide for anyone curious: https://github.com/getsentry/sentry-javascript/blob/develop/MIGRATION.md

The bundle size increase is a little worrying, need to investigate that further.

src/pages/PlanPage/PlanMetrics/planMetrics.ts
```diff
@@ -1,74 +1,54 @@
...
+    })
   }
 
   if (isSamePlan && isTeamPlan(newPlan)) {
-    updateGaugeMetric(
-      TEAM_SEATS_ADDED_AND_REMOVED_METRIC_KEY,
-      seatDelta,
-      ownerId
-    )
+    metrics.gauge('billing_change.user.seats_change', seatDelta, {
```
ajay-sentry, line 38: Curious why we moved away from the shared consts to the magic strings?

src/shared/utils/metrics.ts
```diff
@@ -113,7 +129,7 @@ export const metrics = {
   },
   increment: (
     name: MetricKeyNameUnion<IncrementKeys>,
-    value: IncrementValue,
+    value?: IncrementValue,
```
ajay-sentry, line 132: also curious about this change, what is the default value when none is set?

src/pages/PlanPage/PlanMetrics/planMetrics.ts
nicholas-codecov: The string isn't magic, the first argument is a union type of strings so you can have invalid values, and are available globally unlike the consts which were local to this file. Also with using these union types we're also actually removing a bunch of stuff during the runtime because the types are stripped out during TS compilation. And as an added benefit, straight up strings like this are far better at being gizp'ed then other things.

ajay-sentry (approved):

PR #2797, opened 2024-04-22
sec: 390 - Add validation for potential XSS vuln

ajay-sentry:
# Description

This ticket came from the Sentry pen test where it was found this function exposed a potential vulnerability for XSS where a user could end up going to a malicious URL via lack of validation on the "provider" variable coming into the endpoint.

This is definitely true, as we never validated to see if the provider entered was a valid one prior to appending it to our request URL to be called by the fetch API.

IMO, the simplest and most straight forward approach for this issue was to add validation using a pre-existing "isProvider" helper function that will check to see if the provider passed in is equivalent to the provider given by the `Api.graphQL()` function. If it _is_ a provider, we append it, else we use the base URL (with no additional vars passed in).

Seemingly, this also allowed us to remove another condition added ~6 months back here: https://github.com/codecov/gazebo/pull/2321, which added an additional variable passed from the TermsOfService component called "supportsServerless" which would "chunk down" the provider URL if supportsServerless was true AND there was no provider. I will need to double check with @RulaKhaled though here.

Also added a bunch of UT's for this new validation case.

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/shared/api/helpers.ts
```diff
@@ -13,7 +13,7 @@ export interface NetworkErrorObject {
   dev: `${string} - ${number} ${string}`
 }
 
-const AllProvidersArray = [
+export const AllProvidersArray = [
```
RulaKhaled, line 16: should we take care of capital letters as well? food for thought

RulaKhaled (approved):
looks good, thanks!

PR #2799, opened 2024-04-23
Implement Card component

spalmurray-codecov:
First draft- looking for some feedback!

[Design](https://www.figma.com/file/JOS1ZW499rfWP9nT7ivjf8/Pavilion%3A-Codecov-Design-System-Library?type=design&node-id=4318-3234&mode=design&t=2OTkMmq90kms8llt-0)

The code was copied from [shadcn/ui](https://ui.shadcn.com/docs/components/card) and I added CVA, implemented our styles, and adjusted some types.
I've used CVA across the board because it gives us class overriding via the className prop for free, starting with things like this also makes it really easy to add variants in the future, should the need arise.

I've overwritten the existing card component, but still need to go back and check that old instances of it look fine with the new styles/api.

Storybook screenshot:
Screenshot 2024-04-23 at 10 05 28

Closes https://github.com/codecov/engineering-team/issues/1456

### DIFF
--- a/src/ui/Card/Card.stories.tsx
+++ b/src/ui/Card/Card.stories.tsx
@@ -1,13 +1,6 @@
 import { Meta, StoryObj } from '@storybook/react'
 
-import {
-  Card,
-  CardContent,
-  CardDescription,
-  CardFooter,
-  CardHeader,
-  CardTitle,
-} from './Card'
+import Card from './Card'
 
 const meta: Meta<typeof Card> = {
   title: 'Components/Card',
@@ -20,10 +13,10 @@ type Story = StoryObj<typeof Card>
 export const Default: Story = {
   render: () => (
     <Card>
-      <CardContent>
+      <Card.Content>
         Here is the Card component, you are free to render anything you would
         like here.
-      </CardContent>
+      </Card.Content>
     </Card>
   ),
 }
@@ -31,13 +24,13 @@ export const Default: Story = {
 export const CardWithHeader: Story = {
   render: () => (
     <Card>
-      <CardHeader>
-        <CardTitle>A header can have a title.</CardTitle>
-        <CardDescription>And it can have a description.</CardDescription>
-      </CardHeader>
-      <CardContent>
+      <Card.Header>
+        <Card.Title>A header can have a title.</Card.Title>
+        <Card.Description>And it can have a description.</Card.Description>
+      </Card.Header>
+      <Card.Content>
         The header will place a border between it and the main Card content.
-      </CardContent>
+      </Card.Content>
     </Card>
   ),
 }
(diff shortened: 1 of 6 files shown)


src/ui/Card/Card.tsx
```diff
@@ -0,0 +1,79 @@
+import { cva, VariantProps } from 'cva'
+import React from 'react'
+
+const card = cva(['border border-ds-gray-secondary'])
+interface CardProps
+  extends React.HTMLAttributes<HTMLDivElement>,
+    VariantProps<typeof card> {}
+
+const Card = React.forwardRef<HTMLDivElement, CardProps>(
```
nicholas-codecov, line 9: Swap this to `CardRoot` from `Card`

```suggestion
const CardRoot = React.forwardRef<HTMLDivElement, CardProps>(
```

spalmurray-codecov (commented):
Export as const not default

src/ui/Card/Card.tsx
```diff
@@ -0,0 +1,94 @@
...
+    VariantProps<typeof title> {}
+
+const Title = React.forwardRef<HTMLParagraphElement, TitleProps>(
+  ({ className, size, children, ...props }, ref) => (
+    <h3 ref={ref} className={title({ className, size })} {...props}>
+      {children}
+    </h3>
+  )
+)
+Title.displayName = 'title'
```
spalmurray-codecov, line 50: Added a size variant bc my onboarding polish calls for a smaller title!

nicholas-codecov: Can you swap `large` for `lg`, makes it easier when we get into extra sizes `xl` etc.

spalmurray-codecov: trueeee. Was wondering whether lg made more sense, but your argument has convinced me!

src/ui/Card/Card.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+        <Card.Title size="base">A Smaller Title</Card.Title>
+      </Card.Header>
+      <Card.Content>
+        A smaller title can be used by setting the{' '}
+        <code className="bg-ds-gray-secondary">size</code> variant of{' '}
+        <code className="bg-ds-gray-secondary">Card.Title</code>.
+      </Card.Content>
+    </Card>
+  ),
+}
```
nicholas-codecov, line 52: I would recommend tweaking this story a bit, and allow users to play around with the different variants on their own so they don't have to jump between stories, you can see an example here: https://github.com/codecov/gazebo/blob/main/src/ui/TopBanner/TopBanner.stories.tsx#L31-L44

nicholas-codecov (approved):
Awesome work on this @spalmurray-codecov

PR #2800, opened 2024-04-23
Remove ui/Card to make room for new Card

spalmurray-codecov:
- **Refactor old cards into one (old_ui/Card)**
- **Remove ui/Card**


src/old_ui/Card/Card.jsx
```diff
@@ -1,9 +1,48 @@
 import cs from 'classnames'
+import PropTypes from 'prop-types'
```
spalmurray-codecov, line 2: The actual refactor went like this:

1. Copy ui/Card into old_ui/Card
2. Add an `old` variant with the styles from the old old_ui/Card
3. Handle className prop which the old old_ui/Card used
4. Update imports ui/Card -> old_ui/Card
5. Remove ui/Card

spalmurray-codecov: I.e., big green diff is not new code - just copied from ui/Card

nicholas-codecov: sure

nicholas-codecov: also it's blue =P

Screenshot 2024-04-24 at 13 07 05

nicholas-codecov (approved):
LGTM

PR #2802, opened 2024-04-23
[ref] Refactor commit detail file explorer table

rohitvinnakota-codecov:
Full refactor of the commit detail file explorer table and related hooks. 

Also updated the `CommitPathContents` query to use the new repository endpoint. 

Majority of the logic outside the table remains the same, and most changes are TS conversions. 

# Legal Boilerplate


spalmurray-codecov (approved):
Looks good! I think probably best that I reviewed this lol. Pretty much identical to the other two we did 😄

I tested locally as well, everything seems to be working as expected.

PR #2803, opened 2024-04-23
Fix MembersTable duplicate users

spalmurray-codecov:
Issue was that the page number was being manually set in the parent component, overriding the infinite query's automatic page handling.

Closes https://github.com/codecov/engineering-team/issues/1637

### DIFF
--- a/src/services/users/useInfiniteUser.tsx
+++ b/src/services/users/useInfiniteUser.tsx
@@ -36,7 +36,6 @@ export interface InfiniteUsersQuery {
     isAdmin?: boolean
     ordering?: string
     search?: string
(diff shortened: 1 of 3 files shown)


src/services/users/useInfiniteUser.tsx
```diff
@@ -57,8 +56,8 @@ export const useInfiniteUsers = (
         signal,
         query: {
           pageSize: 25,
-          page: pageParam,
           ...query,
+          page: pageParam,
```
spalmurray-codecov, line 60: Moving this so that even if the user manually sets the page number in the url, the infinite query will not use it.

PR #2805, opened 2024-04-24
Refactor GH Actions repo onboarding into one file

spalmurray-codecov:
Some clean up before bigger onboarding changes in https://github.com/codecov/engineering-team/issues/1437

Basically, we had two almost identical files for repo tokens and org tokens, this PR combines them into one file. Already with the split approach there were discrepancies between the two (actually using different versions of our codecov uploader github action 😬) this will stop that from happening in the future.

Similar PRs for the other two onboarding options to come.


nicholas-codecov (approved):
Awesome work 👍

PR #2810, opened 2024-04-25
fix: use session if exists when doing initial redirect for user

ajay-sentry:
# Description

Previously, when a user would come to app.codecov.io or similar, we would route them to the login page more often than not. This was due to us having "naive" redirect logic that was dependent on the existence of a provider to fetch the current user, using the useCurrentUser hook. 

More often than not, this would not be the case, and a user would frustratingly end up on the login page even if we already had a session cookie set for them! 

My solution then, was to use the useInternalUser() hook (thanks @JerrySentry for the pointer), which will pull the user's session if it exists **first** and use that to redirect the user. If we have a more specific route, or are able to derive the user's info via the provider and useCurrentUser hook, then we still do that.

There are also a couple other small fixes I baked in here:
- Converted to TS
- The test case that compared "search" values wasn't actually testing for that (Fixed now)
- We use the HomePageRedirect component instead of '/login' for the '/' path
- We have validation around the provider param being used now (similar to the fix in https://github.com/codecov/gazebo/pull/2797)

Below you can see the behavior before and after. In the before video, we see the sessionId cookie exists, but when we navigate to app.codecov.io we are directed back to the login page. Only when adding 'gh' since I am logged in using my github account am I redirected properly. In the after video there is so such issue :) 

# Screenshots

**BEFORE**

https://github.com/codecov/gazebo/assets/159853603/8ee29ae8-61c5-4933-8978-298071881963

**AFTER**

https://github.com/codecov/gazebo/assets/159853603/d3af86bd-97b9-4d1b-8579-bf984176f8f5

# Link to Sample Entry

# Legal Boilerplate


src/App.tsx
```diff
@@ -28,21 +29,34 @@ const PullRequestPage = lazy(() => import('./pages/PullRequestPage'))
...
   const { data: currentUser } = useUser()
+  const { data: internalUser } = useInternalUser({})
   const { params } = useLocationParams()
 
-  if (!provider || !currentUser) {
-    return <Redirect to="/login" />
+  let redirectURL = '/login'
+
+  if (internalUser) {
+    redirectURL = `/${internalUser.owners![0]?.service}/${
```
ajay-sentry, line 45: if internalUser exists, my assumption here is owners will at least be an empty array

src/App.spec.tsx
```diff
@@ -43,9 +44,9 @@ const internalUser = {
       integrationId: null,
       name: null,
       ownerid: 123,
-      service: 'github',
+      service: 'cool-service',
       stats: null,
-      username: 'codecov',
+      username: 'cool-guy',
```
spalmurray-codecov, line 49: 😎

src/App.tsx
spalmurray-codecov: I can't find any reason why that assumption would be bad. but just in case you didn't know how to optional chaining with array indexing (like I didn't until recently lol):
```suggestion
    redirectURL = `/${internalUser.owners?.[0]?.service}/${
```

I think that would be a safer play. Generally try to steer clear of `!` myself - it's scary 😰

src/App.tsx
```diff
@@ -28,21 +29,34 @@ const PullRequestPage = lazy(() => import('./pages/PullRequestPage'))
...
+      internalUser.owners![0]?.username
```
spalmurray-codecov, line 46: Same suggestion here
```suggestion
      internalUser.owners?.[0]?.username
```

spalmurray-codecov: You could also toss a lodash `isArray` in the if condition, if that's more your style

ajay-sentry: Yeah, my IDE didn't like that one unfortunately 😭 

What do you think about just changing the if statement to 

`  if (internalUser && internalUser.owners) {
    redirectURL = `/${internalUser.owners[0]?.service}/${internalUser.owners[0]?.username}`
  }`

ajay-sentry: Think its a _litttttttle_ cleaner that way?

spalmurray-codecov (changes requested):
Ran into this when testing locally:

Recording 2024-04-25 at 18 43 21

src/App.tsx
spalmurray-codecov: oh yea that's great too!

spalmurray-codecov (approved):

PR #2811, opened 2024-04-26
feat: Swap to avg and add in data source label

nicholas-codecov:
# Description

This PR swaps the data source on the analytics page from `max` to `avg` and adds in label informing users that the chart displays the average of all selected repos.

Closes codecov/engineering-team#1066

# Notable Changes

- Swap to fetching `avg` data
- Add in chat label

# Screenshots

Screenshot 2024-04-26 at 10 30 32

### DIFF
--- a/src/pages/AnalyticsPage/Chart/useCoverage.js
+++ b/src/pages/AnalyticsPage/Chart/useCoverage.js
@@ -28,14 +28,14 @@ export const useCoverage = ({ params, options = {} }) => {
     isPublic: shouldDisplayPublicReposOnly,
     opts: {
       select: (data) => {
-        if (data?.measurements?.[0]?.max === null) {
-          data.measurements[0].max = 0
+        if (data?.measurements?.[0]?.avg === null) {
+          data.measurements[0].avg = 0
         }
 
         // set prevPercent so we can reuse value if next value is null
         let prevPercent = data?.measurements?.[0]
         const coverage = data?.measurements?.map((measurement) => {
-          let coverage = measurement?.max ?? prevPercent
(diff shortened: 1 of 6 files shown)


src/pages/AnalyticsPage/Chart/useCoverage.js
```diff
@@ -28,14 +28,14 @@ export const useCoverage = ({ params, options = {} }) => {
...
+          data.measurements[0].avg = 0
         }
 
         // set prevPercent so we can reuse value if next value is null
         let prevPercent = data?.measurements?.[0]
         const coverage = data?.measurements?.map((measurement) => {
-          let coverage = measurement?.max ?? prevPercent
+          let coverage = measurement?.avg ?? prevPercent
 
           // can save on a few reassignments
```
spalmurray-codecov, line 40: Curious: is this actually significant? I've never though to optimize at this level before

spalmurray-codecov (approved):
Looks good. One question but not blocking

src/pages/AnalyticsPage/Chart/useCoverage.js
nicholas-codecov: It could potentially be significant. I'm not really sure how the API handles this, but I believe when we set to AVG the values we get back is the average of all the selected repos coverage, instead of just the max coverage value.

nicholas-codecov: You can do a deeper dive on the logic here, and maybe explain it to @aj-codecov and myself hah if you do

https://github.com/codecov/codecov-api/blob/main/timeseries/helpers.py#L61-L88

PR #2816, opened 2024-04-29
Fix flag/component select 404 on flags and components tables

spalmurray-codecov:
Bug was caused by trying to nullish coalesce on empty string - which is not nullish.

Closes https://github.com/codecov/feedback/issues/337


ajay-sentry (approved):
lgtm, discussed updating other instance but going to approve to unblock

PR #2820, opened 2024-04-30
fix: Add fallback to owner param if doesn't exist on useTier hook

ajay-sentry:
# Description

fixes: https://github.com/codecov/internal-issues/issues/457

When calling the useTier hook, we're returning an error to the client because no owner is passed in. This could be due to the presence of no owner being in the URL params, which is then passed down to the ReposTable component and then used to call the hook.

One interesting thing I noticed with the useTier hook called in the ReposTable specifically, is this is the only component where we have a fallback for owner set on another hook, useOwner. My thinking here is this could have been a case where we forgot to add it to this one as well, though even going back to old commit history it's hard to see why the fallback was added here specifically. PR reference here: https://github.com/codecov/gazebo/pull/1650

It am not completely certain if this is the fix, it just seems like a logical breadcrumb and fall-back to have added in the first place

[Related logs](https://console.cloud.google.com/logs/query;query=%22required%20type%20'String!'%22%0Alabels.%22k8s-pod%2Fapp%22%3D%22api-graphql%22;pinnedLogId=2024-04-19T15:07:42.753037541Z%2Fg4roixc8pa4ahddn;summaryFields=jsonPayload%252Fusername,labels%252F%2522k8s-pod%252Fpod-template-hash%2522,jsonPayload%252Frequest_body%252Fvariables%252Fowner,jsonPayload%252Frequest_body%252Fquery,jsonPayload%252Frequest_body%252Fvariables%252Fusername:false:32:beginning;lfeCustomFields=labels%252F%2522compute.googleapis.com%252Fresource_name%2522;cursorTimestamp=2024-04-19T15:07:42.753037541Z;duration=P30D?project=genuine-polymer-165712)

# Link to Sample Entry

# Legal Boilerplate


src/shared/ListRepo/ReposTable/ReposTable.tsx
```diff
@@ -105,7 +105,10 @@ const ReposTable = ({
   })
   const isCurrentUserPartOfOrg = ownerData?.isCurrentUserPartOfOrg
 
-  const { data: tierName } = useTier({ provider, owner })
+  const { data: tierName } = useTier({
+    provider,
+    owner: owner || userData?.user?.username,
```
JerrySentry, line 110: the fallback is the signed in user and `owner` is the org being viewed? Shouldn't this always return the org's plan tier and shouldn't ever return the user's plan tier?

ajay-sentry:
Closing PR since this is likely a more systemic issue

JerrySentry (approved):

PR #2826, opened 2024-05-01
feat: Implement RadioTileGroup component

spalmurray-codecov:
# Description

Adds a new RadioTileGroup component to be used in repo and bundle onboarding. Description added preemptively to work with Failed test onboarding designs, which is coming soon.

Uses radixui react radio group and react label as base for accessibility.

[Repo and bundle onboarding designs](https://www.figma.com/file/ICgJ2z5wJSxs0kzrtOp073/GH-1439?type=design&node-id=1-238&mode=design&t=AJwVUnmg0UpBAybr-0)
[Failed test onboarding designs](https://www.figma.com/file/MbfQ7BE0ENmUOPRIe6GMvV/GH-1494?type=design&node-id=1-2&mode=design&t=GQQKAWFEEFJNIV55-0)

# Screenshots

Storybook:

Screenshot 2024-05-01 at 16 23 05

### DIFF
--- a/src/ui/RadioTileGroup/RadioTileGroup.spec.tsx
+++ b/src/ui/RadioTileGroup/RadioTileGroup.spec.tsx
@@ -64,12 +64,12 @@ describe('RadioTileGroup', () => {
       const tile = await screen.findByText('Asdf')
       const tile2 = await screen.findByText('Jkl;')
 
-      user.click(tile)
+      await user.click(tile)
 
       const selected = await screen.findByTestId('radio-button-circle-selected')
       expect(selected).toBeInTheDocument()
 
-      user.click(tile2)
+      await user.click(tile2)
 
       await waitForElementToBeRemoved(selected)
 
--- a/src/ui/RadioTileGroup/RadioTileGroup.spec.tsx
+++ b/src/ui/RadioTileGroup/RadioTileGroup.spec.tsx
@@ -7,7 +7,7 @@ import { userEvent } from '@testing-library/user-event'
 
 import { RadioTileGroup } from './RadioTileGroup'
 
-describe('Card', () => {
+describe('RadioTileGroup', () => {
   function setup() {
     return {
       user: userEvent.setup(),
@@ -17,8 +17,12 @@ describe('Card', () => {
   it('renders', async () => {
     render(
       <RadioTileGroup>
-        <RadioTileGroup.Item value="asdf" label="Asdf" />
-        <RadioTileGroup.Item value="jkl;" label="Jkl;" />
+        <RadioTileGroup.Item value="asdf">
+          <RadioTileGroup.Label>Asdf</RadioTileGroup.Label>
+        </RadioTileGroup.Item>
+        <RadioTileGroup.Item value="jkl;">
+          <RadioTileGroup.Label>Jkl;</RadioTileGroup.Label>
+        </RadioTileGroup.Item>
       </RadioTileGroup>
     )
     const item1 = await screen.findByText('Asdf')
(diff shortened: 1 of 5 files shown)


src/ui/RadioTileGroup/RadioTileGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+  extends React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Item>,
+    VariantProps<typeof item> {
+  label: string
+  description?: string
+}
+
+const Item = React.forwardRef<
+  React.ElementRef<typeof RadioGroupPrimitive.Item>,
+  ItemProps
+>(({ className, label, description, flex, ...props }, ref) => {
```
spalmurray-codecov, line 55: Considered using children as the label. Lmk if you think that's better. Chose not to bc seemed weird to have label as children but not description. Can't really have both with current markup but could probably figure out how to get it to work.

src/ui/RadioTileGroup/RadioTileGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+      </div>
+      <RadioGroupPrimitive.Indicator className="absolute right-0 top-0 flex h-full w-full justify-end">
+        <div className="absolute h-full w-full rounded-md border-2 border-ds-blue-darker" />
+        <div className="h-full w-full rounded-md border border-ds-blue-darker p-4">
+          <div className="flex h-5 w-full items-center justify-end">
+            <RadioButtonCircle selected />
+          </div>
+        </div>
+      </RadioGroupPrimitive.Indicator>
+    </RadioGroupPrimitive.Item>
```
spalmurray-codecov, line 79: radixui shows the `RadioGroupPrimitive.Indicator` when selected. The CSS here overlays the selected status border and circle **on top of** the unselected state. As far as I could tell, there's not a better way to access the current selected state for use with conditional styles, so this seemed like the best option.

src/ui/RadioTileGroup/RadioTileGroup.tsx
nicholas-codecov: I can see a way to get this to work ... however we will have to ensure that when developing people put them in the correct order.

You'd have your label component as:

```tsx
<div className="flex items-center justify-between gap-4">
  <p className="font-medium">{children}</p>
  <RadioButtonCircle />
</div>
```

Description component as:

```tsx
{description ? (
  <p className="text-left text-ds-gray-quinary">{children}</p>
) : null}
```

And than the radio group as:

```tsx
<div className="flex h-full flex-col gap-2 rounded-md border border-ds-gray-quaternary p-4">
  {children}
</div>
```

src/ui/RadioTileGroup/RadioTileGroup.spec.tsx
```diff
@@ -0,0 +1,75 @@
...
+      render(
+        <RadioTileGroup>
+          <RadioTileGroup.Item value="asdf" label="Asdf" />
+          <RadioTileGroup.Item value="jkl;" label="Jkl;" />
+        </RadioTileGroup>
+      )
+      const tile = await screen.findByText('Asdf')
+      const tile2 = await screen.findByText('Jkl;')
+
+      user.click(tile)
```
nicholas-codecov, line 58: User interactions like these return a promise so they should be `await`'ed

src/ui/RadioTileGroup/RadioTileGroup.stories.tsx
nicholas-codecov, line 1: Would you mind adding in a controlled input story, something along the lines of:

```tsx
export const WithControlledInput: Story = {
  args: {
    direction: 'row',
    flex: 1,
  },
  render: (args) => {
    // eslint-disable-next-line react-hooks/rules-of-hooks
    const [value, setValue] = useState<string | undefined>(undefined)

    return (
      <RadioTileGroup
        className="w-full"
        direction={args.direction}
        value={value}
        onValueChange={(value) => {
          setValue(value)
          // controlled input state isn't always required, you can also do things here ... like navigation etc.
        }}
      >
        <RadioTileGroup.Item value="radio" label="Radio" flex={args.flex} />
        <RadioTileGroup.Item value="tile" label="Tile" flex={args.flex} />
        <RadioTileGroup.Item value="group" label="Group" flex={args.flex} />
      </RadioTileGroup>
    )
  },
}
```

src/ui/RadioTileGroup/RadioTileGroup.tsx
nicholas-codecov: This way to we get a bit finer controls over the classnames for the label and description if there's a one off request

nicholas-codecov (commented):
Quick few things to take a peak at

nicholas-codecov (approved):
LGTM

PR #2835, opened 2024-05-02
feat: Implement cn utility to wrap classnames and tailwind-merge

spalmurray-codecov:
Adds `cn` function as a wrapper around both classnames and tailwind-merge. This is to be used in our new reusable components right out of the gate, but a longer term migration plan for old classnames instances may be in order.

### DIFF
--- a/src/shared/utils/cn.spec.ts
+++ b/src/shared/utils/cn.spec.ts
@@ -25,6 +25,13 @@ describe('cn utility', () => {
 
       expect(className).toEqual('text-white text-base bg-ds-pink-tertiary asdf')
     })
+
+    it('should flatten arrays', () => {
+      const arr = ['b', { c: true, d: false }]
+      const className = cn('a', arr)
+
+      expect(className).toEqual('a b c')
+    })
   })
 
   describe('tailwind-merge functionality', () => {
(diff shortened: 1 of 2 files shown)


RulaKhaled (approved):
Nice!

nicholas-codecov (approved):
Make it so!

src/shared/utils/cn.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+        {
+          'bg-ds-primary-base': false,
+          'bg-ds-pink-tertiary': true,
+        },
+        'asdf',
+        { asdf: null },
+        null
+      )
+
+      expect(className).toEqual('text-white text-base bg-ds-pink-tertiary asdf')
```
ajay-sentry, line 26: Interesting about 'asdf' and 'asdf:null' resolving to 'asdf', it looks like we're doing ORs as the default?

spalmurray-codecov: Yea that is interesting. There appears to be [a version of classnames](https://github.com/JedWatson/classnames?tab=readme-ov-file#alternate-dedupe-version) that handles this. What do you think is better?

spalmurray-codecov: Discussed in Slack. Sticking with default.

PR #2837, opened 2024-05-02
feat: Route to plan if user session + to param == plan

ajay-sentry:
# Description

Other half of the pricing page to plan redirect ticket where we're routing to the plan page if the user clicks the "Try for Free" button on the pricing page.

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/e254e109-b8d5-4763-abcb-efe0461711da

# Link to Sample Entry

# Legal Boilerplate


src/App.tsx
```diff
@@ -31,27 +31,31 @@ const SyncProviderPage = lazy(() => import('./pages/SyncProviderPage'))
 
 interface URLParams {
   provider: string
+  to: string
```
spalmurray-codecov, line 34: I think this line is unnecessary

src/App.spec.tsx
```diff
@@ -23,17 +25,6 @@ jest.mock('./pages/TermsOfService', () => () => 'TermsOfService')
 jest.mock('./pages/EnterpriseLandingPage', () => () => 'EnterpriseLandingPage')
 jest.mock('./pages/SyncProviderPage', () => () => 'SyncProviderPage')
 
-jest.mock('./shared/GlobalBanners', () => () => '')
```
ajay-sentry, line 26: none of these mocks seemed to affect the suite

src/App.tsx
ajay-sentry: thanks for the catch! forgot to remove that

src/App.tsx
```diff
@@ -38,20 +38,23 @@ const HomePageRedirect = () => {
...
   const { params } = useLocationParams()
+  // @ts-expect-error useLocationParams needs to be typed
+  const { setup_action: setupAction, to } = params
 
   let redirectURL = '/login'
 
   if (internalUser && internalUser.owners) {
     redirectURL = `/${internalUser.owners[0]?.service}/${internalUser.owners[0]?.username}`
+    if (to === 'plan') {
+      redirectURL = '/plan' + redirectURL
```
RulaKhaled, line 49: so this would be /plan/login? 🤔

src/App.spec.tsx
RulaKhaled: just a heads up there's a failing test

ajay-sentry: @RulaKhaled Haha yeah, called it out on slack that the mocks are being funky if you may have a workaround in your back pocket :O

src/App.tsx
ajay-sentry: on 47 we're updating the redirectURL if we enter the if clause! /login is just the "base" value if we never hit any of the if statements

ajay-sentry: This pattern should allow us to have links like app.codecov.io?to="_____" which can now utilize the user's session and direct them to where we need to go in the app

For now, I think we only have a single case for this, but if we decide to build out this pattern more it might make sense to have a config somewhere mapping the to values to a path

spalmurray-codecov: this is cool

spalmurray-codecov (approved):
lgtm

PR #2850, opened 2024-05-07
fix: Hide sunburst on very large repos

nicholas-codecov:
# Description

This PR contains a temp fix that hides the sunburst chart if there are more than 200,000 files in a given repo, until we have some more time to further investigate and come up with a solution for this.

# Notable Changes

- Grab file count in new `useCoverageTabData` hook
- Hide Sunburst currently if file count is less than or equal to `200_000`
- Update/add in tests

### DIFF
--- a/src/pages/RepoPage/CoverageTab/CoverageTab.jsx
+++ b/src/pages/RepoPage/CoverageTab/CoverageTab.jsx
@@ -49,7 +49,7 @@ function CoverageTab() {
 
   let displaySunburst = false
   const fileCount = data?.branch?.head?.totals?.fileCount
-  if (typeof fileCount === 'number' && fileCount <= 10_000) {
+  if (typeof fileCount === 'number' && fileCount <= 100_000) {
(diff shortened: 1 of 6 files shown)


src/pages/RepoPage/CoverageTab/CoverageTab.jsx
```diff
@@ -25,13 +27,26 @@ const Loader = () => (
...
+  const { data } = useCoverageTabData({
+    provider,
+    owner,
+    repo,
+    branch: branch,
+  })
+
+  let displaySunburst = false
+  const fileCount = data?.branch?.head?.totals?.fileCount
+  if (typeof fileCount === 'number' && fileCount <= 200_000) {
```
ajay-sentry, line 46: should this be 10_000? And maybe we can set to a const to avoid having magic numbers; like MAX_FILES_FOR_SUNBURST

ajay-sentry (approved):
looks good

PR #2857, opened 2024-05-07
ref: Convert useComparisonForCommitAndParent to TS and remove repositoryDeprecated

ajay-sentry:
# Description

Converts useComparisonForCommitAndParent and related files to TS, and removes repositoryDeprecated from query.

# Code Example

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/ab6bce5b-9786-47ae-84c2-71301423ab8e

# Link to Sample Entry

# Legal Boilerplate


src/services/comparison/utils.ts
```diff
@@ -1,13 +1,26 @@
...
+}): string | null {
   if (isNewFile) return 'New'
   if (isRenamedFile) return 'Renamed'
   if (isDeletedFile) return 'Deleted'
   return null
 }
 
-export function transformImpactedFileData(impactedFile) {
+export function transformImpactedFileData(
+  impactedFile: z.infer<typeof ImpactedFileSchema>
```
rohitvinnakota-codecov, line 22: instead of this, you can export the `typeof` in the hook. This still works but that's generally what everyone else has been doing. 

I think we have a `export type ImpactedFileType` already in `useCommit`

rohitvinnakota-codecov (dismissed):

src/services/comparison/utils.ts
ajay-sentry: That's a good call out, swapped in [dc7e63c](https://github.com/codecov/gazebo/pull/2857/commits/dc7e63c895fbc6af732eaea8cffd86868c0696e3)

rohitvinnakota-codecov (approved):

PR #2859, opened 2024-05-08
fix: Move the banner above summary, and fix the use of team hook 

RulaKhaled:
# Description
The purpose of this PR is to left the onboarding banner up the summary. The banner should be visible for the user under the coverage tab 

# Notable Changes

- [x] https://github.com/codecov/codecov-api/pull/551
- Added `isFirstPR` field under repo. We needed a quick way to know whether this repo has an unmerged first PR in Codecov 
- Call repo hook in coverage tab. No need to call team settings in coverage tab not sure why we ever did that 
- Create the onboarding banner under coverage tab 

# Screenshots
Screenshot 2024-05-07 at 1 54 28 PM

closes: https://github.com/codecov/engineering-team/issues/1541

# Legal Boilerplate

### DIFF
--- a/src/pages/RepoPage/CoverageTab/subroute/FileExplorer/shared/RepoContentsResult.tsx
+++ b/src/pages/RepoPage/CoverageTab/subroute/FileExplorer/shared/RepoContentsResult.tsx
@@ -1,4 +1,3 @@
-import A from 'ui/A'
 import Banner from 'ui/Banner'
 import BannerContent from 'ui/Banner/BannerContent'
 
@@ -16,6 +15,15 @@ const RepoContentsResult: React.FC<RepoContentsProps> = ({
   hasComponentsSelected,
 }) => {
   let copy: JSX.Element | string = ''
+
+  if (
+    !isSearching &&
+    !isMissingHeadReport &&
+    !hasFlagsSelected &&
+    !hasComponentsSelected
+  ) {
+    return null
+  }
 
   if (isSearching) {
     copy = 'No results found'
@@ -36,22 +44,6 @@ const RepoContentsResult: React.FC<RepoContentsProps> = ({
         No coverage report uploaded for the selected flags in this branch's
         head commit
       `
-  } else {
-    copy = (
-      <p>
-        Once merged to your default branch, Codecov will show your report
-        results on this dashboard.{' '}
-        <A
-          to={{ pageName: 'settings' }}
-          hook={'repo-to-edit-branch'}
-          variant="semibold"
-          isExternal={false}
-          data-testid="settings-page"
-        >
-          edit default branch
-        </A>
-      </p>
-    )
   }
 
   return (
--- a/src/pages/RepoPage/CoverageTab/subroute/FileExplorer/shared/RepoContentsResult.tsx
+++ b/src/pages/RepoPage/CoverageTab/subroute/FileExplorer/shared/RepoContentsResult.tsx
@@ -1,3 +1,4 @@
+import A from 'ui/A'
 import Banner from 'ui/Banner'
 import BannerContent from 'ui/Banner/BannerContent'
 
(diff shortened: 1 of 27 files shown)


src/pages/RepoPage/CoverageTab/subroute/FileExplorer/shared/RepoContentsResult.tsx
nicholas-codecov, line 1: `m`: I know you're moving the banner to the top, however I think it's still worth showing some form of an error message down here, instead of just an empty table, users "shouldn't" miss the banner at the top, but they might, so I think we should still leave some error message down in the table so they're at least aware of some issue occurring.

nicholas-codecov (commented):
Small request, that I think should help the users out

src/pages/RepoPage/CoverageTab/subroute/FileExplorer/shared/RepoContentsResult.tsx
RulaKhaled: Do you suggest we leave the banner there as well, or loop in Kyle for a new copy?

nicholas-codecov: Uhh i think new copy would probably be best imo, i would show something similar to how we currently display the other error messages ... the banner always felt out of place

nicholas-codecov: Or if you feel comfortable coming up with something, i think that'll be alright as well, it could be the same copy that's in the banner ... just not in the banner

RulaKhaled: went with copy with no banner Screenshot 2024-05-21 at 12 36 13 PM

nicholas-codecov: Hrrm, is there anyway to include this like the other errors where the table header is still present? just to keep things consistent across gazebo

nicholas-codecov: Hrrm, what happened then? it was there previously with the other errors, it also seems like all the other errors are wrapped in a banner? that makes this one a pretty inconsistent with the other error messages that could appear here

RulaKhaled: visibility issue, feel free to read https://github.com/codecov/engineering-team/issues/1541 OR reach out to @katia-sentry and @codecovdesign for more context.

RulaKhaled: Fwiw this banner is different from other errors we display, as it's related to user onboarding (it's not even an error), i suppose that's why @codecovdesign created the issue

codecovdesign: >  different from other errors we display, as it's related to user onboarding (it's not even an error)

Yes, that is correct, it's not an error rather it's a moment in onboarding where the data isn't showing since it's not merged to default branch. Sounds like there are no blockers here – If there is a follow up for inconsistency, could we create an issue with the details to capture and resolve separately?

RulaKhaled: Thank you @codecovdesign 

@nicholas-codecov can you create an issue for the inconsistency please? Also as it is not a blocker for this PR can we move forward?

nicholas-codecov: Since we're introducing the inconsistency in this PR in the first place, and it's only two lines. Can we revert those two lines and bring back the assignment to the `copy` var? We can then go through and update all of them together so we get to tackle the issue without introducing any inconsistency.

RulaKhaled: @nicholas-codecov so delete the copy that you asked me to add?

nicholas-codecov (approved):

PR #2862, opened 2024-05-08
ref: Move useReposBackfilled off Repo Deprecated

ajay-sentry:
# Description

Similar to other conversions in https://github.com/codecov/engineering-team/issues/355, converts useReposBackfilled off RepoDeprecated, with some other small TS conversions and test fixes

# Screenshots

Screenshot 2024-05-08 at 11 44 25 AM

# Link to Sample Entry

# Legal Boilerplate


src/services/repo/hooks.spec.tsx
```diff
@@ -31,15 +32,20 @@ const wrapper =
...
 
 beforeAll(() => server.listen())
-afterEach(() => server.resetHandlers())
+afterEach(() => {
+  queryClient.clear()
+  server.resetHandlers()
+})
 afterAll(() => server.close())
 
+console.error = () => {}
```
nicholas-codecov, line 41: Can you mock this console in a setup, but also reset it after the test is finished, just to ensure that the console is back and running normally.

src/services/repo/useRepoBackfilled.tsx
```diff
@@ -0,0 +1,127 @@
...
+})
+
+interface URLParams {
+  provider: string
+  owner: string
+  repo: string
+}
+
+export function useRepoBackfilled() {
+  const { provider, owner, repo } = useParams<URLParams>()
```
nicholas-codecov, line 71: `m`: We've shifted away from grabbing these in the service hooks directly and instead have moved to favouring them being passed in as args, do you mind making that quick update here?

nicholas-codecov (commented):
Couple of things to peak at

src/services/repo/hooks.spec.tsx
ajay-sentry: Yep yep! Fixed in [5ef5686](https://github.com/codecov/gazebo/pull/2862/commits/5ef5686ed967ef9475f3562b92ddb03847bef5e3)

src/services/repo/useRepoBackfilled.tsx
ajay-sentry: Definitely agree with you here, looking at this reference though: https://github.com/codecov/gazebo/blob/9f3692ff90a2a74d2be697f178c724d7b7cdf9d8/src/pages/RepoPage/FlagsTab/BackfillBanners/hooks.ts#L3-L4, we end up going down a rabbit hole increasing the scope of this PR to a degree I'm not super comfortable with. Would prefer to leave as is and tackle in a separate PR

src/services/repo/hooks.spec.tsx
```diff
@@ -212,6 +221,7 @@ describe('useUpdateRepo', () => {
           wrapper: wrapper(),
         })
 
+        // @ts-expect-error
```
nicholas-codecov, line 224: Interested as to what's going on here? is it just that the current hook that returns the mutation hasn't been uploaded to TS yet?

src/services/repo/hooks.spec.tsx
```diff
@@ -266,5 +305,60 @@ describe('useRepoBackfilled', () => {
...
+        await waitFor(() => expect(result.current.isError).toBeTruthy())
+        await waitFor(() =>
+          expect(result.current.error).toEqual(
+            expect.objectContaining({
+              status: 404,
+            })
+          )
+        )
+      })
+      it('can return owner not activated error', async () => {
```
nicholas-codecov, line 347: Just me being a little picky, moving forward if you're working on tests do you mind adding in an empty line in between these test blocks, I feel like it makes things a lot cleaning when you're working around the tests and "mini-mizing" them in your editor.

nicholas-codecov (approved):
Approved, with a question, and a comment

src/services/repo/hooks.spec.tsx
ajay-sentry: For sure, I think stylistic things like this could be brought up in our weekly meeting as well if we want to make team norms around them!

src/services/repo/useRepoBackfilled.tsx
ajay-sentry: Screenshot 2024-05-14 at 9 23 47 AM
Screenshot 2024-05-14 at 9 24 27 AM

Doing a quick search in the codebase, we have 54 hooks to update. I can go ahead and create an epic for that; though I don't really see this getting prioritized

PR #2863, opened 2024-05-08
feat: Add metrics to A and Button on click

ajay-sentry:
# Description

This PR aims to give us insight into click events occurring in the DOM for both the A and Button elements. We're using the existing Sentry metrics configuration to increment a metric on each click by hydrating the onClick event with this metric increment call.

Instead of doing each individually, this will give us a framework to get it for all A's and Buttons for free, and when our new event system is built out we can just swap out these for those instead.

Closes https://github.com/codecov/engineering-team/issues/1552

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/e5fef0a3-d316-4ebf-8a32-61f8f6337b7d

- In the above video we can see the event metric names being pushed to the console

# Link to Sample Entry

# Legal Boilerplate


nicholas-codecov (changes requested):
Some concerns here that these will break typesafety if converted to TS

src/ui/A/A.jsx
```diff
@@ -58,13 +59,24 @@ function A({ to, hook, variant = 'default', children, isExternal, ...props }) {
...
 
+  const handleClick = () => {
+    if (props.onClick) {
+      props.onClick()
+    }
+    if (to) {
+      metrics.increment(`link.click.${to.pageName}`)
+    } else {
+      metrics.increment(`link.click.${hook}`)
+    }
```
nicholas-codecov, line 70: Thoughts on refining this check to ensure that `pageName` exists or adding in a `.?`, as this isn't TypeSafe atm, just so we don't crash the App trying to collect metrics?

src/ui/A/A.jsx
```diff
@@ -58,13 +59,24 @@ function A({ to, hook, variant = 'default', children, isExternal, ...props }) {
     'inline-flex items-center gap-1'
   )
 
+  const handleClick = () => {
```
nicholas-codecov, line 62: Should we grab the event here and pass it along to the `onClick` call handler in case we're using it anywhere?

src/ui/Button/Button.jsx
```diff
@@ -114,6 +115,17 @@ function Button({
...
+  const handleClick = () => {
+    if (props.onClick) {
+      props.onClick()
+    }
+    if (to) {
+      metrics.increment(`button.click.${to.pageName}`)
+    } else {
+      metrics.increment(`button.click.${hook}`)
+    }
+  }
```
nicholas-codecov, line 127: Same stuff as above here 👀

nicholas-codecov (commented):
Some quick comments

src/ui/A/A.jsx
ajay-sentry: I think the ? would be unnecessary here since we're already guarding on "to" being present on line 66, nor would it crash the app.

Here's an example:
Screenshot 2024-06-13 at 10 40 03 AM

nicholas-codecov: ah yea thought the type error would be one level down

ajay-sentry:
Closing out, will revisit when Codecov specific eventing work is complete

PR #2872, opened 2024-05-10
ref: Remove repositoryDeprecated from api tests

nicholas-codecov:
# Description

This PR updates the `api` util file to TS, as well as removing the `repositoryDeprecated` field in the `api` tests.

# Notable Changes

- Update `api` to TS
- Update tests and remove `repositoryDeprecated`


src/shared/api/api.ts
```diff
@@ -47,30 +55,46 @@ function _fetch({
   })
 }
 
-function prefillMethod(method) {
-  return (config) =>
+function prefillMethod(method: string) {
+  return (config: any) =>
```
spalmurray-codecov, line 59: There's no type we can use here?

nicholas-codecov: Config isn't typed yet ... and until we can ... I'd rather this just be any for the time being ... even tho it hurts every bone in my body.

spalmurray-codecov: Ok as long as you feel bad about it 😆

nicholas-codecov: Feeling absolutely sick to my stomach about it

PR #2873, opened 2024-05-13
 chore: Remove/replace useCoverageWithFilters

nicholas-codecov:
# Description

This PR actually removes the `useCoverageWithFilters` hook instead replacing it's use with the original hook and extending it to support the features that were specific to `useCoverageWithFilters` reducing the chances of actually creating a bug and difference between the two (we weren't fetching the same info for the same use 😬)

GH codecov/engineering-team#355

# Notable Changes

- Remove `useCoverageWithFilters`
- Remove associated utils
- Update `useFileWithMainCoverage` to support new cases
- Update `useCommitBasedCoverageForFileViewer` to use `useFileWithMainCoverage` ... twice
- Update imports
- Update tests


src/services/pathContents/utils.ts
```diff
@@ -14,7 +14,7 @@ export function extractCoverageFromResponse(
   if (!coverageFile) return null
   const lineWithCoverage = keyBy(coverageFile?.coverage, 'line')
   const fileCoverage = mapValues(lineWithCoverage, 'coverage')
-  const coverageTotal = coverageFile?.totals?.coverage
```
ajay-sentry, line 17: was this a bug fix too?

ajay-sentry (approved):
seems legit, send it

src/services/pathContents/utils.ts
nicholas-codecov: Nope, i had originally aliased the field in GQL, this is me removing it now and updating it, with TS makes it easier to find the errors

ajay-sentry: nice! love it

PR #2882, opened 2024-05-16
ref: Refresh CopyClipboard component

spalmurray-codecov:
# Description

Updates the CopyClipboard component to bring it in line with the new system of components using shadcn and cva.

Closes https://github.com/codecov/internal-issues/issues/480


rohitvinnakota-codecov (approved):

src/ui/CopyClipboard/CopyClipboard.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const CopyClipboard = React.forwardRef<HTMLButtonElement, CopyClipboardProps>(
+  ({ value, onClick = () => {}, variant, className, ...props }, ref) => {
+    const [showSuccess, setShowSuccess] = useState(false)
+
+    const handleCopy = () => {
+      copy(value)
+      setShowSuccess(true)
+      setTimeout(() => setShowSuccess(false), 1500)
+      onClick()
```
nicholas-codecov, line 34: What're your thoughts on adding the `token` here in the `onClick` callback?

src/ui/CodeRenderer/FileHeader/FileHeader.jsx
```diff
@@ -26,9 +26,7 @@ function CodeRendererCoverageHeader({ header, headName, coverage, fileLabel }) {
       <div className="flex items-center gap-1 text-ds-gray-quinary">
         <span>{header}</span>
         <span className="font-semibold">{headName}</span>
-        {headName && (
-          <CopyClipboard string={headName} showLabel={false} variant="muted" />
-        )}
+        {headName ? <CopyClipboard value={headName} /> : null}
```
nicholas-codecov, line 29: 🙌

src/ui/CopyClipboard/CopyClipboard.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    return (
+      <button onClick={handleCopy} ref={ref} {...props}>
```
nicholas-codecov, line 38: Can we add an `aria-label` here ... we may need another prop for this so people can define what the button does ... something like `ariaLabel=`Copy ${label}`, that way for like the upload token we could have `Copy upload token`

nicholas-codecov (commented):
Couple things to peak at

src/ui/CopyClipboard/CopyClipboard.tsx
spalmurray-codecov: Because onClick is a native attribute of the button - it won't let me change the type to be `(value: string) => void`

So could either change the name or just not bother.

An argument in favour of leaving it as is: if the caller wants to use the value in the callback they can still do that because they have the value up where they call `CopyClipboard`.

LMK what you think

nicholas-codecov: Hrrm i think this is identifying a code smell here, since your `onClick` is typed as a button `onClick` you shouldn't be calling it in your `handleCopy` function because your not passing it the `event` that is present on an `onClick`, I wonder if we should switch this from `onClick` to `onCopy` that's a bit more intuitive as to the action being made OR override the `onClick` type and remove it from being passed to the button JSX element and calling it directly in the `handleCopy` function

PR #2883, opened 2024-05-16
ref: Refactor useSingularImpactedFileComparison

RulaKhaled:
# Description
Refactor hook and get rid of repo deprecated 

# Legal Boilerplate


src/services/pull/useSingularImpactedFileComparison.tsx
```diff
@@ -0,0 +1,225 @@
...
+          ... on OwnerNotActivatedError {
+            message
+          }
+        }
+      }
+    }
+    ${FileComparisonWithBase}
+`
+
+const ImpactedFileSchema = z.object({
```
rohitvinnakota-codecov, line 45: Can we import this from `useComparisonForCommitAndParent` just so we have consistent definitions and clean up the file a bit?

RulaKhaled: I'd rather leave it as it is since they're not using any shared query/fragment. This would cause errors if we ever changed one of these two hooks' schemas. It's a bad idea for different hooks' schemas to depend on each other.

rohitvinnakota-codecov (approved):

PR #2889, opened 2024-05-20
fix: Session Expiration redirects for teams and enterprise

ajay-sentry:
# Description

A couple bug fixes in this PR, all centered around getting the session redirect to work properly for enterprise customers; and consequently non-enterprise customer regression fixes.

Fixes:
- Incorrect redirect path set for enterprise causing infinite render 
- Removing unnecessary logic for rendering/mounting the SessionExpiryTracker component
- Adding a cookie removal for the session_expiry amount if the user forces a sign out
- Adding the expiry banner to the enterprise login page if the user has an expired session
- Removing the sessionid cookie and redirect user after the session expiry value is elapsed

In the below videos we can see in the console the polling mechanism at play (I had to bump down the session expiry to a minute and the polling interval to 6 seconds to easily reproduce) for both enterprise and non-enterprise customers

Polling interval stuff in API is here: https://github.com/codecov/codecov-api/blob/c3f5e7f903c63f8a5eadc5a69a2846df57e65a0a/codecov_auth/views/github.py#L168

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/b584515b-77f6-4de8-b5f1-551c71c4bf76

https://github.com/codecov/gazebo/assets/159853603/53cf611c-ed20-4b21-9cc6-c5a181ae390c

# Link to Sample Entry

# Legal Boilerplate


src/config.js
```diff
@@ -10,6 +10,12 @@ const defaultConfig = {
   GH_APP: 'codecov',
 }
 
+export const LOCAL_STORAGE_SESSION_EXPIRED_KEY = 'expired-session'
```
ajay-sentry, line 13: moved these to shared consts for clarity and to remove all the redeclarations

PR #2890, opened 2024-05-22
fix: Directly render bundle onboarding

nicholas-codecov:
# Description

This PR shuffles around how we render bundle onboarding tackling an infinite redirect issue we're currently seeing when a repo is not active ... but has bundle analysis enabled.

# Notable Changes

- Render `BundleOnboarding` directly in `RepoPage`
- Remove `BundleOnboarding` from `BundlesTab`
- Update tests

### DIFF
--- a/src/pages/RepoPage/RepoPage.tsx
+++ b/src/pages/RepoPage/RepoPage.tsx
@@ -15,6 +15,7 @@ import RepoBreadcrumb from './RepoBreadcrumb'
 import RepoPageTabs from './RepoPageTabs'
 
 const BundlesTab = lazy(() => import('./BundlesTab'))
+const BundleOnboarding = lazy(() => import('./BundlesTab/BundleOnboarding'))
 const CommitsTab = lazy(() => import('./CommitsTab'))
 const CoverageTab = lazy(() => import('./CoverageTab'))
 const NewRepoTab = lazy(() => import('./CoverageOnboarding'))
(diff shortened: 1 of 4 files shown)


src/pages/RepoPage/RepoPage.tsx
```diff
@@ -123,7 +124,7 @@ function Routes({
             ]}
             exact
           >
-            <BundlesTab />
+            <BundleOnboarding />
```
RulaKhaled, line 127: Can you check if `jsOrTsPresent`? for some reason it causes infinite loop if it's ever hard coded in url.
maybe in https://github.com/codecov/gazebo/blob/66f3413100e3905d8f85c207a7901ad8d19f6933/src/pages/RepoPage/BundlesTab/BundleOnboarding/BundleOnboarding.tsx#L126

nicholas-codecov: What?

nicholas-codecov: What? I'm confused as to what you're talking about, this hard redirect will redirect you to `/:provider` and the code you've highlighted is protected by the check you're requesting.

RulaKhaled: it's obviously not part of the change which is why i didn't see it :)

RulaKhaled (approved):

PR #2891, opened 2024-05-23
Update self hosted settings requests to use GQL 

rohitvinnakota-codecov:
Porting from old GQL REST endpoint to the new settings resolver and mutation for self hosted setups. 

This PR closes https://github.com/codecov/engineering-team/issues/1473.

Related API PR: 
https://github.com/codecov/codecov-api/pull/559

https://github.com/codecov/gazebo/assets/148245014/e603bbf3-c3c8-4adb-8f17-c35281c72389

# Legal Boilerplate


src/services/account/useUpdateSelfHostedSettings.tsx
```diff
@@ -0,0 +1,74 @@
...
+            disappearAfter: 10000,
+          })
+        }
+      }
+    },
+    onError: () => {
+      addToast({
+        type: 'error',
+        text: <UpdateSelfHostedSettingsMessage />,
+        disappearAfter: 10000,
```
ajay-sentry, line 66: nit: can make this a const in lieu of a magic number

src/pages/AdminSettings/AdminMembers/MemberList/MemberTable.tsx
```diff
@@ -143,7 +143,7 @@ function MemberTable() {
   }, [fetchNextPage, inView, hasNextPage])
 
   const tableContent = useMemo(() => {
-    if (!data?.length || !seatData) {
+    if (!data?.length || !seatData || seatData.__typename !== 'Repository') {
```
ajay-sentry, line 146: This check should already be caught in the hook right? Or is this to prevent the "error" page from showing up

rohitvinnakota-codecov: This is just to prevent the type error with table content data. The network boundary should catch the other types.

ajay-sentry (approved):

PR #2896, opened 2024-05-23
ref: Convert onboarding code snippets to new component

spalmurray-codecov:
Refactors code snippet instances to use the new `CodeSnippet` component. Some tests had to be changed as a result of test id of the copyclipboard no longer being directly exposed - which is a drawback of the less than fully composable implementation we went with.

Closes https://github.com/codecov/internal-issues/issues/482


ajay-sentry (approved):
looks good! Tested on Preview deploy and looked to copy links correctly across the board

PR #2898, opened 2024-05-24
feat: Remove pointer events from CodeRenderer when scrolling

nicholas-codecov:
# Description

This PR adds in some new animation frame util helpers that allow us to easily disable pointer events when scrolling when viewing a single file (since there's a lot of pointer events). This is one small step into increasing performance when viewing files.

GH codecov/engineering-team#1088

# Notable Changes

- Add in animation util helpers
- Disable pointers when scrolling in code renderer
- Update a couple tests


JonasBa (approved):
Changes look good, but when talking about performance, it's good to add a before/after perf comparison (either through profile or screencap showing FPS). I would encourage you to add a comment explaining why pointer events can harm performance. Comments like that serve as an explanation for future folks contributing to the codebase, but they can also serve as a small learning for folks who may previously not have been familiar with this behavior

src/ui/CodeRenderer/CodeRenderer.spec.tsx
```diff
@@ -0,0 +1,171 @@
...
+      expect(lines[0]).toBeInTheDocument()
+
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+
+      // eslint-disable-next-line testing-library/no-container
+      container.remove()
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+
+      await waitFor(() => expect(cancelAnimationFrameSpy).toHaveBeenCalled())
```
spalmurray-codecov, line 168: I do not understand what is going on in this test specifically - maybe we can go over this PR in our 1-1 on Tuesday?

src/ui/FileList/FileList.stories.tsx
```diff
@@ -264,16 +264,15 @@ function Example() {
                           ? Math.floor(Math.random() * 100)
                           : null
                       }
-                      headNumber={i + 1}
-                      baseNumber={i}
+                      headNumber={`${i + 1}`}
+                      baseNumber={`${i}`}
```
spalmurray-codecov, line 268: why do this?

spalmurray-codecov (approved):
Code looks good, but I don't understand all the code - hoping you can explain the details to me next week 😁

src/ui/CodeRenderer/CodeRenderer.spec.tsx
nicholas-codecov: Yes of course!

src/ui/FileList/FileList.stories.tsx
nicholas-codecov: `baseNumber` requires a string variable ...

spalmurray-codecov: ahhh ok and we weren't giving it one before got it

src/shared/utils/animationFrameUtils.ts
```diff
@@ -41,6 +41,7 @@ export function requestAnimationTimeout(
   return frame
 }
 
+// small wrapper function around cancelAnimationFrame to use with requestAnimationTimeout
```
JonasBa, line 44: If the function body is small and not complicated, then I think it doesnt warrant an explanation as it will typically be of low value.

src/ui/CodeRenderer/CodeRenderer.tsx
```diff
@@ -38,10 +38,20 @@ function CodeRenderer({
 
   useLayoutEffect(() => {
     const onScroll = (_event: Event) => {
+      /**
```
JonasBa, line 41: Nice comment 🙏🏼

src/ui/CodeRenderer/CodeRenderer.tsx
```diff
@@ -38,10 +38,20 @@ function CodeRenderer({
...
+       * the line number indicators) during the repaint. By disabling these
+       * events, we can reduce the amount of work the browser needs to do
+       * during the repaint, as it does not need to check these events.
+       */
       if (!pointerEventsRaf.current && tableRef.current) {
         tableRef.current.style.pointerEvents = 'none'
       }
 
+      // We then re-enable pointer events after the scroll event has finished so
+      // the user can continue interacting with the line number indicators.
```
JonasBa, line 54: Nit, but you used /* type comment block above

PR #2904, opened 2024-05-28
fix: Table and FileList Style imports

nicholas-codecov:
# Description

This is a small change as to how we handle `Table` and `FileList` CSS imports. Removing them throughout the app and just importing them both once in the root `App` component. We currently seem to be overlapping the styles occasionally and this could be causing some unforeseen issues. As well we're duplicating the amount CSS we're pulling in because every component that pulls one of these CSS files is generating a new CSS output during the bundle process, which this will reduce.

This change will also make sure we don't have to worry about importing these styles anymore and/or accidentally forgetting them.


src/ui/Table/Table.mdx
```diff
@@ -15,7 +15,7 @@ Table UI is a styles only "component" that provides the design system aspect of
 Nest your table with a `div`, this lets the table start scrolling rather then breaking page layouts when a screen becomes too small for the content and there has been no custom responsiveness added.
 
 ```js
```
ajay-sentry, line 17: We can probably remove this empty js block :)

ajay-sentry (approved):
Discussed on slack with Nick, since these CSS files contain explicit class variables, we have high confidence this won't lead to unintended styling regression by doing moving the imports to the highest level of the app. 

If we also get bundle size improvements, even better.

PR #2908, opened 2024-05-29
chore: Remove Google tag manager

spalmurray-codecov:
Removes Google tag manager from the app. Previously we used this to load Segment, but are moving away from segment and can remove GTM entirely. We have all but removed Segment already at this point and don't actually use it, but we have been continuing to load segment in the app, which lead to the linked vulnerability.

This change has been tested locally, on staging, and on prod through GTM previews for any adverse side effects.

Closes https://github.com/codecov/internal-issues/issues/389


src/services/tracking/useTracking.js
```diff
@@ -3,16 +3,11 @@ import * as Sentry from '@sentry/react'
 import { useUser } from 'services/user'
 
 import { useTrackFeatureFlags } from './featureFlags'
-import { gtmUser, setDataLayer } from './gtm'
 import { firePendo, pendoDefaultUser, useUpdatePendoWithOwner } from './pendo'
 import { getUserData } from './utils'
 
+//a
```
ajay-sentry, line 9: nit: unused

spalmurray-codecov: 💀

ajay-sentry (approved):

PR #2909, opened 2024-05-30
feat: Use new POST endpoint for logout

spalmurray-codecov:
# Description

Uses the new POST endpoint for Gazebo's sign out button. Currently logout by both POST and GET are supported on the API side, so we will change this and monitor for a day or so before completing the process and removing the GET endpoint API side.

Closes https://github.com/codecov/internal-issues/issues/487

# Demo

Recording 2024-05-31 at 14 37 35


ajay-sentry (approved):
tested on the preview deploy and seemed to work properly, confirmed the post request was being hit and routed to login page after

PR #2912, opened 2024-05-30
Correctly display cff's when many flags per upload

adrian-codecov:
# Description
We have custom logic on how we display regular and carried forward uploads in our commit page. The previous implementation omitted certain sessions due to the implementation of splice skipping indexes - more here https://stackoverflow.com/questions/9882284/looping-through-array-and-removing-items-without-breaking-for-loop. This change is to fix the implementation there.

Ticket: https://github.com/codecov/feedback/issues/352

# Notable Changes
- Adjusted the deleteDuplicate fn and added test for it

# Legal Boilerplate


ajay-sentry (approved):
lgtm!

PR #2916, opened 2024-06-04
fix: Small fix to remove jumping tab text

nicholas-codecov:
# Description

Just a small fix to always apply a border that's transparent when not selected and set the colour when it is, which removes the jumping text when selected. Also a few tweaks to the TS so that we don't need to cast types.

# Notable Changes

- Use `border-transparent` as default style so that we always account for the border height
- Small TS tweaks
- Use the new built-in `clipboard` functionality from the `CodeSnippet` component (thanks @spalmurray-codecov)

# Screenshots

Before: 

https://github.com/codecov/gazebo/assets/105234307/1e9f4087-6326-4e99-8f59-333d75fcbbe9

After:

https://github.com/codecov/gazebo/assets/105234307/7f8090d1-5ece-4810-87ad-98ce3e0a2c69

### DIFF
--- a/src/pages/RepoPage/FailedTestsTab/GitHubActions/FrameworkTabs/FrameworkTabs.tsx
+++ b/src/pages/RepoPage/FailedTestsTab/GitHubActions/FrameworkTabs/FrameworkTabs.tsx
@@ -2,7 +2,6 @@ import { useState } from 'react'
 
 import { cn } from 'shared/utils/cn'
 import { CodeSnippet } from 'ui/CodeSnippet'
-import { CopyClipboard } from 'ui/CopyClipboard'
 
 const Frameworks = {
   PYTEST: 'Pytest',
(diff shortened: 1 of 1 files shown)


src/pages/RepoPage/FailedTestsTab/GitHubActions/FrameworkTabs/FrameworkTabs.tsx
```diff
@@ -27,27 +27,22 @@ export function FrameworkTabs() {
...
-                selectedFramework === f && 'border-b-2 border-ds-gray-octonary'
-              )}
-              onClick={() => setSelectedFramework(f as FrameworkType)}
-            >
-              {f}
-            </button>
-          ))}
-        </div>
+      <div className="flex gap-1">
+        {Object.values(Frameworks).map((framework) => (
```
nicholas-codecov, line 31: Using `Object.values` gives us a union type of the keys being used in the `FrameworkCopy` const, so that we don't need to cast the type when we set the state.

src/pages/RepoPage/FailedTestsTab/GitHubActions/FrameworkTabs/FrameworkTabs.tsx
```diff
@@ -27,27 +27,22 @@ export function FrameworkTabs() {
...
+          <button
+            key={framework}
+            className={cn(
+              'px-4 py-2 border-b-2 border-transparent',
+              selectedFramework === framework && 'border-ds-gray-octonary'
+            )}
```
nicholas-codecov, line 37: The the old `cs` function this would require us to check both states or else `border-transparent` would always be applied, with the newer `cn` function it will automatically remove the `border-transparent` class as the `border-ds-gray-octonary` comes later so it has precedent over `border-transparent`

RulaKhaled (approved):
thanks!

PR #2928, opened 2024-06-06
ref: Convert useInvoice hook to TS and GQL

ajay-sentry:
# Description

converts the useInvoice hook to TS and GQL, and updates related tests

Confirmed invoice page still loads properly

# Screenshots

Screenshot 2024-06-06 at 9 16 11 AM

# Link to Sample Entry

# Legal Boilerplate


src/services/account/useInvoice.ts
```diff
@@ -0,0 +1,99 @@
...
+
+import Api from 'shared/api'
+import { NetworkErrorObject } from 'shared/api/helpers'
+
+import { InvoiceSchema } from './useInvoices'
+
+const query = `
+query Invoice($owner: String!, $id: String!) {
+  owner(username: $owner) {
+    invoice(invoice_id: $id) {
```
rohitvinnakota-codecov, line 12: Can we have a follow up to covert this to camel case? It's in line with other inputs we have for GQL. eg: `isPublic`

rohitvinnakota-codecov (approved):

src/services/account/useInvoice.ts
ajay-sentry: Updated! Confirmed when running locally it still works

PR #2930, opened 2024-06-06
feat/ref: convert Credit Card Form to TS + add on change behavior

ajay-sentry:
# Description

Converts the CreditCardForm to TS, and adds some extra functionality to show the error message details to the end user when we have an on change event that generates an error.

Previously we were just highlighting the card box to be red, but now we can show the user their error messages in real time :) 

Closes https://github.com/codecov/engineering-team/issues/1867

# Screenshots

https://github.com/codecov/gazebo/assets/159853603/1fe74d3e-754a-4ca1-bab8-705ac1a088eb

# Link to Sample Entry

# Legal Boilerplate


src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/PaymentCard/CreditCardForm.tsx
```diff
@@ -17,21 +25,21 @@ function CreditCardForm({ closeForm, provider, owner }) {
     owner,
   })
 
-  function submit(e) {
+  function submit(e: any) {
```
calvin-codecov, line 28: kind of trivial but we could do `e: React.FormEvent`

src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/PaymentCard/CreditCardForm.tsx
```diff
@@ -17,21 +25,21 @@ function CreditCardForm({ closeForm, provider, owner }) {
...
     e.preventDefault()
-    updateCard(elements.getElement(CardElement), {
+    updateCard(elements!.getElement(CardElement)!, {
```
calvin-codecov, line 30: just curious, how do we know `elements` is defined?

ajay-sentry: Mmmm I made the assumption that because we wait for the card element to load prior to entering information and this element object is handled after the user enters information and then submits that we shouldn't need to worry about it being null

Related docs https://docs.stripe.com/stripe-js/react#useelements-hook

src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/PaymentCard/CreditCardForm.tsx
calvin-codecov: We could possibly file this as a followup ticket to get this feature change out and then also do an investigation to change any other forms not using react-hook-form along with this one at the same time.

nicholas-codecov: Quickly skimming through the codebase, this looks to be the only form that currently does not use `react-hook-form` to manage the state for the various elements of the form.

ajay-sentry: Great! Created https://github.com/codecov/engineering-team/issues/1884 to track that work

src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/PaymentCard/CreditCardForm.tsx
nicholas-codecov: Where do we do this waiting for the card element to load, is it a suspense boundary in the parent component? Just wondering if we should disable form submission until everything is loaded in, as they show in the docs:

```js
    if (!stripe || !elements) {
      // Stripe.js hasn't yet loaded.
      // Make sure to disable form submission until Stripe.js has loaded.
      return;
    }
```

ajay-sentry: I thought about it more from a practical perspective; a user in a vast majority of cases is not going to immediately hit "submit" right after opening the form.

A user will enter their information into the component, which would need to be loaded first, before they would submit. I get why stripe suggests using the guard, but I don't feel like we need it.

If we feel strongly about adding it though I don't mind making the update

nicholas-codecov: Feels like a case where it's better to have it and not need it, than not have it and need it. It doesn't look like too much of a change tho?

ajay-sentry: Practically I don't think we need it, and if we don't have it and _need it_ the element on the form won't load in the first place.

ajay-sentry: Anyway, not a big deal for me to add since we feel strongly about it. Updated in [b95a93f](https://github.com/codecov/gazebo/pull/2930/commits/b95a93f8335f601431b4aef0cf03a9248aa2089b)

nicholas-codecov (approved):

PR #2931, opened 2024-06-07
feat: Add in span for branch contents request

nicholas-codecov:
# Description

Small change adding in a custom span for tracing the `BranchContents` query so we can better investigate as to what is causing slow loading times for this request.


spalmurray-codecov (approved):
this is cool

JerrySentry (approved):
lgtm

PR #2934, opened 2024-06-10
ref: Only display bundle and patch data on PullsTable

nicholas-codecov:
# Description

This PR cleans up the current pulls table removing the project coverage, coverage from base columns, and team versions of these tables. This PR also updates the patch and bundle columns to render their status: `COMPLETED`, `ERROR`, or `PENDING` to better inform the user.

codecov/engineering-team#1497

# Notable Changes

- Remove team version of pulls table and usePulls
- Bump react intersection package
- Remove project coverage, coverage from base columns
- Remove Team versions

# Screenshots

Before:

Screenshot 2024-06-10 at 07 43 51

After:

Screenshot 2024-06-10 at 07 38 23


src/pages/RepoPage/PullsTab/PullsTable/PullsTable.spec.tsx
```diff
@@ -280,28 +261,12 @@ describe('PullsTable', () => {
...
-      expect(coverage).toBeInTheDocument()
-    })
-
     describe('bundle analysis is enabled', () => {
       it('renders bundle analysis column', async () => {
         const { queryClient } = setup({ bundleAnalysisEnabled: true })
         render(<PullsTable />, { wrapper: wrapper(queryClient) })
 
-        const bundleAnalysis = await screen.findByText('Upload: ✅')
+        const bundleAnalysis = await screen.findByText('Upload: ⏳')
```
ajay-sentry, line 269: is the default this hourglass icon now?

src/pages/RepoPage/PullsTab/PullsTable/createPullsTableData.tsx
```diff
@@ -1,34 +1,67 @@
...
     return []
   }
+
   return pulls.filter(Boolean).map((pull: Pull) => {
-    let patch, change
-    if (pull?.compareWithBase?.__typename === 'Comparison') {
-      patch = pull?.compareWithBase?.patchTotals?.percentCovered ?? 0
-      change = pull?.compareWithBase?.changeCoverage ?? 0
+    let patch = <p>-</p>
+    if (pull?.head?.coverageStatus === 'ERROR') {
```
ajay-sentry, line 27: There an enum for these? 'ERROR', 'PENDING' and 'COMPLETED'

nicholas-codecov: Don't really need it, the `coverageStatus` type is a union of string literals `ERROR | PENDING | COMPLETED`, so it'll error and fail the build if you change the return type. Also using strings directly like this is overall better for compression and smaller bundle sizes

src/pages/RepoPage/PullsTab/PullsTable/PullsTable.spec.tsx
nicholas-codecov: No, not the default value. It's just reflective of the API response that's being mocked.

src/pages/RepoPage/PullsTab/PullsTable/createPullsTableData.tsx
ajay-sentry: > overall better for compression and smaller bundle sizes

If you have a source on this, would love to read more about the magnitude of gains here.

with respect to it erroring on the build by changing the return type, I'm sure there is a way to update the zod schema to account for this type on either end even with an enum.

In any case, just a thought that may improve readability for some

nicholas-codecov: Here's an example PR of swapping to regular strings from the object style enum I believe you're talking about: https://github.com/getsentry/sentry-javascript/pull/4280. 

A different concern with using the object syntax like we do in Gazebo (which I think we need to iterate on, and move away from) is that the field names themselves can not be minified. There are also issues with tree shaking that come up since we code split quite heavily so this also brings up issues with bundle sizes.

An alternative I'd be alright exploring would be something more along the lines of:

```ts
export const COMMIT_STATUS_COMPLETED = 'COMPLETED'
export const COMMIT_STATUS_ERROR = 'ERROR'
export const COMMIT_STATUS_PENDING = 'PENDING'

const CommitStatusSchema = z.union([
  z.literal(COMMIT_STATUS_COMPLETED),
  z.literal(COMMIT_STATUS_ERROR),
  z.literal(COMMIT_STATUS_PENDING),
])
```

ajay-sentry: Very neat! 2% savings definitely adds up over time, thanks for sharing that.

I think having a shared schema like that solves my original concern which is the improved readability, and we won't need the object syntax either. I'm all for it

src/pages/RepoPage/PullsTab/PullsTable/createPullsTableData.tsx
```diff
@@ -1,34 +1,67 @@
...
+      pull?.head?.bundleStatus === 'COMPLETED' &&
+      pull?.bundleAnalysisCompareWithBase?.__typename ===
+        'BundleAnalysisComparison'
     ) {
-      bundleAnalysis = <>Upload: &#x2705;</>
-    } else {
-      bundleAnalysis = <>Upload: &#x274C;</>
+      const change =
+        pull?.bundleAnalysisCompareWithBase?.bundleChange?.size.uncompress
+      const content = `${change > 0 ? '+' : ''}${formatSizeToString(change)}`
```
ajay-sentry, line 63: if the change < 0, should we show a "-" sign?

ajay-sentry (approved):

src/pages/RepoPage/PullsTab/PullsTable/createPullsTableData.tsx
nicholas-codecov: by default as a negative number the `-` is already present

ajay-sentry: ah okay makes sense, I wasn't sure since the "formatSizeToString" was taking the abs of the number

Looking at your screenshot that seems to be the case, just a little hard to parse atm

nicholas-codecov: Ahh, no worries. `formatSizeToString` only abs the size to for the size checks to see if we can concatenate to a larger size (kB, MB, etc.) we than use the passed value directly in the format, rather than the absolute'ed value.

PR #2936, opened 2024-06-10
Revert info architecture update

spalmurray-codecov:
Reverts the info architecture changes, just in case things go wrong in prod!


nicholas-codecov (approved):
Just in case

PR #2942, opened 2024-06-11
chore: Remove size delta references

nicholas-codecov:
# Description

Just removing references of `sizeDelta` as we are removing this from the API.

### DIFF
--- a/src/pages/CommitDetailPage/CommitBundleAnalysis/BundleMessage/BundleMessage.tsx
+++ b/src/pages/CommitDetailPage/CommitBundleAnalysis/BundleMessage/BundleMessage.tsx
@@ -47,9 +47,9 @@ const BundleMessage: React.FC = () => {
     comparison?.__typename === 'BundleAnalysisComparison' &&
     isNumber(comparison?.bundleChange?.size?.uncompress)
   ) {
-    const sizeDelta = comparison.bundleChange.size.uncompress
-    const positiveSize = Math.abs(sizeDelta)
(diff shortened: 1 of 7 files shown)


src/pages/CommitDetailPage/CommitBundleAnalysis/BundleMessage/BundleMessage.tsx
```diff
@@ -47,9 +47,9 @@ const BundleMessage: React.FC = () => {
     comparison?.__typename === 'BundleAnalysisComparison' &&
     isNumber(comparison?.bundleChange?.size?.uncompress)
   ) {
-    const sizeDelta = comparison.bundleChange.size.uncompress
-    const positiveSize = Math.abs(sizeDelta)
-    if (sizeDelta < 0) {
+    const uncompressDelta = comparison.bundleChange.size.uncompress
```
ajay-sentry, line 50: any reason why we called this const uncompressDelta vs uncompress? Same with another instance below

nicholas-codecov: Just to be a bit more explicit that it's a change in size instead of just a static size

ajay-sentry: Gotcha, thanks for the context!

ajay-sentry (approved):

PR #2947, opened 2024-06-13
feat: Add Reusable Alert Component 

ajay-sentry:
# Description

This PR aims to add a new reusable alert component using shadcn and cva. There were a few additional colors I had to add to our config file since we weren't using them prior, and I did update one red color that was already being used prior. It was only used in one other place so I didn't think we'd be affected much / at all from it.

There are some new stories highlighting what you can do with the component.

[Figma](https://www.figma.com/design/JOS1ZW499rfWP9nT7ivjf8/Pavilion%3A-Codecov-Design-System-Library?node-id=4318-3206&t=gkougYAOICDRz9AZ-0)

Closes https://github.com/codecov/engineering-team/issues/1815

# Screenshots

Screenshot 2024-06-13 at 2 00 03 PM

# Link to Sample Entry

# Legal Boilerplate


src/ui/Alert/Alert.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+import { cn } from 'shared/utils/cn'
+
+import successImage from './assets/check-circle.svg'
+import warningImage from './assets/exclamation.svg'
+import infoImage from './assets/information-circle.svg'
+import errorImage from './assets/x-circle.svg'
+
+const alertVariants = cva(
+  '[&>svg]:text-foreground relative w-full border-l-4 p-4 [&>svg+div]:translate-y-[-3px] [&>svg]:absolute [&>svg]:left-4 [&>svg]:top-4 [&>svg~*]:pl-7',
```
spalmurray-codecov, line 12: I've never seen selectors used like this! Super cool, had no idea tailwind supported this.

That said, I'm wondering why these styles are here and not on the `img` tag or something else lower down?

src/ui/Alert/Alert.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const AlertRoot = React.forwardRef<HTMLDivElement, RootProps>(
+  ({ className, variant, ...props }, ref) => (
+    <div
+      ref={ref}
+      role="alert"
+      className={cn(alertVariants({ variant }), className)}
+      {...props}
+    >
+      {variantToIcon(variant)}
```
spalmurray-codecov, line 66: This child spacing issue is hard. I'm thinking ask design what they think of aligning the text like you have it here.

If not, you could solve it with wrapper sub-components (like in our Card):
- AlertHeader (contains the svg and top text, provide a font-weight variant)
- AlertBody (contains the bottom text)

spalmurray-codecov: This feels like over-complication, but I can't think of a better solution atm. I think we just get design to agree to what you have lol

src/ui/Alert/Alert.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+const description = cva(['text-sm [&_p]:leading-relaxed'])
+
+interface DescriptionProps
+  extends React.HTMLAttributes<HTMLHeadingElement>,
+    VariantProps<typeof description> {}
+
+const Description = React.forwardRef<HTMLParagraphElement, DescriptionProps>(
+  ({ className, ...props }, ref) => (
+    <div ref={ref} className={cn(description({ className }))} {...props} />
```
spalmurray-codecov, line 95: You have this as a `HTMLParagraphElement` but the actual element is a `div`, not a `p`. This could lead to issues as the caller will assume it's a `p`. Is there a reason you did this? Might be a workaround

nicholas-codecov (commented):
We should already have these icons accessible through the `Icon` component or you can add them to the list [here](https://github.com/codecov/gazebo/blob/main/scripts/icons.js), here's the [x-circle](https://github.com/codecov/gazebo/blob/main/src/ui/Icon/svg/outline/x-circle.svg) for example. To colour them, iirc you can use the fill or text class (there's examples in the codebase).

src/ui/Alert/Alert.tsx
ajay-sentry: This is actually just some leftover cruft I forgot to remove when copying over the shadcn code!

src/ui/Alert/Alert.tsx
ajay-sentry: Already have sign off from design on the current look :) But agree with you it's not the most elegant atm. 

I've thought about using grids, since the main "issue" is the icon is persistent regardless of the header or not. So being able to manipulate the image such that its always present, but can "push" the header text if that text is present is something that I'm sure can be done trivially if my CSS knowledge was a bit better 😂

src/ui/Alert/Alert.tsx
ajay-sentry: tbh with you I just copied this over from the shadcn manual code. They also have a paragraphElement with the div here: https://ui.shadcn.com/docs/components/alert

admittedly, I haven't really looked into these types too much

src/ui/Alert/Alert.tsx
spalmurray-codecov: https://stackoverflow.com/questions/19179424/how-to-wrap-text-around-an-image-using-html-css

didn't know this was a thing. Could be possible for sure

src/ui/Alert/Alert.tsx
spalmurray-codecov: Oh! Well if they're doing it, it's probably fine to leave as is. If it ain't broke for Vercel...

nicholas-codecov: The alternative is we just make it a `p` element

spalmurray-codecov: yea @nicholas-codecov

ajay-sentry: Easy enough! Changed in [d874761](https://github.com/codecov/gazebo/pull/2947/commits/d8747618bf054230f4c5dea6ef1189750d03a24d)

src/ui/Alert/Alert.tsx
```diff
@@ -0,0 +1,122 @@
...
+      warning: 'border-orange-500 bg-orange-100',
+    },
+  },
+  defaultVariants: {
+    variant: 'info',
+  },
+})
+
+export function variantToIcon(variant?: string | null) {
+  let classname = 'float-left -mt-1 mr-2 align-middle'
```
spalmurray-codecov, line 22: I think this is a typo
```suggestion
  let classname = 'float-left mt-1 mr-2 align-middle'
```

src/ui/Alert/Alert.tsx
```diff
@@ -0,0 +1,122 @@
...
+      {...props}
+    >
+      {variantToIcon(variant)}
+      {props.children}
+    </div>
+  )
+)
+AlertRoot.displayName = 'Alert'
+
+const title = cva(['mb-1 font-semibold leading-none tracking-tight'])
```
spalmurray-codecov, line 96: Wondering why `leading-none` and `tracking-tight` I've not used these before. Why necessary here?

If the line height/letter spacing differ from design, I feel like this is likely a misalignment across the app between Figma and real. If that's the case we should align one way or the other.

src/ui/Alert/Alert.tsx
ajay-sentry: was intended actually! That's a -1 pixel, basically move the image slightly above so the middle is more inline with the middle of the text

src/ui/Alert/Alert.tsx
```diff
@@ -0,0 +1,122 @@
...
+interface TitleProps
+  extends React.HTMLAttributes<HTMLHeadingElement>,
+    VariantProps<typeof title> {}
+
+const Title = React.forwardRef<HTMLParagraphElement, TitleProps>(
+  ({ className, ...props }, ref) => (
+    // eslint-disable-next-line jsx-a11y/heading-has-content
+    <h5 ref={ref} className={cn(title({ className }))} {...props} />
+  )
+)
```
spalmurray-codecov, line 107: Why is this only an issue for heading lol.

Super huge nit, but I'd rather do this than have the ignore ngl:
```suggestion
const Title = React.forwardRef<HTMLParagraphElement, TitleProps>(
  ({ children, className, ...props }, ref) => (
    <h5 ref={ref} className={cn(title({ className }))} {...props} >{ children }</h5>
  )
)
```

src/ui/Alert/Alert.tsx
spalmurray-codecov: duuude I'm learning so much tailwind from this one PR this is wild

src/ui/Alert/Alert.tsx
ajay-sentry: Removed in [c04acc4](https://github.com/codecov/gazebo/pull/2947/commits/c04acc45d286db2ece698e606c2c0b353ab0e826)

spalmurray-codecov (approved):

src/ui/Alert/Alert.tsx
ajay-sentry: lol im pretty indifferent tbh, but it's super tiny, fixed in [b66d04b](https://github.com/codecov/gazebo/pull/2947/commits/b66d04b12f6a57b53325eec3e052d817bf4e74ba)

src/ui/Alert/Alert.tsx
ajay-sentry: This is actually my first gig using tailwind in production too LOL
Having to google a lot of my CSS -> Tailwind conversions

spalmurray-codecov: What if I told you this isn't my first gig using tailwind in production 🥲 

Keep learning stay humble 💯 😂

nicholas-codecov: @ajay-sentry read the tailwind docs ... they are amazing

PR #2957, opened 2024-06-19
[feat] Add user onboarding metrics tracking + mutation 

rohitvinnakota-codecov:
# Description

This closes https://github.com/codecov/engineering-team/issues/1830. 

Notion: https://www.notion.so/sentry/Eng-Design-Tracking-Codecov-user-signup-to-first-commit-conversion-rate-6097411516f84563820276aafe3436ea

The useStoreCodecovMetric hook calls the storeCodecovMetric mutation [defined in API
](https://github.com/codecov/engineering-team/issues/1829) There are already a few layers of validation to sanitize input and prevent duplicate entries. In this PR, another check is added so as not to overcrowd the number of client network requests. 

We store all added metrics in `localstorage` and do not make a request if the client sees that we have already added that metric. The size of what can go into the `localstorage` object is also restricted to 30 entires to be reasonable with space constraints. 

image

# Legal Boilerplate

### DIFF
--- a/src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
+++ b/src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
@@ -33,7 +33,7 @@ interface StoreEventMetricMutationArgs {
 }
 
 const LOCAL_STORAGE_KEY = 'UserOnboardingMetricsStored'
-const MAX_ENTRIES = 50
+const MAX_ENTRIES = 30
 
 const generateMetricString = (
   owner: string,
--- a/src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
+++ b/src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
@@ -1,0 +1,103 @@
+import { useMutation } from '@tanstack/react-query'
+import { useParams } from 'react-router-dom'
+
+import Api from 'shared/api'
+
+const mutationQuery = `
+  mutation storeEventMetric(
+    $input: StoreEventMetricsInput!
+  ) {
+    storeEventMetric(
+      input: $input
+    ) {
+      error {
+        ... on UnauthenticatedError {
+          message
+        }
+        ... on ValidationError {
+          message
+        }
+      }
+    }
+  }
+`
+
+interface Params {
+  provider: string
+}
+
+interface StoreEventMetricMutationArgs {
+  owner: string
+  event: string
+  jsonPayload: object
+}
+
+const LOCAL_STORAGE_KEY = 'UserOnboardingMetricsStored'
+const MAX_ENTRIES = 50
+
+const generateMetricString = (
+  owner: string,
+  event: string,
+  jsonPayload: string
+) => {
+  return `${owner}|${event}|${jsonPayload}`
+}
+
+const getStoredMetrics = (): string[] => {
+  const storedMetrics = localStorage.getItem(LOCAL_STORAGE_KEY)
+  return storedMetrics ? JSON.parse(storedMetrics) : []
+}
+
+const isMetricInLocalStorage = (metricString: string) => {
+  const storedMetrics = getStoredMetrics()
+  return storedMetrics.includes(metricString)
+}
+
+const addMetricToLocalStorage = (metricString: string) => {
+  let storedMetrics = getStoredMetrics()
+
+  // limit size of metrics stored to ensure we are in reasonable localstorage limits
+  if (storedMetrics.length >= MAX_ENTRIES) {
+    storedMetrics = storedMetrics.slice(1)
+  }
+
+  storedMetrics.push(metricString)
+  localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(storedMetrics))
+}
+
+export const useStoreCodecovEventMetric = () => {
+  const { provider } = useParams<Params>()
+
+  const mutation = useMutation({
+    mutationFn: async ({
+      owner,
+      event,
+      jsonPayload,
+    }: StoreEventMetricMutationArgs) => {
+      const jsonPayloadString = JSON.stringify(jsonPayload)
+      const metricString = generateMetricString(owner, event, jsonPayloadString)
+      if (isMetricInLocalStorage(metricString)) {
+        return
+      }
+
+      const variables = {
+        input: {
+          orgUsername: owner,
+          eventName: event,
+          jsonPayload: jsonPayloadString,
+        },
+      }
+
+      addMetricToLocalStorage(metricString)
+
+      return Api.graphqlMutation({
+        provider,
+        query: mutationQuery,
+        variables,
+        mutationPath: 'storeEventMetric',
+      })
+    },
+  })
+
+  return mutation
+}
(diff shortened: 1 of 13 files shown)


src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
```diff
@@ -0,0 +1,103 @@
...
+  let storedMetrics = getStoredMetrics()
+
+  // limit size of metrics stored to ensure we are in reasonable localstorage limits
+  if (storedMetrics.length >= MAX_ENTRIES) {
+    storedMetrics = storedMetrics.slice(1)
+  }
+
+  storedMetrics.push(metricString)
+  localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(storedMetrics))
+}
```
spalmurray-codecov, line 66: what is the purpose of adding metrics to local storage? Was thinking maybe for batching requests, but it doesn't seem like that's what we're doing here

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
```diff
@@ -0,0 +1,103 @@
...
+  }
+`
+
+interface Params {
+  provider: string
+}
+
+interface StoreEventMetricMutationArgs {
+  owner: string
+  event: string
```
spalmurray-codecov, line 31: making event a union of string literals would be helpful IMO. It was nice having the type-safety with @nicholas-codecov's Sentry metrics wrapper. Not 0 work though, so could be convinced we don't need this at this point.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
```diff
@@ -0,0 +1,103 @@
...
+        input: {
+          orgUsername: owner,
+          eventName: event,
+          jsonPayload: jsonPayloadString,
+        },
+      }
+
+      addMetricToLocalStorage(metricString)
+
+      return Api.graphqlMutation({
```
spalmurray-codecov, line 93: At my last job, we rolled our own metrics for a time. One problem we ran into was metrics not getting sent for hard navigations and redirects because the redirect would happen before the request could be sent. Have you tested at all with that? Does TS query handle this?

IIRC, our solution was to use `sendBeacon`.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
rohitvinnakota-codecov: There is some validation at the [service level](https://github.com/codecov/shared/blob/f9c0acde3cfc54c72f42f812b0212a5066429e89/shared/django_apps/codecov_metrics/service/codecov_metrics.py). 

The reason I did not add types here is because we will probably add more metrics in the future, and my thinking is it more upfront maintenance to have parity between the service and here. 

Happy to revisit at a later date when we have a comprehensive list that's finalized.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
rohitvinnakota-codecov: It's so we don't call GQL multiple times unnecessarily. Once something is logged, we store the fact that it _was_ logged and do not make additional network calls.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
rohitvinnakota-codecov: I tested collection on every page and the DB updated as expected. This is good information and I will keep it in mind should something not appear to be working. 🙏

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
spalmurray-codecov: Fair enough! Could we return a 4xx on bad event type instead of just logging a warning? This way the constraints are strictly enforced and we can't end up with random stuff in the DB. Also makes it clear that you've made a typo in the FE when setting up metrics 😅

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
spalmurray-codecov: Sounds good!

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
spalmurray-codecov: Oh I totally missed the early `return` in the bad event type path. That works.

Yea I think it would be great to have the type safety here eventually. I'm happy with you making an issue and returning to this later! 👍

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
```diff
@@ -0,0 +1,103 @@
...
+  storedMetrics.push(metricString)
+  localStorage.setItem(LOCAL_STORAGE_KEY, JSON.stringify(storedMetrics))
+}
+
+export const useStoreCodecovEventMetric = () => {
+  const { provider } = useParams<Params>()
+
+  const mutation = useMutation({
+    mutationFn: async ({
+      owner,
```
spalmurray-codecov, line 73: Where did we end up on tracking owner and user? Last I can recall, we were going to track both, but not require either.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
rohitvinnakota-codecov: Right now we are only storing the first instance of events(using `get_or_create` on the backend) so it will be a redundant call. Intentionally doing this both to reduce a lot of DB writes as well as focusing on what's important in the user journey. 

This can be extended in the future by customizing payload information...but I think we'd need to revisit the design to see if just using postgres is scalable for our use case if we are doing a large amount of collection.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
rohitvinnakota-codecov: We're just using org id at the moment(required). User can be supplied if need but is optional.

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
rohitvinnakota-codecov: https://github.com/codecov/engineering-team/issues/1998

src/services/codecovEventMetrics/useStoreCodecovEventMetric.tsx
spalmurray-codecov: Ok sounds good. Another question I just thought of:

How do we plan to handle data from existing orgs? 

For example, this code gets merged, Codecov org is already onboarded, but the next time we setup a new repo, these data points will be recorded for the first time. Not necessarily something we need to have an answer for right now, more just something that could be important to keep in mind as we build our queries.

rohitvinnakota-codecov: We could do a join on the org table and filter by creation date. 

It will need some custom querying for sure so I will raise that point when sharing with the team, thanks.

spalmurray-codecov (approved):
let's do it 🚀

PR #2960, opened 2024-06-20
fix: Hide Address 2 if null

ajay-sentry:
# Description

Saw address 2 was coming back as null and showing on the UI, we want to hide it.

# Screenshots

**BEFORE**
Screenshot 2024-06-20 at 1 48 16 PM

**AFTER**
Screenshot 2024-06-20 at 1 48 02 PM

# Link to Sample Entry

# Legal Boilerplate


adrian-codecov (approved):
Would line1 ever also be null?

nora-codecov (approved):

ajay-sentry:
@adrian-codecov No it won't, only line 2 has that possibility, at least per the Address Element spec. 
Screenshot 2024-06-20 at 2 01 10 PM

PR #2961, opened 2024-06-20
fix: Remove Cookie Session Expiry cookie when expired and redirect

ajay-sentry:
# Description

This stemmed from thinking through what is causing the redirect, which is the presence of the expiration cookie timestamp.

That cookie is added on login here: https://github.com/codecov/codecov-api/blob/c3f5e7f903c63f8a5eadc5a69a2846df57e65a0a/codecov_auth/views/github.py#L169-L171 in API on login, and only for Github

Because we're already setting a "session expired" key here: https://github.com/codecov/gazebo/blob/902fa2fcff89b96d1fc0ab470a22b7088b21fc61/src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx#L74 and that key is removed and reset only when the user logs in, since Session Expiry Tracker is only present on the BaseLayout here: https://github.com/codecov/gazebo/blob/5d142294fee83428c92ca4f33a8861a7e15b4dde/src/layouts/BaseLayout/BaseLayout.jsx#L72

My hunch is that we shouldn't face any regression here since the Login page uses a different layout: 

https://github.com/codecov/gazebo/blob/b07fcc7bae68f7a015dac9804f8649a97b6b5859/src/App.tsx#L72-L74

Closes https://github.com/codecov/engineering-team/issues/1994

**THE FIX:**

This ultimate fix calls is to call the logout API to clear the user's session, as well as removing the session-expiry cookie at the same time. This also fixes the issue where users who had logged in prior had to login again to view other user's codecov orgs. Now they will see the guest header and navigate freely without logging in

# Screenshots

Couple videos prior to adding query param stuff (but showing the logic works)

https://github.com/codecov/gazebo/assets/159853603/2d5264ac-c678-45a5-a6fe-cff7d73721d2

https://github.com/codecov/gazebo/assets/159853603/e76d2373-c21f-4ff1-9fea-611c747952d8

**NON-ENTERPRISE**
**QUERY PARAM**
Screenshot 2024-06-21 at 9 45 46 AM
**NO QUERY PARAM**
Screenshot 2024-06-21 at 9 45 57 AM
**ENTERPRISE**
**QUERY PARAM**
Screenshot 2024-06-21 at 9 52 11 AM
**NO QUERY PARAM**
Screenshot 2024-06-21 at 9 52 20 AM

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/layouts/EnterpriseLoginLayout/EnterpriseLoginLayout.tsx
+++ b/src/layouts/EnterpriseLoginLayout/EnterpriseLoginLayout.tsx
@@ -1,4 +1,5 @@
 import { Suspense } from 'react'
+import { useLocation } from 'react-router-dom'
 
 import { LOCAL_STORAGE_SESSION_EXPIRED_KEY } from 'config'
 
@@ -19,13 +20,17 @@ const FullPageLoader = () => (
 )
 
 function EnterpriseLoginLayout({ children }: { children: React.ReactNode }) {
+  const location = useLocation()
+
   const showExpiryBanner = localStorage.getItem(
     LOCAL_STORAGE_SESSION_EXPIRED_KEY
   )
   return (
     <>
       <Header />
-      {showExpiryBanner && <SessionExpiredBanner />}
+      {(location.search.includes('expired') || showExpiryBanner) && (
+        <SessionExpiredBanner />
+      )}
       <Suspense fallback={<FullPageLoader />}>
         <ErrorBoundary sentryScopes={[['layout', 'base']]}>
           <NetworkErrorBoundary>
--- a/src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
+++ b/src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
@@ -3,6 +3,7 @@ import React, { useCallback, useEffect, useState } from 'react'
 import { Redirect } from 'react-router-dom'
 
 import config, {
+  COOKIE_SESSION_EXPIRY,
   COOKIE_SESSION_ID,
   LOCAL_STORAGE_SESSION_EXPIRED_KEY,
   LOCAL_STORAGE_SESSION_TRACKING_KEY,
@@ -17,7 +18,7 @@ const SessionExpiryTracker: React.FC = () => {
   localStorage.removeItem(LOCAL_STORAGE_SESSION_EXPIRED_KEY)
 
   const [redirectToLogout, setRedirectToLogout] = useState(false)
-  const sessionExpiryTimeString = Cookies.get('session_expiry')
+  const sessionExpiryTimeString = Cookies.get(COOKIE_SESSION_EXPIRY)
   const getCheckDelay = (sessionExpiryTime: Date) => {
     const timeLeft = sessionExpiryTime.getTime() - new Date().getTime()
     return timeLeft > THIRTY_MINUTES_MILLIS
@@ -69,6 +70,7 @@ const SessionExpiryTracker: React.FC = () => {
   }
 
   Cookies.remove(COOKIE_SESSION_ID, { path: '' })
+  Cookies.remove(COOKIE_SESSION_EXPIRY)
   localStorage.setItem(LOCAL_STORAGE_SESSION_EXPIRED_KEY, 'true')
   localStorage.removeItem(LOCAL_STORAGE_SESSION_TRACKING_KEY)
 
--- a/src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
+++ b/src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
@@ -8,8 +8,8 @@ import config, {
   LOCAL_STORAGE_SESSION_TRACKING_KEY,
 } from 'config'
 
-const ONE_MINUTE_MILLIS = 6 * 1000
-const TWO_MINUTES_MILLIS = 2 * 60 * 1000 // ONE_MINUTE_MILLIS
+const ONE_MINUTE_MILLIS = 60 * 1000
+const TWO_MINUTES_MILLIS = 2 * ONE_MINUTE_MILLIS
 const THIRTY_MINUTES_MILLIS = 30 * ONE_MINUTE_MILLIS
 
 const SessionExpiryTracker: React.FC = () => {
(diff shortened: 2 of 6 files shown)


src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
```diff
@@ -3,21 +3,21 @@ import React, { useCallback, useEffect, useState } from 'react'
...
 import config, {
-  COOKIE_SESSION_ID,
+  COOKIE_SESSION_EXPIRY,
   LOCAL_STORAGE_SESSION_EXPIRED_KEY,
   LOCAL_STORAGE_SESSION_TRACKING_KEY,
 } from 'config'
 
-const ONE_MINUTE_MILLIS = 60 * 1000
-const TWO_MINUTES_MILLIS = 2 * ONE_MINUTE_MILLIS
+const ONE_MINUTE_MILLIS = 6 * 1000
```
spalmurray-codecov, line 11: mistake here - left from testing I'd imagine

spalmurray-codecov (approved):
Nice! I was trying to achieve something similar the other day but ran into edge cases with the redirect. Fingers crossed your fix works!

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
```diff
@@ -64,19 +65,29 @@ const SessionExpiryTracker: React.FC = () => {
...
 
-  Cookies.remove(COOKIE_SESSION_ID, { path: '' })
-  localStorage.setItem(LOCAL_STORAGE_SESSION_EXPIRED_KEY, 'true')
-  localStorage.removeItem(LOCAL_STORAGE_SESSION_TRACKING_KEY)
-
-  return config.IS_SELF_HOSTED ? (
-    <Redirect to={'/'} />
-  ) : (
-    <Redirect to={'/login'} />
-  )
```
nicholas-codecov, line 79: What was the reasoning behind removing the two redirects?

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
ajay-sentry: As in the title, still a draft that I was hoping we could try and figure out the banner thing before I cleaned up :)

spalmurray-codecov: Ahhh gotcha.

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
ajay-sentry: Mentioned on slack, will post here as well:

The Root Cause (to me) looks to be the persistence of the `sessionid` cookie even after the user was allegedly "logged out." The user was getting pushed to the login page thanks to the redirect at the bottom, and was being "locked" there because of the presence of the `session-expiry` cookie forcing the user to "login" and refresh that `session-expiry` cookie.
Now that we're actually removing the `sessionid` cookie, we can also remove the `session-expiry` cookie, leaving the user in a "clean" state, so on their next login we add both cookies back and can avoid infinite redirects.

This redirect is superfluous from the root cause and can be removed to reduce complexity in this part of the code

src/layouts/EnterpriseLoginLayout/EnterpriseLoginLayout.tsx
```diff
@@ -19,13 +20,17 @@ const FullPageLoader = () => (
...
+  const location = useLocation()
+
   const showExpiryBanner = localStorage.getItem(
     LOCAL_STORAGE_SESSION_EXPIRED_KEY
   )
   return (
     <>
       <Header />
-      {showExpiryBanner && <SessionExpiredBanner />}
+      {(location.search.includes('expired') || showExpiryBanner) && (
```
spalmurray-codecov, line 31: I'd prefer to get the qparam by key and then check value. I don't see how this would happen but with this it's possible for `/login?branch=fix-session-expired` to trigger the banner. Just would rather keep it super explicit. If you disagree, I'll survive 😂

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
```diff
@@ -87,7 +86,11 @@ const SessionExpiryTracker: React.FC = () => {
     return null
   }
 
-  return <Redirect to={'/login'} />
+  return config.IS_SELF_HOSTED ? (
+    <Redirect to={'/?session=expired'} />
+  ) : (
+    <Redirect to={'/login?session=expired'} />
```
spalmurray-codecov, line 92: fair enough

spalmurray-codecov (approved):

src/layouts/EnterpriseLoginLayout/EnterpriseLoginLayout.tsx
ajay-sentry: lmfao that's a fair take and a good call out, tbh I'm too lazy to update all the tests and retest everything with the explicit query param 🙃

nicholas-codecov: +1 to that, we do have a query string package to help set and read query params.

src/ui/SessionExpiryTracker/SessionExpiryTracker.tsx
```diff
@@ -64,18 +64,32 @@ const SessionExpiryTracker: React.FC = () => {
...
-  Cookies.remove(COOKIE_SESSION_ID, { path: '' })
-  localStorage.setItem(LOCAL_STORAGE_SESSION_EXPIRED_KEY, 'true')
-  localStorage.removeItem(LOCAL_STORAGE_SESSION_TRACKING_KEY)
-
   return config.IS_SELF_HOSTED ? (
-    <Redirect to={'/'} />
+    <Redirect to={'/?session=expired'} />
   ) : (
-    <Redirect to={'/login'} />
+    <Redirect to={'/login?session=expired'} />
```
nicholas-codecov, line 92: Thoughts on using our query string helper here to set these, a bit more explicit and also adds in browser compat in case there are any issues

PR #2967, opened 2024-06-24
feat: Hide old header stuff when new header flag is true

spalmurray-codecov:
Hides all primary and secondary Navs along with old header whenever the `newHeader` feature flag is `true`. This is prep for implementing the new header/nav and will not affect any users until the proper rollout of the feature.

Closes https://github.com/codecov/engineering-team/issues/1935
Closes https://github.com/codecov/engineering-team/issues/1951


calvin-codecov (approved):
There are a couple instances of describe text blocks that are using the spelling "flaging" instead of "flagging".

Looks great!

Side question out of curiosity about the feature changes: so we're removing breadcrumbs with the new header?

PR #2969, opened 2024-06-25
feat: Create BundleTrendChart

nicholas-codecov:
# Description

This PR adds in a new `ui/BundleTrendChart` component which will be added into the bundle tab shortly providing trend data for users and their bundles.

Closes codecov/engineering-team#1801

# Notable Changes

- Add in `ui/BundleTrendChart`, CSS, and `NoData` component
- Add in tests and stories
- Add in new helper function

# Screenshots

Screenshot 2024-06-25 at 07 53 44

### DIFF
--- a/src/shared/utils/bundleAnalysis.spec.ts
+++ b/src/shared/utils/bundleAnalysis.spec.ts
@@ -1,4 +1,8 @@
-import { formatSizeToString, formatTimeToString } from './bundleAnalysis'
+import {
+  formatSizeToString,
+  formatTimeToString,
+  localizeBundleSize,
+} from './bundleAnalysis'
 
 describe('formatSizeToString', () => {
   describe('size is less then one kilobyte', () => {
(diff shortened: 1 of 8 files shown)


src/ui/BundleTrendChart/BundleTrendChart.tsx
spalmurray-codecov, line 1: I feel like this should belong in the BA page directory instead of `src/ui`. lmk what you think.

src/shared/utils/bundleAnalysis.spec.ts
```diff
@@ -97,3 +101,69 @@ describe('formatTimeToString', () => {
...
+    describe('size is a negative number', () => {
+      it('returns size in kilobytes', () => {
+        const result = localizeBundleSize(-10000)
+        expect(result).toBe(-10)
+      })
+    })
+  })
+
+  describe('size is greater then one megabyte and smaller then gigabyte', () => {
+    describe('size is a positive number', () => {
```
adrian-codecov, line 139: I like this nested describes to better ... describe a test 👏

spalmurray-codecov (approved):

src/shared/utils/bundleAnalysis.ts
```diff
@@ -57,3 +57,21 @@ export const formatTimeToString = (milliseconds: number) => {
...
+
+export const localizeBundleSize = (bytes: number) => {
+  const positiveBytes = Math.abs(bytes)
+
+  if (positiveBytes < KILOBYTE) {
+    return bytes
+  }
+
+  if (positiveBytes >= KILOBYTE && positiveBytes < MEGABYTE) {
+    return bytes / KILOBYTE
```
adrian-codecov, line 69: We're okay with decimal numbers for this fn yeah?

spalmurray-codecov: I believe this should always return an integer unless the input is not an integer. And I think the input will always be an integer

spalmurray-codecov: Unless there's some weird JS stuff going on lol

src/ui/BundleTrendChart/BundleTrendChart.tsx
nicholas-codecov: I have a sneaking suspicion that we may end up extending the analytics page soon with bundle an test analytics stuff, so we'll want to re-use it there, so I'll probably keep it in `src/ui` so it's "easier" to pull into multiple places.

src/shared/utils/bundleAnalysis.spec.ts
nicholas-codecov: I mean ... the top level is more the weight(?) of the number ... rather than the actual sign of the number 🤦‍♂️

src/shared/utils/bundleAnalysis.ts
nicholas-codecov: We only ever pass in integer values here, however we could get decimals returned which is fine, as we use rounding, ceil, floor, and Intl.NumberFormat to handle the values and remove the extra decimals

adrian-codecov: cool cool, yeah in like 3/2 you'd get 1.5, so wanted to confirm we wanted this, which makes sense anyway

adrian-codecov (approved):

src/shared/utils/bundleAnalysis.ts
nicholas-codecov: if you look at the top of that file we have some consts defined so we're only dividing by multiples of 10 only moving the decimal place

src/shared/utils/bundleAnalysis.spec.ts
adrian-codecov: Haha I think you misunderstood what I said, I was complimenting the test and meant more like "Oh! I like how we have nested describes to better inform us what test is about!"

PR #2970, opened 2024-06-25
fix: send cardholder name when editing billing address

calvin-codecov:
# Description

We are not currently sending through the "Cardholder name" to be changed so it always ends up being `null` after updating the information.

Closes https://github.com/codecov/engineering-team/issues/2013 with a corresponding API change.

Previous change: https://github.com/codecov/codecov-api/commit/6efefafc7866b4b16214d45bd2bddaa76a4926fd#diff-5ab8e0d637ba546946e51547b9b585671a50592c7d35a75eb63cb8058864beeb

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


spalmurray-codecov (approved):
Nice send it!

PR #2972, opened 2024-06-27
ref: Convert useUser to TS

spalmurray-codecov:
Converts useUser hook to TS and cleans up failing tests.


src/services/user/useUser.ts
```diff
@@ -0,0 +1,166 @@
...
+  email
+  privateAccess
+  onboardingCompleted
+  businessEmail
+  termsAgreement
+  user {
+    name
+    username
+    avatarUrl
+    avatar: avatarUrl
```
rohitvinnakota-codecov, line 93: Are these the same? Do we need both?

rohitvinnakota-codecov (approved):

src/services/user/useUser.ts
spalmurray-codecov: They are the same. We had both in the old query and so I decided to just keep them. It seems like we use both. I will make a ticket to go back and consolidate this though as it's definitely no ideal lol

PR #2973, opened 2024-06-27
feat: Implement new header's User dropdown

spalmurray-codecov:
Implements the UserDropdown for the new Header. For now, this is an almost exact copy the old user dropdown, but we'll move this over to the new Dropdown component once we implement that in https://github.com/codecov/engineering-team/issues/1942. The only difference between this and the old user dropdown is minor spacing change and dark border on the Avatar.

[Design](https://www.figma.com/design/kbJfmcDgAuptsJR5KB28O9/GH-293-(nav%2C-header%2C-and-related-elements)?node-id=322-1385&t=oCWsofSFCNg0QXiK-0)

Closes https://github.com/codecov/engineering-team/issues/1952

## Screenshot
Screenshot 2024-07-04 at 08 35 03

### DIFF
--- a/src/layouts/BaseLayout/BaseLayout.spec.jsx
+++ b/src/layouts/BaseLayout/BaseLayout.spec.jsx
@@ -402,7 +402,7 @@ describe('BaseLayout', () => {
         wrapper: wrapper(),
       })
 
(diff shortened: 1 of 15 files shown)


nicholas-codecov (approved):
LGTM

PR #2975, opened 2024-06-28
fix: refresh user query on continue from org selection

calvin-codecov:
# Description

Redirect on clicking "Continue" button was technically working and rerouting correctly but the new user info after mutating the user's -> owner -> defaultOrgUsername wasn't being refreshed so the code still thought we needed to show the org selector component. As it exists, a refresh after clicking continue would reflect the correct behavior. With this change to invalidate the query, we now will get the new user data and change the UI accordingly.

Closes https://github.com/codecov/engineering-team/issues/1842

# Code Example

# Notable Changes

# Screenshots

https://github.com/codecov/gazebo/assets/170470397/c9c76dbf-531e-499b-ae3d-305da6ebda08

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/services/defaultOrganization/useUpdateDefaultOrganization.spec.tsx
+++ b/src/services/defaultOrganization/useUpdateDefaultOrganization.spec.tsx
@@ -68,12 +68,8 @@ describe('useUpdateDefaultOrganization', () => {
           result.current.data.data.updateDefaultOrganization.username
         await waitFor(() => expect(username).toBe('Gilmore'))
 
-        // expect(invalidateQueries).toHaveBeenLastCalledWith({
-        //   queryKey: ['currentUser'],
-        // })
         expect(invalidateQueries).toHaveBeenCalledTimes(2)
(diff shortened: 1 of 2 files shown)


src/services/defaultOrganization/useUpdateDefaultOrganization.spec.tsx
```diff
@@ -60,12 +60,21 @@ describe('useUpdateDefaultOrganization', () => {
...
+        const invalidateQueries = jest.spyOn(queryClient, 'invalidateQueries')
 
         await waitFor(() => expect(result.current.isSuccess).toBeTruthy())
 
         const username =
           result.current.data.data.updateDefaultOrganization.username
         await waitFor(() => expect(username).toBe('Gilmore'))
+
+        // expect(invalidateQueries).toHaveBeenLastCalledWith({
+        //   queryKey: ['currentUser'],
```
spalmurray-codecov, line 72: Comments can be removed probably :)

spalmurray-codecov (approved):
Nice!

PR #2988, opened 2024-07-05
fix: Remove session expiry tracker

spalmurray-codecov:
Removes the SessionExpiryTracker code that requires users to log back in every eight hours. Corresponding PR API side to come, but this can be merged without issue.

We will be replacing this behaviour with something else in a future PR.

Also adds code to clean up old Cookie and localStorage items. To be removed later.

Part of https://github.com/codecov/engineering-team/issues/2025


nicholas-codecov (approved):
Looks all good to me, just one question, what is the expiry value for `COOKIE_SESSION_EXPIRY`? Wondering if we should leave in a clean up for a bit of time to remove the cookie from the users browser.

nicholas-codecov (approved):
LGTM

PR #2994, opened 2024-07-09
chore: Remove spotlight

nicholas-codecov:
# Description

This PR removes Spotlight, as we're not really using it, and just looking to clean up some stuff.

### DIFF
--- a/src/sentry.ts
+++ b/src/sentry.ts
@@ -78,7 +78,6 @@ export const setupSentry = ({
   const integrations = [replay, tracing]
 
   // Only show feedback button in production
-  // spotlight takes the place of the feedback widget in dev mode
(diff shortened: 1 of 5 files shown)


RulaKhaled (approved):

src/sentry.ts
```diff
@@ -130,11 +129,4 @@ export const setupSentry = ({
     },
     ...deClutterConfig,
   })
-
```
ajay-sentry, line 133: there's 1 last reference to spotlight here: https://github.com/codecov/gazebo/blob/007ba8652703d30f1612751397ba6cc6a7873e61/src/sentry.ts#L82 if you have a chance to delete that!

ajay-sentry (approved):

PR #3004, opened 2024-07-11
ref: Rework RepoBreadcrumbContext so it's available in new header

spalmurray-codecov:
Refactors RepoBreadcrumbContext such that it is initialized above the Header on BaseLayout. The reason for this is because we are now rendering the breadcrumb in the header, we need access to the breadcrumb context up in the header.

Screenshot 2024-07-11 at 09 53 30


src/layouts/BaseLayout/BaseLayout.jsx
```diff
@@ -73,25 +74,27 @@ function BaseLayout({ children }) {
...
       <Suspense fallback={<FullPageLoader />}>
-        <ErrorBoundary sentryScopes={[['layout', 'base']]}>
-          <NetworkErrorBoundary>
-            <main className="container mb-8 mt-2 flex grow flex-col gap-2 md:p-0">
-              <GlobalBanners />
-              <OnboardingOrChildren>{children}</OnboardingOrChildren>
-            </main>
-          </NetworkErrorBoundary>
-        </ErrorBoundary>
+        <RepoBreadcrumbProvider>
```
spalmurray-codecov, line 78: We need access to the breadcrumb context inside of Header on line 82, so I've moved it up from RepoPage to here.

src/layouts/Header/components/Navigator/Navigator.tsx
```diff
@@ -0,0 +1,17 @@
+import { useRouteMatch } from 'react-router-dom'
+
+import { useCrumbs } from 'pages/RepoPage/context'
+import Breadcrumb from 'ui/Breadcrumb'
+
+function Navigator() {
```
spalmurray-codecov, line 6: This is still a work in progress.

src/pages/RepoPage/RepoPage.tsx
```diff
@@ -245,6 +246,27 @@ function RepoPage() {
       refetchOnWindowFocus: refetchEnabled,
     },
   })
+  const { setBaseCrumbs } = useCrumbs()
```
spalmurray-codecov, line 249: We don't have the Repo until the RepoPage is visited, so we populate the repo in the context only once we get here.

src/pages/RepoPage/context.tsx
```diff
@@ -0,0 +1,53 @@
...
+      }}
+    >
+      {children}
+    </RepoBreadcrumbContext.Provider>
+  )
+}
+
+export function useCrumbs() {
+  return useContext(RepoBreadcrumbContext)
+}
```
nicholas-codecov, line 53: Can we add a check in here to ensure that the context is actually being provided? You can see an example of that in  [here](https://github.com/codecov/gazebo/blob/main/src/ui/TopBanner/TopBanner.tsx#L58-L70) tho it is a bit more complex with the use of a zod schema, you can probably skip that part, but throwing an error if it doesn't exist is a nice way of ensuring that people don't misuse this.

src/pages/RepoPage/context.tsx
```diff
@@ -0,0 +1,53 @@
...
+  text?: string
+  children?: JSX.Element
+  readOnly?: boolean
+}
+
+const RepoBreadcrumbContext = createContext<{
+  setBaseCrumbs: (crumbs: Breadcrumb[]) => void
+  breadcrumbs: Breadcrumb[]
+  setBreadcrumbs: (crumbs: Breadcrumb[]) => void
+}>({
```
nicholas-codecov, line 15: `l`: Thoughts on pulling this out into it's own type, may also be able to import it incase anyone needs the type if there's something funky going on with the context?

nicholas-codecov (approved):
Two things too peak at, looks all good

src/pages/RepoPage/context.tsx
spalmurray-codecov: yea good idea!

PR #3010, opened 2024-07-12
ref: Convert useMyContexts to TS

spalmurray-codecov:
Converts useMyContexts to TS.


src/services/user/useMyContexts.ts
```diff
@@ -0,0 +1,113 @@
...
+      }),
+    select: ({ pages, pageParams }) => {
+      const me = pages[0]?.me ?? null
+      const myOrganizations = pages.map((page) => page?.me?.myOrganizations)
+      const flatOrganizations = myOrganizations.flatMap((page) =>
+        mapEdges(page)
+      )
+
+      return {
+        pages: pages,
```
ajay-sentry, line 98: nit: these (pages and pageParams) are implicit so you don't need to do pages: pages, could just do pages

ajay-sentry (approved):
lgtm!

PR #3016, opened 2024-07-15
dev: Use Yarn 4.3.1 instead of npm

ajay-sentry:
# Description

Transition off of npm and onto yarn. This PR updates all instances of npm in the README as well as the ci commands and local commands to use yarn.

A couple things I had to do to get this to work:
- globally installed yarn with `npm i -g yarn.`
- Pinned yarn version with volta
- Update everywhere that says "npm run" to yarn
- When booting up the app had some runtime errors, which is where all the package updates came from. The only "downgrade" in terms of package versioning was ["react-intersection-observer"
](https://github.com/thebuilder/react-intersection-observer/releases?page=2) due to some breaking changes with the imports in v9.5 that I'm not sure how slipped through the cracks.

Side things:

- Remove the netlify.toml since we don't use that anymore. Shoutout @nicholas-codecov 
- Deleted package-lock.json since we don't need that anymore
- Added some .gitignore stuff for things we don't want to commit on the yarn side (and no more package-lock.json)

Closes https://github.com/codecov/engineering-team/issues/2089

# Notable Changes AKA How does this affect me?

Main things:

- install yarn globally with `npm i -g yarn` (Not needed if using Volta)
- Delete your node_modules
- run `yarn` to reinstall node_modules
- run `yarn start` or `yarn dev` to start gazebo!

[There's a list of commands from npm -> yarn here](https://classic.yarnpkg.com/lang/en/docs/migrating-from-npm/)

References:

- [Yarn docs](https://classic.yarnpkg.com/lang/en/docs/)
- [Gitignore stuff
](https://yarnpkg.com/getting-started/qa#which-files-should-be-gitignored)

# Link to Sample Entry

# Legal Boilerplate


.github/workflows/ci.yml
```diff
@@ -42,9 +42,8 @@ jobs:
       #        if: ${{ steps.cache-node-modules.outputs.cache-hit != 'true' }}
 
       - name: Install packages
-        if: ${{ steps.cache-node-modules.outputs.cache-hit != 'true' }}
```
ajay-sentry, line 45: I need to add this line of code back after we merge this, wanted to just trigger the yarn installs while I was still making updates

.npmrc
```diff
@@ -1 +0,0 @@
-engine-strict=false
```
ajay-sentry, line 1: This is now at the bottom of the package.json

.github/workflows/ci.yml
nicholas-codecov: If the lock file is part of the cache key, how are we getting hits? I would assume that the yarn lock file would be slightly different? Another option would be to slightly tweak the cache key to something like: `${{ runner.os }}-${{ env.cache-name }}-yarn-${{ hashFiles('**/yarn.lock') }}`.

.yarnrc.yml
```diff
@@ -0,0 +1,2 @@
+yarnPath: .yarn/releases/yarn-4.3.1.cjs
+nodeLinker: node-modules
```
nicholas-codecov, line 2: `\n` police 🚨

nicholas-codecov (commented):
Some comments, also wondering if you've chatted with infra about the deploy pipelines and moving them over to yarn?

.yarnrc.yml
ajay-sentry: lol updated in [ad5b9dd](https://github.com/codecov/gazebo/pull/3016/commits/ad5b9ddc57d44448bd2648038482a1d0f8de0b71)

nicholas-codecov (approved):

PR #3019, opened 2024-07-16
feat: Move bundle tab trend dropdown

nicholas-codecov:
# Description

This PR adjusts the trend dropdown on the bundles tab to be inline with the toggle chart button so that the user can still change the trend which applies to the assets table. Also included is a small tweak to remove passing an artificial date and instead pass `null` which the API will interpret as the first data point.

# Notable Changes

- Rework `ToggleElement` to allow element to be placed inline
- Move `TrendDropdown` from `BundleChart` to `BundleContent`
- Pass `null` for `after` instead of artificial date - the change to adjust the query to accept this was already done
- Create/Update tests

# Screenshots

Screenshot 2024-07-16 at 13 37 08

Screenshot 2024-07-16 at 13 37 01


src/pages/RepoPage/BundlesTab/BundleContent/ToggleElement.tsx
```diff
@@ -0,0 +1,53 @@
...
+          data-state={isHidden ? 'closed' : 'open'}
+          onClick={() => {
+            setIsHidden(!isHidden)
+            localStorage.setItem(localStorageKey, String(!isHidden))
+          }}
+        >
+          <span className="transition-transform duration-200">
+            <Icon size="md" variant="solid" name="chevronRight" />
+          </span>
+          {isHidden ? showElement : hideElement}
```
ajay-sentry, line 41: this is a little confusing, isHidden == true then show element?

ajay-sentry: I understand the intent, but maybe there's a way to modify the wording here a bit 🤔

ajay-sentry (approved):
1 comment, not blocking though

PR #3024, opened 2024-07-17
Fix Whitespace in `BotErrorBanner`

Swatinem:
Previously, there was no space between `to InstallCodecov's App`, which is fixed now.
I also removed the `span`, as all other error variants don’t have one either.

image


nicholas-codecov (approved):
Send it

PR #3036, opened 2024-07-18
feat: Add impersonation banner to new header

spalmurray-codecov:
Adds a pink banner that will show when the user is impersonating another user. Also slightly adjusts the layout of the guest header to match what it was before.


ajay-sentry (approved):
nice!

PR #3037, opened 2024-07-18
style: Hide self-hosted header items on small screen sizes

spalmurray-codecov:
Hides the self-hosted header items on small screen sizes as they clutter things up and make weird layouts occur.


ajay-sentry (approved):
✅

PR #3043, opened 2024-07-21
feat: Create logic for OKTA banners

RulaKhaled:
# Description
Conditionally render OKTA banners according to enabled/enforced settings. 

NOTE: keeping this as a draft until we have the API work ready 

# Notable Changes
- Created okta banners logic
- Tests 

# Screenshots
nothing visual changed. 

# Legal Boilerplate


src/shared/GlobalTopBanners/OktaBanners/OktaBanners.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  const { data } = useOktaConfig({
+    provider,
+    username: owner || '',
+    opts: { enabled: !!owner },
+  })
+
+  const oktaConfig = data?.owner?.account?.oktaConfig
+
+  if (!owner || !oktaConfig?.enabled || !data?.owner?.isUserOktaAuthenticated)
```
ajay-sentry, line 25: nit: can potentially move the !owner call to before the hook if we want to avoid calling the hook

Its a slight perf upgrade, not sure how much its worth in practice

RulaKhaled: Can't unconditionally call a hook!

PR #3047, opened 2024-07-23
fix: Update Enterprise Login header

suejung-sentry:
# Description
Update Enterprise Login page to use new white Guest Header instead of old dark header, where GuestHeader is what shows if the user is not logged in.

Other updates:
- Hide "Pricing" and "Start Free Trial" from Enterprise customers GuestHeader
- Hide "Login" button/link when on the "Login page" for both Enterprise and Cloud. Because when clicking the button/link it appears to be a no-op since all it's doing is a redirect to the page you're on already. 

Closes https://github.com/codecov/engineering-team/issues/2049

# Screenshots
## [Enterprise] Login page - Before:
See that it is the dark color
Screenshot 2024-07-24 at 11 03 10 PM

## [Enterprise] Login page - After:
See that it is now the white color with relevant links for Enterprise
Screenshot 2024-07-25 at 3 17 31 PM

## [Enterprise] Guest user any page but Login page - Before:
See it has "Pricing" and "Start Free Trial" link/buttons
Screenshot 2024-07-25 at 3 14 16 PM

## [Enterprise] Guest user any page but Login page - After:
See it omits "Pricing" and "Start Free Trial" link/buttons. "Login" is now a primary button
Screenshot 2024-07-25 at 3 12 45 PM

## [Cloud] Login page - Before:
See that it has a "Login" link despite being on the Login page
Screenshot 2024-07-25 at 2 55 51 PM

## [Cloud] Login page - After:
See that the "Login" link is omitted on this Login page
Screenshot 2024-07-25 at 2 40 57 PM

## [Cloud] Guest user any page but Login page - Before:
See expected header design
Screenshot 2024-07-25 at 2 56 38 PM

## [Cloud] Guest user any page but Login page - After:
See no change to the expected header design
Screenshot 2024-07-25 at 2 41 08 PM

# Legal Boilerplate


ajay-sentry (approved):
lgtm!

src/services/navigation/useNavLinks/useStaticNavLinks.js
```diff
@@ -409,6 +409,11 @@ export function useStaticNavLinks() {
     login: {
       text: 'Login',
       path: () => {
+        // Enterprise login page is at different url than Cloud; see App.tsx
+        if (config.IS_SELF_HOSTED) {
```
ajay-sentry, line 413: Do we still need these conditionals with this change? https://github.com/codecov/gazebo/blob/b07fcc7bae68f7a015dac9804f8649a97b6b5859/src/App.tsx#L72-L79

suejung-sentry: Ah thanks, yeah, good question! So I thought I'd leave that there because-

1. If we're confident the only way a user would end up at the Login Page is via a link/redirect from our app (via `useStaticNavLinks`), we could remove the logic in App.tsx of which component renders when url is `https://{host}/login`. But I thought if say the Enterprise user had `/login` bookmarked, I wouldn't want to show the incorrect page of the Cloud login page to them after this PR.. Or say if another service had a callback url registered somewhere I don't know about, wouldn't wanna break that..

2. I thought it'd be nice if the routing logic that's both in App.tsx and `useStaticNavLinks` would remain consistent. So that would now include this instance of the "Login Page" being at `/` for Enterprise and `/login` for Cloud.

ajay-sentry: Yeah thats a fair point! Having both "shores up" any possibility of a user trying to go to either `/` or `/login` when they should be routed to the other. Down to keep it

PR #3054, opened 2024-07-25
Feat: Fix Failed Test Table Sort, link Hook, and add Spec

ajay-sentry:
# Description

This PR aims to finalize the failed test table stuff, mainly linking the hook, presenting the information correctly, and adding tests for the table.

We were able to test the infinite scroll works as expected, as well as sorting each column

There is an interesting bug that still needs to be fixed though related to scrolling on initial page load when commitsFailed descending order is the column being sorted on

# Screenshots

https://github.com/user-attachments/assets/9eb1a581-aff7-4a26-b139-95a5e4225fc2

# Link to Sample Entry

# Legal Boilerplate


spalmurray-codecov (approved):
lgtm

PR #3057, opened 2024-07-25
feat: Implement useRepoConfigurationStatus

spalmurray-codecov:
Implements useRepoConfigurationStatus, which is the data fetching hook for the new Configuration Manager settings page. No invocation exists yet as it will come in a future PR. Just hook implementation and unit tests here.

### DIFF
--- a/src/pages/RepoPage/SettingsTab/tabs/ConfigurationManager/hooks/useRepoConfigurationStatus/useRepoConfigurationStatus.tsx
+++ b/src/pages/RepoPage/SettingsTab/tabs/ConfigurationManager/hooks/useRepoConfigurationStatus/useRepoConfigurationStatus.tsx
@@ -1,0 +1,142 @@
+import { useQuery } from '@tanstack/react-query'
+import { z } from 'zod'
+
+import {
+  RepoNotFoundErrorSchema,
+  RepoOwnerNotActivatedErrorSchema,
+} from 'services/repo'
+import { TierNames } from 'services/tier'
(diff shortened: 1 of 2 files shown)


src/pages/RepoPage/SettingsTab/tabs/ConfigurationManager/hooks/useRepoConfigurationStatus/useRepoConfigurationStatus.tsx
```diff
@@ -0,0 +1,142 @@
...
+                </p>
+              ),
+            },
+            dev: 'useRepoConfigurationStatus - 403 Owner not activated error',
+          } satisfies NetworkErrorObject)
+        }
+
+        return {
+          plan: data?.owner?.plan,
+          repository: data?.owner?.repository,
```
nicholas-codecov, line 137: `l`: Could `?? null` these bad boys than you only have to handle `null` cases rather than both `null` and `undefined`

nicholas-codecov (approved):
LGTM

src/pages/RepoPage/SettingsTab/tabs/ConfigurationManager/hooks/useRepoConfigurationStatus/useRepoConfigurationStatus.tsx
spalmurray-codecov: great suggestion TY!

PR #3064, opened 2024-07-26
Remove the "Subscription Update Failed" banner when there is no error

rohitvinnakota-codecov:
# Description

This closes https://github.com/codecov/engineering-team/issues/1795.

The error banner will no longer render when the user navigates back from the stripe checkout session. We have a `DelinquentAlert` in the parent component that handles any relevant messaging. 

# Legal Boilerplate


ajay-sentry (approved):
lgtm! Let's also update this comment: https://github.com/codecov/gazebo/blob/8944577e6a1c7f21819bb4199134e0819aa1859b/src/pages/PlanPage/subRoutes/CurrentOrgPlan/InfoMessageStripeCallback/InfoMessageStripeCallback.tsx#L6

PR #3089, opened 2024-07-31
feat: Adjust Sentry user feedback instances

spalmurray-codecov:
Adjusts the Sentry user feedback integration such that the widget only loads on the user onboarding pages (TermsOfService and DefaultOrgSelector). Also Refactors the existing HelpDropdown instantiation of the feedback integration to be contained in sentry.ts, and makes a minor copy change in the HelpDropdown form.

Closes https://github.com/codecov/engineering-team/issues/2173
Closes https://github.com/codecov/engineering-team/issues/1941

### DIFF
--- a/src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
+++ b/src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
@@ -15,16 +15,17 @@ type DropdownItem = {
 }
 
 function HelpDropdown() {
-  const { data: form } = useQuery({
+  const { data: form, isSuccess: isFormSuccess } = useQuery({
     queryKey: ['HelpDropdownForm'],
     queryFn: () => SentryUserFeedback.createForm(),
+    suspense: false,
   })
 
   useLayoutEffect(() => {
-    if (!form) return
+    if (!isFormSuccess) return
     form.appendToDom()
     return form.removeFromDom
-  }, [form])
+  }, [form, isFormSuccess])
 
   const items: DropdownItem[] = [
     {
@@ -36,7 +37,7 @@ function HelpDropdown() {
       children: 'Support center',
     },
     {
-      onClick: form?.open,
+      onClick: isFormSuccess ? form.open : () => {},
       hook: 'open-modal',
       children: 'Share feedback',
     },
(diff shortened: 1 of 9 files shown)


src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
```diff
@@ -25,13 +27,17 @@ type Item = {
...
-  document.body.style.overflow = ''
-  document.querySelector('#help-dropdown-widget')?.remove()
-}
-
 function HelpDropdown() {
-  useEffect(() => removeSentryForm, [])
+  const { data: form } = useQuery({
+    queryKey: ['HelpDropdownForm'],
+    queryFn: () => SentryUserFeedback.createForm(),
+  })
```
spalmurray-codecov, line 34: This was a clean way to handle the async createForm call. Alternative is some cursed combination of useState/Effect/Memo.

(also wanted to try this out after Nick explained to me that TS query could be used for any async state management, not just network requests 😜)

src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
```diff
@@ -25,13 +27,17 @@ type Item = {
...
+
+  useLayoutEffect(() => {
+    if (!form) return
```
nicholas-codecov, line 37: You could also use something like `isSuccess` here if you wanted 👀

src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
nicholas-codecov: Hrrm, suspense would currently be enabled here, may want to play around with toggling this incase it brings up any odd issues.

src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
```diff
@@ -44,23 +50,7 @@ function HelpDropdown() {
...
-            isEmailRequired: true,
-            autoInject: false,
-            id: 'help-dropdown-widget',
-            onFormClose: removeSentryForm,
-          })
-          const form = await sentryFeedback.createForm()
-          form.appendToDom()
-          form.open()
-        },
+        onClick: form?.open,
```
nicholas-codecov, line 53: Could also use the `isSuccess` here as well, and if not just an empty function?

nicholas-codecov (approved):
Fancy work sir, nice use of TSQuery, left some comments to peak at, nothing major

src/layouts/Header/components/HelpDropdown/HelpDropdown.tsx
spalmurray-codecov: good call

PR #3091, opened 2024-08-01
feat: Introduce VirtualFileRenderer to handle rendering of large files

nicholas-codecov:
# Description

This PR introduces a new code renderer, to replace our current implementation. The secret sauce behind this code renderer is the use of virtual lists to do the handle the rendering of highlighted code to the screen. To get the correct syntax highlighting we're required to wrap each token of the provided code string in it's own `<span>` tag, you can imagine this growing to a significantly large number depending on the size of the file. This is where the heart of the problem lays, React is required to paint all of these spans to the page which is an expensive computation once you reach a large enough file. The power of the new code renderer is that we're virtualizing these rows so we're only asking React to render all visible lines and around 100 overscan for less janky scrolling. This significantly reduces the amount of painting to the screen that React is required to do and in turn making it so we can render files faster, and handle files with sizes in the megabytes.

This issue was originally raised from this [tweet](https://x.com/andhaveaniceday/status/1805471407654289810?s=12&t=mwzZkOg9ugD5RwD-MXkneg), these changes should resolve this issue and allow the user to view this file in Codecov.

**Huge shoutout** to @JonasBa for helping make this happen 🎉 

### Some performance metrics:

- File: [services/notification/notifiers/tests/unit/test_comment.py](https://github.com/codecov/worker/blob/main/services/notification/notifiers/tests/unit/test_comment.py)
- 5612 lines (5435 loc)
- Codecov API request size: `326 kB`

**Old renderer:**

Screenshot 2024-08-02 at 11 48 32

- 4346ms of actual work (non-idle time)

**New renderer:**

Screenshot 2024-08-02 at 11 48 49

- 481ms of actual work (non-idle time)

# Notable Changes

- Use virtual lists to render all of the lines
- Refactor `prismLanguageMapper` to TS
- Create/Update tests

# Screenshots


src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,363 @@
...
+          overflowX: 'overlay',
+        }}
+      >
+        <div ref={widthDivRef} className="w-full">
+          {/* @ts-ignore */}
+          <Highlight
+            {...defaultProps}
+            // @ts-ignore
+            code={code}
+            language={prismLanguageMapper('js')}
```
nicholas-codecov, line 346: Need to grab this from a prop

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,363 @@
...
+                }}
+                className="absolute left-0 top-0 pl-[92px]"
+              >
+                <ColorBar
+                  lineNumber={item.index}
+                  locationHash={location.hash}
+                  coverage={coverage[item.index]}
+                />
+                <div className="h-[18px] w-full" style={{ ...lineStyle }}>
+                  {tokens[item.index]?.map((token: any, key: any) => (
```
nicholas-codecov, line 185: Need to fix these types

src/shared/utils/prismLanguageMapper.spec.ts
```diff
@@ -0,0 +1,209 @@
+import { prismLanguageMapper } from './prismLanguageMapper'
+
+describe('prismLanguageMapper', () => {
+  describe('when called with a file with a valid extension', () => {
+    describe('html', () => {
```
JonasBa, line 5: these seems very similar, I would it.each all of the cases here

src/shared/utils/prismLanguageMapper.ts
```diff
@@ -0,0 +1,51 @@
...
+  ['tsx', 'tsx'],
+  ['wasm', 'wasm'],
+  ['yaml', 'yaml'],
+])
+
+export function prismLanguageMapper(fileName: string): Language {
+  const fileExtension = fileName.split('.').pop() ?? ''
+
+  // we have to return something or else TS will complain
+  const lang = prismSupportedLanguages.get(fileExtension) ?? 'markup'
```
JonasBa, line 49:
```suggestion
	return prismSupportedLanguages.get(fileExtension) ?? 'markup'
```

src/shared/utils/prismLanguageMapper.ts
```diff
@@ -0,0 +1,51 @@
...
+  return lang as Language
```
JonasBa, line 50:
```suggestion
```

src/shared/utils/prismLanguageMapper.ts
```diff
@@ -0,0 +1,51 @@
+/*
+ * Maps file extension names to prism supported languages.
+ * The full list can be found in the prism-react-renderer package, just cmd +
+ * click on the import string.
+ */
+
+import { type Language } from 'prism-react-renderer'
+
+const prismSupportedLanguages = new Map([
```
JonasBa, line 9:
```suggestion
const prismSupportedLanguages = new Map<string, Language>([
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+window.requestAnimationFrame = (cb) => {
+  cb(1)
+  return 1
+}
+window.cancelAnimationFrame = () => {}
+
+const scrollToMock = jest.fn()
+window.scrollTo = scrollToMock
+
+class ResizeObserver {
```
JonasBa, line 18:
```suggestion
class ResizeObserverMock {
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+    this.callback([{ target: { getAttribute: () => ({ scrollWidth: 100 }) } }])
+  }
+  unobserve() {
+    // do nothing
+  }
+  disconnect() {
+    // do nothing
+  }
+}
+global.window.ResizeObserver = ResizeObserver
```
JonasBa, line 35:
```suggestion
global.window.ResizeObserver = ResizeObserverMock
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+      <Route path="/">{children}</Route>
+      <Route
+        path="*"
+        render={({ location }) => {
+          testLocation = location
+          return null
+        }}
+      />
+    </MemoryRouter>
+  )
```
JonasBa, line 69: :wat:?

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+
+      const codeRenderer = screen.getByTestId('virtual-file-renderer')
+      await waitFor(() =>
+        expect(codeRenderer).toHaveStyle('pointer-events: none')
+      )
+      await waitFor(() =>
+        expect(codeRenderer).toHaveStyle('pointer-events: auto')
+      )
```
JonasBa, line 255: This is the same test as above and very low value. I would just add 
```
await waitFor(() =>
   expect(codeRenderer).toHaveStyle('pointer-events: auto')
)

File: src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx [hunk 11]
```diff
@@ -0,0 +1,361 @@
...
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+
+      // eslint-disable-next-line testing-library/no-container
+      container.remove()
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+      await fireEvent.scroll(window, { target: { scrollX: 100 } })
+
+      await waitFor(() => expect(cancelAnimationFrameSpy).toHaveBeenCalled())
+    })
+  })
```
JonasBa, line 294: Not really sure what you are trying to test here? Is there something more tangible that we can assert besides that cancelAnimationFrame is called?

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+          }
+        )
+
+        const line = screen.getByText(0)
+        await user.click(line)
+
+        await waitFor(() => expect(testLocation.hash).toBe('#L0'))
+      })
+
+      it('highlights the line', async () => {
```
JonasBa, line 317:
```suggestion
      it('highlights the line on click', async () => {
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+        const line = screen.getByText(0)
+        await user.click(line)
+
+        const bar = await screen.findByTestId('covered-bar')
+        expect(bar).toBeInTheDocument()
+        expect(bar).toHaveClass('bg-ds-coverage-covered')
+      })
+
+      describe('user clicks the line again', () => {
+        it('clears the line', async () => {
```
JonasBa, line 339:
```suggestion
      it('removes highlighting when clicking on highlighted line', () => {
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const history = useHistory()
+  const location = useLocation()
+
+  const initialRender = useRef(true)
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  // virtualize code lines
+  const virtualizer = useWindowVirtualizer({
+    count: tokens.length,
+    estimateSize: () => 18,
```
JonasBa, line 45: I would add a comment or pull this into a const
```suggestion
    estimateSize: () => 18, // line row height
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+    overscan: 75,
+    scrollMargin: codeDisplayOverlayRef.current?.offsetTop ?? 0,
+  })
+
+  const div = codeDisplayOverlayRef.current
```
JonasBa, line 50: This will read better down as there is some intent/naming to it
```suggestion
  const codeDisplayOverlayElement = codeDisplayOverlayRef.current
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+  if (div) {
+    // set the parent div height to the total size of the virtualizer
+    div.style.height = `${virtualizer.getTotalSize()}px`
+    div.style.position = 'relative'
+  }
```
JonasBa, line 55: Does this have to get called on each render? Could is just be done once when the tokens are rendered? Idk how this works, but I assume that as many lines are rendered, this method (depending on the implementation) might become slower

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+
+  // TODO - finalize this logic
+  // scroll to line on the initial render
```
JonasBa, line 58:
```suggestion
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+  useEffect(() => {
```
JonasBa, line 59: Can this be a useLayoutEffect?

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+      <div
+        // @ts-expect-error - TODO - Update inert when inert is available in React 19
+        inert=""
+        className="pointer-events-none size-full"
+      >
+        {virtualizer.getVirtualItems().map((item) => {
+          // get any specific things from code highlighting library for this given line
+          const { style: lineStyle } = getLineProps({
+            // casting this cause it is guaranteed to be present in the array
+            line: tokens[item.index] as Token[],
```
JonasBa, line 116: Would this do the same thing?
```suggestion
            line: tokens[item.index]!,
```

src/ui/VirtualFileRenderer/ColorBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+        data-testid="uncovered-bar"
+        className="pointer-events-none absolute left-[-72px] h-full w-[calc(100%+72px)] bg-ds-coverage-uncovered opacity-25"
+      />
+    )
+  } else if (coverage === 'P') {
+    return (
+      <div
+        data-testid="partial-bar"
+        className="pointer-events-none absolute left-[-72px] h-full w-[calc(100%+72px)] bg-ds-coverage-partial opacity-25"
+      />
```
JonasBa, line 32: Nit, but you can just apply the class and testid through a single ternary and avoid the 3 else ifs
```suggestion
      <div
        data-testid={coverage === "P" ? partial-bar : ...}
        className={`pointer-events-none absolute left-[-72px] h-full w-[calc(100%+72px)] opacity-25 ${coverage === "P" ? bg-ds-coverage-partial : ...}`}
      />
```

src/ui/VirtualFileRenderer/ColorBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+    )
+  } else if (coverage === 'H') {
+    return (
+      <div
+        data-testid="covered-bar"
+        className="pointer-events-none absolute left-[-72px] h-full w-[calc(100%+72px)] bg-ds-coverage-covered opacity-25"
+      />
+    )
+  } else {
+    return null
```
JonasBa, line 42: This is hiding a future bug. You should never do this when dealing with an exhaustive type and should always throw an error.

src/shared/utils/prismLanguageMapper.ts
```diff
@@ -0,0 +1,51 @@
+/*
+ * Maps file extension names to prism supported languages.
+ * The full list can be found in the prism-react-renderer package, just cmd +
+ * click on the import string.
```
JonasBa, line 4: Not everyone has the same editor or key bindings and it is better to add a source code link to the line in gh

src/ui/VirtualFileRenderer/ColorBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  coverage?: 'H' | 'M' | 'P'
+}
+
+// exporting for testing purposes
+export const ColorBar = ({
+  coverage,
+  locationHash,
+  lineNumber,
+}: CoverageBarProps) => {
+  if (locationHash && locationHash === `#L${lineNumber}`) {
```
JonasBa, line 13: You can just compare to the string literal
```suggestion
  if (locationHash === `#L${lineNumber}`) {
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+    div.style.height = `${virtualizer.getTotalSize()}px`
+    div.style.position = 'relative'
+  }
+
+  // TODO - finalize this logic
+  // scroll to line on the initial render
+  useEffect(() => {
+    if (initialRender.current) {
+      initialRender.current = false
+      const index = parseInt(location.hash.slice(2), 10)
```
JonasBa, line 62: You need to handle negative or bad line links
```suggestion
      const index = parseInt(location.hash.slice(2), 10)
      if(isNaN(index) || index < 0){
        Sentry.captureMessage or throw error
      }
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
+import { useWindowVirtualizer } from '@tanstack/react-virtual'
```
JonasBa, line 1: I've found that leaving a comment on how the virtualization is done (particularly the textarea trick and how it enables native cmd+f) is something worth writing about here. It will help future readers understand how it work and learn something without having to look at the details of the implementation

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+                'bg-ds-coverage-partial after:absolute after:inset-y-0 after:right-0 after:border-r-2 after:border-dotted after:border-ds-primary-yellow'
+            )}
+            onClick={() => {
+              if (location.hash === `#L${item.index}`) {
+                location.hash = ''
+              } else {
+                location.hash = `#L${item.index}`
+              }
+              history.push(location)
+            }}
```
JonasBa, line 99: Does this require pushing here? 
```suggestion
            onClick={() => {
              location.hash = location.hash === `#L${item.index}` ? '' : `#L${item.index}`
            }}
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+          })
+
+          return (
+            <div
+              ref={virtualizer.measureElement}
+              key={item.index}
+              data-index={item.index}
+              style={{
+                width: wrapperRef?.current
+                  ? wrapperRef.current.clientWidth
```
JonasBa, line 127: This is going to cause layout thrashing and potential perf issues as it will read the clientWidth of the wrapper for each rendered row.

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
nicholas-codecov: So this is to wrap the tests, because we're looking to test for when the user clicks on the line number it updates the URL that requires this wrapper so that we have browser history present for the test.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+  coverage,
+  codeDisplayOverlayRef,
+}: CodeBodyProps) => {
+  const history = useHistory()
+  const location = useLocation()
+
+  const initialRender = useRef(true)
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  // virtualize code lines
```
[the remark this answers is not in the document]
nicholas-codecov: not resolving for the shame

src/ui/VirtualFileRenderer/ColorBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div
+        data-testid="highlighted-bar"
+        className="pointer-events-none absolute left-[-72px] h-full w-[calc(100%+72px)] bg-ds-blue-medium opacity-25"
+      />
+    )
+  } else if (coverage === 'M') {
+    return (
+      <div
+        data-testid="uncovered-bar"
+        className="pointer-events-none absolute left-[-72px] h-full w-[calc(100%+72px)] bg-ds-coverage-uncovered opacity-25"
```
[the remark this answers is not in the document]
nicholas-codecov: There is a lint rule present in our codebase that enforces the order of the classnames, in the order that is suggested by Tailwind

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: Should also probably handle line numbers that are too large.

src/ui/VirtualFileRenderer/ColorBar.tsx
nicholas-codecov: So the type can be either highlighted, one of the three coverage values, or undefined, and when it's undefined we don't want this to be rendered, so this case is what we want.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: Would setting this to a const under the ref definition work? or would i need to set it in something like a useEffect and set a piece of state?

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: Thoughts on handling only these cases: `!isNaN(index) && index > 0 && index < tokens.length` and ignoring the rest? By default the value is `NaN` so it would constantly throw an error.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: This seems to have worked fine, so now it's only getting rendered in the parent component, should stop any issues.

src/ui/VirtualFileRenderer/ColorBar.tsx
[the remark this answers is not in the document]
JonasBa: That is just sad

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: You mean default value is blank, but gets parsed as NaN? I think a blank value is valid, but something like #Lwat that doesnt parse to a lineno should be handled

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: I think the URL updates if you just assign to location.hash, but it seems like you push because you need to trigger react to update and the UI to relfect the change?

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: The way I would usually handle this is by attaching a resize observer and storing the value. That way you get the value once when it renders and store it in state (or assign it directly via css var), then keep it in sync as it resizes

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
JonasBa: Hmm, and is there no other way to access or listen to location? It seems weird to have to render and assign to a local variable to escape this, surely there is a better way?

src/ui/VirtualFileRenderer/ColorBar.tsx
JonasBa: You should add a line at the start to cover undefined then and a captureMessage to detect any other values. Right now if someone adds a new coverage type, you will need to keep track of all the areas where this is defined and update it, which is easy to miss and forget to update. Instead you should do something like

```
function exhaustive(_x: never): void {}

function coverage(type: "A"|"B"|undefined)
if(type === undefined) return // valid

switch(type){
  case "A":
    return value
  case "B":
    return other value
   default: {
     Sentry.captureMessage(`Invalid coverage value ${type}`);
     exhaustive(type)
   }
}
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: This size is static, its the max width of the textarea, so it won't change even on resize.

src/ui/VirtualFileRenderer/ColorBar.tsx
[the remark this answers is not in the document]
nicholas-codecov: I mean ... it just be that way sometimes 🤷‍♂️

src/ui/VirtualFileRenderer/ColorBar.tsx
nicholas-codecov: What is this `exhaustive` function for?

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: So this value is actually the width of the text area so it's fairly static even with resizing the window, so I think we can probably skip the resize observer in this instance and just leave it be, lmk what you think.

src/ui/VirtualFileRenderer/ColorBar.tsx
nicholas-codecov: Ahh I think it's just a placeholder for this example ... i think ... I can just return `null` as the default return and capture the message beforehand.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: So this is what I came up with instead, handle the "valid" cases and ignore the rest

```tsx
if (isNumber(index) && index > 0 && index < tokens.length) {
  requestAnimationFrame(() => {
    virtualizer.scrollToIndex(index, { align: 'start' })
  })
}
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: Will do 🫡

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: It can be, however ... it doesn't resolve the issue of the scroll to function not being called before rendering, which I haven't found any "solution" provided via docs or issues in their repo.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: So this `getTotalSize` seems to just be the number of rows * estimate size so that it can set the correct height to expand the window so that it feels like a full window scroll, I don't think there's any major perf issues here, this is fairly similar as to how they have it in their example in the documentation, however they have the style being directly applied to the component, however because of our code highlighting setup that's a bit trickier to do, I could possibly refactor things but that would be a fairly big refactor at this point I believe.

src/ui/VirtualFileRenderer/ColorBar.tsx
JonasBa: Exhaustive check is to ensure that you are handling each case. If you for example remove the B case, then the type you are trying to pass to exhaustive will be "B", which will fail. If you handle B, then the type is narrowed to `never`, because it cannot happen

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: I think that is because the instance of location is the one exposed by the router, and not something the one exposed on the window.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: I would add the resize observer to be safe. It is safer, more correct and might save you some time from finding misalignment issues in the future.

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+          coverage={coverageData}
+          fileName="tsx"
+        />,
+        {
+          wrapper: wrapper(),
+        }
+      )
+
+      const coveredLine = screen.getByText(0)
+      expect(coveredLine).toBeInTheDocument()
```
JonasBa, line 124: Unless this line performs an assertion, then I think you can remove it as the line after will fail if its not in the document

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+        wrapper: wrapper(),
+      }
+    )
+
+    const lineNumbers = screen.getAllByText(/\d+/)
+    expect(lineNumbers).toHaveLength(8)
+  })
+
+  describe('covered lines', () => {
+    it('uses the correct color', () => {
```
JonasBa, line 111:
```suggestion
    it('applies coverage background', () => {
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+  )
+
+describe('VirtualFileRenderer', () => {
+  function setup() {
+    const user = userEvent.setup()
+
+    return { user }
+  }
+
+  it('renders the code', () => {
```
JonasBa, line 78: Do you want to ensure that both textarea and tokens are rendered here?

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+      )
+
+      const uncoveredLine = screen.getByText(1)
+      expect(uncoveredLine).toBeInTheDocument()
+      expect(uncoveredLine).toHaveClass('bg-ds-coverage-uncovered')
+    })
+  })
+
+  describe('partial lines', () => {
+    it('uses the correct color', () => {
```
JonasBa, line 149:
```suggestion
    it('applies partial coverage background color', () => {
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+          coverage={coverageData}
+          fileName="tsx"
+        />,
+        {
+          wrapper: wrapper(),
+        }
+      )
+
+      const partialLine = screen.getByText(2)
+      expect(partialLine).toBeInTheDocument()
```
JonasBa, line 162: same comment about removing this

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,361 @@
...
+          code={code}
+          coverage={coverageData}
+          fileName="tsx"
+        />,
+        {
+          wrapper: wrapper(),
+        }
+      )
+
+      const partialLine = screen.getByText(2)
```
JonasBa, line 161: Is there a better way you can query this that is more future proof? Right now if you were to change the test setup, this would break.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: I think the solution here is to do something like

if loading
  render <LoadingPlaceholder/>
else
  render <VirtualizedList/>
  
That way you should be able to entirely avoid raf and just call it inside a useLayoutEffect as the first render will already contain all the data

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
JonasBa: (in any case, its fine like this)

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,323 @@
...
+import { useEffect, useLayoutEffect, useRef } from 'react'
+import { useHistory, useLocation } from 'react-router-dom'
+
+import { requestAnimationTimeout } from 'shared/utils/animationFrameUtils'
+import { cn } from 'shared/utils/cn'
+import { prismLanguageMapper } from 'shared/utils/prismLanguageMapper'
+
+import { ColorBar } from './ColorBar'
+
+// copied from prism-react-renderer since they don't export it
```
JonasBa, line 14: Worth checking if you can infer this from anything they expose since it relates to a 3rd party dependency

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: sounds good 👍

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
nicholas-codecov: Updated to check both the text area and virtual code renderer

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: So i was a bit off on how this method actually works, however diving deeper it seems that most of the most complex calculations of this function are memoized after their first run so the cost of running this function goes down and I believe fairly performant.

- `getTotalSize`: https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts#L1001-L1019
- `getMeasurements`: https://github.com/TanStack/virtual/blob/main/packages/virtual-core/src/index.ts#L585

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: So I checked this out, and the version we're using there's really only this `const markup = Prism.languages['markup']` however the types don't actually align with the `Language` type and break things so there isn't really anything that we can use, at least with the version we're currently using.

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
nicholas-codecov: alrighty done

src/ui/VirtualFileRenderer/ColorBar.tsx
nicholas-codecov: Ahhh okay noice, done

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
nicholas-codecov: So this is the way outlined in the "Testing" section for React-Router V5: https://v5.reactrouter.com/web/guides/testing/checking-location-in-tests

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
nicholas-codecov: Hrrm off the top of my head, I think the closet thing would be something along with these lines:

```
      const coveredLine = virtualOverlay.querySelector(
        '.bg-ds-coverage-covered'
      )
      expect(coveredLine).toHaveClass('bg-ds-coverage-covered')
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
JonasBa: Cant argue with official docs :)

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,339 @@
...
+  const location = useLocation()
+  const [wrapperWidth, setWrapperWidth] = useState<number | '100%'>('100%')
+
+  const initialRender = useRef(true)
+  const wrapperRef = useRef<HTMLDivElement>(null)
+
+  const virtualizer = useWindowVirtualizer({
+    count: tokens.length,
+    estimateSize: () => 18, // line row height
+    overscan: 75,
```
nicholas-codecov, line 47: Lower this a bit to like 10, 20, 30 or so

src/shared/utils/prismLanguageMapper.ts
```diff
@@ -0,0 +1,50 @@
...
+  ['wasm', 'wasm'],
+  ['yaml', 'yaml'],
+])
+
+export function prismLanguageMapper(fileName: string): Language {
+  const fileExtension = fileName.split('.').pop() ?? ''
+
+  // we have to return something or else TS will complain
+  return prismSupportedLanguages.get(fileExtension) ?? 'markup'
+}
```
JonasBa, line 50: It will rightfully complain though, as there is no guarantee that .get will not return undefined. If you want, something like this might make this a bit more readable and explain the intent here

```suggestion
const DEFAULT_LANGUAGE_TYPE: Language = 'markdown'
export function prismLanguageMapper(fileName: string): Language {
  const fileExtension = fileName.split('.').pop() ?? ''

  // we have to return something or else TS will complain
  return prismSupportedLanguages.get(fileExtension) ?? DEFAULT_LANGUAGE_TYPE
}
```

JonasBa: Something that just occurred to me here is that we dont have any visibility into how often the default case will be hit in prod so it would be useful to captureMessage and monitor this to see if there is missing support for some popular languages folks use on codecov

```ts
const supportedLanguage = prismSupportedLanguages.get(fileExtension)
if(supportedLanguage) return supportedLanguage

Sentry.captureMessage('Unsupported language type for filename ${fileName}') // make sure to add a fingerprint here btw
return DEFAULT_LANGUAGE_TYPE
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.tsx
```diff
@@ -0,0 +1,365 @@
...
+ * highlight/select code for copy pasting, and also to enable users to cmd/ctrl
+ * + f to search for text in the code. We also use the width of the textarea to
+ * set the width of the code display element so the code display element can
+ * scroll horizontally in sync with the text area.
+ *
+ * The VirtualFileRenderer contains logic as well that enables pointer-events
+ * to be disabled when ever the user scrolls and then after an allotted amount
+ * of time they're turned back on. This is to reduce the amount of work the
+ * browser needs to do during the repaint, as it does not need to check these
+ * events.
```
JonasBa, line 20: This feel too detailed at this point, I think you just want to explain the main idea which is to overlay the tokens above textarea and leverage the platform for native features.

```suggestion
```

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
```diff
@@ -0,0 +1,418 @@
...
+    )
+
+    const textArea = screen.getByTestId('virtual-file-renderer')
+    expect(textArea).toBeInTheDocument()
+
+    const codeBlock = within(textArea).getByText(/<Breadcrumb/)
+    expect(codeBlock).toBeInTheDocument()
+  })
+
+  it('renders code in virtualized list', () => {
```
JonasBa, line 116: I would test that this renders for some unknown language too. You could supply it a bad value and check that it renders (iirc as markup)

src/shared/utils/prismLanguageMapper.ts
```diff
@@ -0,0 +1,57 @@
...
+  ['ts', 'typescript'],
+  ['tsx', 'tsx'],
+  ['wasm', 'wasm'],
+  ['yaml', 'yaml'],
+])
+
+const DEFAULT_LANGUAGE_TYPE: Language = 'markup'
+
+export function prismLanguageMapper(fileName: string): Language {
+  const fileExtension = fileName.split('.').pop() ?? ''
```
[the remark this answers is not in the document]
nicholas-codecov: Thoughts on if there is no `.` then we should just render it as markup?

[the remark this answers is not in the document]
JonasBa: There could be multiple ways, one is to just not render any tokens since we cant tell what file type it is, the other could be to set a value like markup, but then also allow users to select a different language (they'll know best what the contents could be)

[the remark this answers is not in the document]
nicholas-codecov: Hrrm alrighty, I wonder how big of problem this may be. So just some background information, Codecov will only display files in the UI that we have coverage for, so if you have a file like `.prettierrc` unless you somehow write tests for that, it won't show up in the UI. With that in mind I think we should rollout what we have now, and if things blow up then we can tackle them there. Adding in a selector for selecting languages seems ... a little complex to add into this PR.

src/ui/VirtualFileRenderer/VirtualFileRenderer.spec.tsx
nicholas-codecov: Will do

src/shared/utils/prismLanguageMapper.ts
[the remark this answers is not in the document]
JonasBa: No, I def wouldnt add it to this PR, but it might make sense to progressively enhance this later on. Defaulting to markup seems like a good default for now

spalmurray-codecov (approved):

PR #3092, opened 2024-08-01
unfeat: Remove branch breadcrumb on repo page

spalmurray-codecov:
Removes the branch breadcrumb on RepoPage. Part of the header overhaul work, but waited to do this until the old header was removed to make things easier.

Closes https://github.com/codecov/engineering-team/issues/2066

### DIFF
--- a/src/pages/CommitDetailPage/CommitDetailPage.tsx
+++ b/src/pages/CommitDetailPage/CommitDetailPage.tsx
@@ -82,6 +82,7 @@ const CommitDetailPage: React.FC = () => {
         text: shortSHA,
       },
     ])
+    return () => setBreadcrumbs([])
   }, [
     setBreadcrumbs,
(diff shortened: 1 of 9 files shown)


src/pages/CommitDetailPage/CommitDetailPage.tsx
```diff
@@ -82,6 +82,7 @@ const CommitDetailPage: React.FC = () => {
         text: shortSHA,
       },
     ])
+    return () => setBreadcrumbs([])
```
spalmurray-codecov, line 85: previously we were calling `setBreadcrumbs` with the branch when leaving these pages, but since we no longer are, we need to do this cleanup step on unmount.

calvinyau (approved):
Looking spiffy!

calvin-codecov (approved):

PR #3097, opened 2024-08-05
dev: remove double yarn install on dockerfile

ajay-sentry:
# Description

saw that yarn was being installed at v1.22.19, and then later at 4.3.1, this PR aims to get rid of the original npm builder that was just getting overwritten with the second yarn install. I think previously we were using the separate npm builder because downloading and using the packages was taking 2+ minutes where now it takes ~30 seconds

This leads to a timing improvement overall of ~1.5 minutes pre/post (25% faster)

Pre: https://github.com/codecov/gazebo/actions/runs/10255316136/job/28372007061 (6:14)
Post: https://github.com/codecov/gazebo/actions/runs/10275127882/job/28433206745 (4:38)

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,8 +1,8 @@
 # syntax = docker/dockerfile:1.3
 FROM node:20-alpine3.17 as npm
 RUN apk -U add git 
-RUN corepack enable && corepack prepare yarn@stable --activate
 RUN mkdir /home/workspace
+RUN corepack enable
 COPY package.json /home/workspace
 WORKDIR /home/workspace
 RUN yarn install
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -50,11 +42,10 @@ ENV REACT_APP_ENV=$REACT_APP_ENV_ARG
 ENV REACT_APP_CODECOV_VERSION=$REACT_APP_CODECOV_VERSION
 ENV GENERATE_SOURCEMAP=false
 RUN mkdir /home/workspace
-COPY --from=npm /home/workspace/node_modules /home/workspace/node_modules
 WORKDIR /home/workspace
 RUN apk -U add git
-RUN apk add --no-cache yarn
 COPY . /home/workspace
+RUN corepack enable
```
suejung-sentry, line 48: Super cool!

Does corepack have access to the package.json even if we're no longer copying it over with `COPY package.json /home/workspace` ?

suejung-sentry (approved):

docker/Dockerfile
suejung-sentry: Oh nvm, it's getting copied in above (?). Wonder why it was duplicated..

ajay-sentry: not too sure tbh 🤔

PR #3103, opened 2024-08-07
feat: Clarify onboarding org choices for business

calvin-codecov:
# Description

Closes https://github.com/codecov/engineering-team/issues/1905

# Code Example

# Notable Changes
For accounts that have selected "Business use" on the Terms page, we are:
- removing their personal org from the list
- showing them "No organizations found" if they don't have any other organizations either

I had to make some modifications to the Select component but I tried to make it is generic/resuable as possible. Essentially for an `item` that has an `isDisabled` field, it will lose its hover highlighting, cursor change, and will not be selectable. I decided on this solution as the designs call for "No organizations found" to be in the list options, above the "Add Github App" link. I explored using the already built "No results found" (based on `onSearch` function existence) but it didn't fit the purpose.

# Screenshots
Screenshot 2024-08-07 at 11 25 22 AM

# Link to Sample Entry

# Legal Boilerplate


src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
```diff
@@ -134,6 +144,14 @@ function DefaultOrgSelector() {
   if (userIsLoading) return null
   if (!userIsLoading && !currentUser) return <Redirect to="/login" />
 
+  console.log('current user:', currentUser?.user?.username)
```
spalmurray-codecov, line 147: remove `console.log`s

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
```diff
@@ -134,6 +144,14 @@ function DefaultOrgSelector() {
...
+  const filteredOrganizations =
+    isBusinessIntent &&
+    myOrganizations[0]?.org?.username === currentUser?.user?.username
+      ? myOrganizations.slice(1)
+      : myOrganizations
+  console.log('myOrgs:', filteredOrganizations)
```
spalmurray-codecov, line 153: here too

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
spalmurray-codecov, line 1: If you have the willpower to do so, maybe convert these to TS!

src/ui/Select/Select.jsx
```diff
@@ -138,6 +138,17 @@ const Select = forwardRef(
     }
 
     function _renderItem(item, index) {
+      if (item?.isDisabled) {
```
spalmurray-codecov, line 141: Think we should unit test this in `Select.spec.jsx` as well

spalmurray-codecov (approved):

src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
calvin-codecov: 🤦 i went back in to delete these and got distracted with fixing something else. apologies

PR #3108, opened 2024-08-08
feat: Add 429 Network Error page and sentry metric

ajay-sentry:
# Description

As part of the rate limit updates for REST and GQL, we want to add an error page for users who are currently rate limited. It's essentially the same page as before, but featuring a 429 error which is used by DRF (and soon to be our GQL endpoint)

Also, I saw that we had duplicate SVGs with just different error names, which is an anti-pattern. IMO, an svg/asset shouldn't be duplicated for different "aspects" of its usage, but rather named for what it is and then used for the various situations.

# Screenshots

Screenshot 2024-08-08 at 2 57 58 PM

# Link to Sample Entry

# Legal Boilerplate


src/services/repo/useRepo.tsx
```diff
@@ -121,14 +121,12 @@ export function useRepo({ provider, owner, repo, opts = {} }: UseRepoArgs) {
           }
         }
 
-        return (
-          {
```
ajay-sentry, line 125: this was a random fix, right side of expression would never be hit since we're always returning an object

src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
```diff
@@ -209,8 +217,8 @@ class NetworkErrorBoundary extends Component {
           src={illustration}
         />
         <h1 className="mt-6 text-2xl">{title}</h1>
-        {description && <p className="mt-6">{description(data)}</p>}
-        <NetworkErrorMessage />
+        {description && <p className="mt-2">{description(data)}</p>}
+        {showDocs && <NetworkErrorMessage />}
```
nicholas-codecov, line 221: Can you swap this for a ternary 👀

nicholas-codecov: We do now have a [notion doc](https://www.notion.so/sentry/Use-ternaries-rather-than-in-JSX-5ccc5225e4fb468a8165071e891afaac) explaining the reasoning behind favouring the ternaries over double ampersands.

nicholas-codecov (approved):
Just one small thing to address

src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
ajay-sentry: oh yeah no prob, lemme swap the others too

PR #3114, opened 2024-08-09
ref: Convert BaseLayout to TS

spalmurray-codecov:
Converts BaseLayout.jsx to BaseLayout.tsx. Also shows normal header when impersonating and not isFullExperience.


src/layouts/BaseLayout/BaseLayout.tsx
```diff
@@ -70,7 +82,7 @@ function BaseLayout({ children }) {
     <>
       <Suspense fallback={<FullPageLoader />}>
         <RepoBreadcrumbProvider>
-          {isFullExperience ? (
+          {isFullExperience || isImpersonating ? (
```
spalmurray-codecov, line 85: this is to prevent the weird case when impersonating someone who has not completed onboarding

suejung-sentry (approved):
nice!

PR #3116, opened 2024-08-12
chore: Update yarn version and align pinned nvm node

spalmurray-codecov:
Updates yarn to 4.4.0 and brings pinned node version in .nvmrc in line with pinned version in volta.


package.json
```diff
@@ -182,11 +182,11 @@
...
   "volta": {
     "node": "20.15.1",
-    "yarn": "4.3.1"
+    "yarn": "4.4.0"
   },
   "overrides": {
     "react-refresh": "^0.14.0"
   },
-  "packageManager": "yarn@4.3.1",
+  "packageManager": "yarn@4.4.0",
```
suejung-sentry, line 190: @ajay-sentry , do you know if we need to pin the `yarn` version used in our deploy workflows to match that we use for development specified here? 
For example, the deploy version `yarn@stable` at here:
https://github.com/codecov/devops-workflows/blob/a37bdc7a81fbf0363ae842998a86acb1301fb8a9/.github/workflows/gazebo-release.yml#L93

suejung-sentry (approved):
approved! have a quick non-blocking question to ajay in above!

package.json
ajay-sentry: Technically we don't need to, though it would give us 1 less "degree of freedom" to analyze in the event of discrepancies (probability is fairly low here)

my take originally with doing @stable there was to reduce the need of having to update those devops workflows if we can help it since it's a bit of a cumbersome process.

I'm cool with pinning or not pinning, can make a case for either :)

suejung-sentry: Ah yeah makes sense! I would personally lean towards pinned in prod or at least consistent across dev and prod (e.g., both on stable). But as you mentioned, it's probably low risk and adds friction, so we can leave it be

PR #3119, opened 2024-08-13
feat: Shift error boundary to include header

ajay-sentry:
# Description

This PR aims to fix the "weirdness" happening when the GQL endpoint would return a 429 error back to Gazebo. The header component wasn't wrapped in the errorBoundary component which was leading to a blank page being shown instead of the rate limit page

The error message coming back from GQL is currently generic vs. the one from DRF that shows the actual throttle time, but we can always extend this later on if we want to give the user that level of feedback as well.

Closes https://github.com/codecov/engineering-team/issues/2292

# Screenshots

https://github.com/user-attachments/assets/4fc2776b-4346-47c1-b2a3-f5973f05f271

# Link to Sample Entry

# Legal Boilerplate


spalmurray-codecov (approved):
lgtm. Must have missed this in the header changeover

src/index.tsx
```diff
@@ -39,13 +39,27 @@ const ProfiledApp = Sentry.withProfiler(FeatureFlagApp)
...
+        if (
+          typeof error === 'object' &&
+          error !== null &&
+          'status' in error &&
+          error.status === TOO_MANY_REQUESTS_ERROR_CODE
+        ) {
+          return false
+        }
+        // Otherwise, retry up to 3 times
+        return failureCount < 3
```
nicholas-codecov, line 61: Does failure count start at 0 or 1?

src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
```diff
@@ -129,23 +129,32 @@ function ResetHandler({ reset }) {
     }
   }, [history, queryClient, reset])
 
+  const handleSignOut = async () => {
+    await fetch(`${config.API_URL}/logout`, {
+      method: 'POST',
+      credentials: 'include',
+    })
+    history.replace('/login')
+  }
```
nicholas-codecov, line 138: Hrrm, can we also clear the query client, and reset the error boundary state? Just so that when the user is logged out everything is back into a clean state, and they can start anew.

nicholas-codecov (commented):
Small tweaks

src/index.tsx
ajay-sentry: Failure count starts at 0 for the initial failure, then each retry it gets incremented 1 time

so first retry is at 1, then 2, then 3, then exit function

src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
ajay-sentry: Sure, updated in https://github.com/codecov/gazebo/pull/3119/commits/432553e44ff3f78fa670a97322e0428c0a2ed36b

src/index.tsx
nicholas-codecov: Alrighty perfect, wasn't sure if they'd be like "Oh this failed once ... `failureCount === 1` off the bat

nicholas-codecov (approved):

PR #3124, opened 2024-08-14
feat(sentry): Allow filtering of unwanted third party errors

lforst:
This PR bumps the Sentry webpack plugin to the latest version and adds a `clientKey` alongside the `thirdPartyErrorFilterIntegration` which will set a `third_party_error` tag to error events in Sentry whenever the Sentry SDK detects that an error came from a third-party source (for example a browser extension).

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -5,6 +5,7 @@ const WebpackBar = require('webpackbar')
 const { resolve } = require('path')
 
 const SentryPlugin = sentryWebpackPlugin({
+  applicationKey: 'gazebo',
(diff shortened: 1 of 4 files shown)


lforst:
@nicholas-codecov would be amazing if you could take a look! :)

smeubank:
@nicholas-codecov once you merge we can also discuss creating a dashboard in your codecov sentry org, se we can get a similar view that we have in sentry org

nicholas-codecov (approved):

PR #3138, opened 2024-08-20
feat: Even more language support

nicholas-codecov:
# Description

This PR extends support for our file renderer with the following languages:

- cshtml
- fortran
- fsharp
- powershell
- solidity
- swift
- visual-basic


src/shared/utils/prismLanguageMapper.ts
```diff
@@ -17,6 +17,9 @@ import Prism from 'prism-react-renderer/prism'
  */
 require('prismjs/components/prism-csharp')
 require('prismjs/components/prism-dart')
+require('prismjs/components/prism-cshtml')
```
Swatinem, line 20: just some food for thought: is it possible to load these on-demand after matching the file extension?
its also weird that you are using `require` here instead of bare `import` statements. is there a particular reason for that?

Swatinem (approved):

src/shared/utils/prismLanguageMapper.ts
nicholas-codecov: Uhh was just following along with the example that they provided in the documentation for the package we're using, it seems like just replacing `require` with `import` everything works fine. Gonna quickly make the change to only load in the language required.

nicholas-codecov: Actually my "quick change" is a bit more involved, I think I'll merge what I have now, and tidy things up with your suggestion afterwards, sound all good to you @Swatinem ?

PR #3141, opened 2024-08-21
feat: Add Sentry Profiler and profiling for code renderers

nicholas-codecov:
# Description

This PR adds in Sentry's browser profiling integration, as well as HOC profilers for the old and new code renderers. I'll eventually remove the HOC profilers this is to track some data as we transition from the old to new code renderer, whereas the browser profiler integration will continue to hang around. We'll set the profiling collection rate at 10%, which is a percentage of our trace collection rate, so we don't get flooded with profiles, we can always increase this value if we choose to do so.

# Notable Changes

- Add Sentry browser profiling
- Add HOC profiler to `CodeRenderer` and `VirtualFileRenderer`

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -5,6 +5,7 @@ const defaultConfig = {
   STRIPE_KEY: '',
   SENTRY_ENVIRONMENT: 'staging',
   SENTRY_TRACING_SAMPLE_RATE: 1.0,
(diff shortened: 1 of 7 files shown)


spalmurray-codecov (approved):
Nice! Am pleasantly surprised at how easy this is!

PR #3142, opened 2024-08-21
ref: Optimize SVGs

nicholas-codecov:
# Description

This PR optimizes some of our svgs by reducing their precision.


spalmurray-codecov (approved):
lgtm

Swatinem (approved):
`src/assets/svg/flagManagement.svg` and `src/assets/flagManagement.svg` have the same filename, and they look very similar, the second one seems to be a crop of the first one.
May it be worth as a followup to unify those, and apply the crop client-side somehow?

PR #3143, opened 2024-08-21
fix: Line coverage off by one in VirtualFileRenderer

spalmurray-codecov:
Fixes line coverage being off by one in the new VirtualFileRenderer.


ajay-sentry (approved):
dank!

PR #3145, opened 2024-08-26
fix: Show enable button in components/flags for activated users only

RulaKhaled:
# Description
We show `enable  anayltics` button for non logged in users, this fixes so that if current user is not part of org, we don't show the enabled button.

issue: https://github.com/codecov/feedback/issues/452

# Notable Changes
- Update to reflect is current user part of the org 
- Fix tests 

# Screenshots
Screenshot 2024-08-26 at 5 19 45 PM
Screenshot 2024-08-26 at 5 19 39 PM

# Legal Boilerplate

### DIFF
--- a/src/services/repo/useRepoSettingsTeam.tsx
+++ b/src/services/repo/useRepoSettingsTeam.tsx
@@ -32,7 +32,7 @@ interface FetchRepoSettingsTeamArgs {
 const RequestSchema = z.object({
   owner: z
     .object({
-      isCurrentUserActivated: z.boolean().nullable(),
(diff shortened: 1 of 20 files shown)


ajay-sentry (approved):
looks good!

src/services/repo/useRepoSettingsTeam.tsx
```diff
@@ -64,9 +63,6 @@ query GetRepoSettingsTeam($name: String!, $repo: String!) {
       ... on NotFoundError {
         message
       }
-      ... on OwnerNotActivatedError {
```
RulaKhaled, line 67: this is handled in the /repo page level

PR #3150, opened 2024-08-28
fix: update FailedTestsTab

joseph-sentry:
# Description
This updates the CodecovCLI page in the FailedTestsTab to use the FrameworkTabsCard. This also reorders the steps on the CodecovCLI page such that the JUnit XML generation step comes first for consistency with the other page.

# Screenshots
image


RulaKhaled (approved):
Nice work!

PR #3156, opened 2024-09-03
ref: Convert useUploads to TS + performance improvements

ajay-sentry:
# Description

This PR aims to fix some of the performance issues that were being faced for users with a lot of carry forward flags when looking at a commit from the commits page. I'll detail on the PR itself where a lot of the optimization were coming from since there's quite a few changes, but the "main" meat and potatoes is honestly converting the object we're appending all the flags from an array to a set. This allows us to get reduce our quadratic time complexity to linear, ideally helping quite a bit (in the cases where users may have 1000+ CFF)

Outside of the optimizations, we did TS conversions of a few files, and renaming some of the variables coming from the hook to aid readability, as well as removing the lodash calls/imports since we're kinda using them for things where just doing the "thing" natively is going to give us a more performant and just as readable version of the code.

Some references that helped along the way:

[Countby performance ](https://stackoverflow.com/questions/75883279/whats-the-fastest-way-to-countby-in-javascript)
[benchmark for set vs. includes](https://www.measurethat.net/Benchmarks/Show/23556/0/arrayincludes-vs-sethas)

I _think_ I can look at this page on sentry to compare performance pre/post, but if anyone has a better idea for that would appreciate as well: https://codecov.sentry.io/performance/summary/vitals/?project=5514400&query=&statsPeriod=30d&transaction=%2F%3Aprovider%2F%3Aowner%2F%3Arepo%2Fcommit%2F%3Acommit

Related to https://github.com/codecov/engineering-team/issues/1921 and https://github.com/codecov/engineering-team/issues/1880

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/CommitDetailPage/CommitCoverage/UploadsCard/UploadsCard.jsx
```diff
@@ -7,11 +7,11 @@ import { useUploads } from './useUploads'
 
 import YamlModal from '../YamlModal'
 
-const NULL = 'null'
+const NONE = 'none'
```
ajay-sentry, line 10: we shouldn't use null as a constant value, just removing this code smell

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
+import { z } from 'zod'
+
+import {
+  UploadErrorCodeEnumSchema,
+  UploadStateEnumSchema,
+  UploadTypeEnumSchema,
+} from 'services/commit'
+import { UploadStateEnum, UploadTypeEnum } from 'shared/utils/commit'
+
+interface ErrorObject {
```
ajay-sentry, line 10: all of this stuff at the top is just TS interface boilerplate stuff

src/shared/utils/extractUploads.js
```diff
@@ -1,86 +0,0 @@
-import countBy from 'lodash/countBy'
-import groupBy from 'lodash/groupBy'
-
-import { UploadStateEnum, UploadTypeEnum } from 'shared/utils/commit'
-
-function humanReadableOverview(state, count) {
```
ajay-sentry, line 6: count in this function is unused

src/shared/utils/extractUploads.js
```diff
@@ -1,86 +0,0 @@
...
-  if (state === UploadStateEnum.error) return 'errored'
-  if (state === UploadStateEnum.processed) return 'successful'
-  if (state === UploadStateEnum.complete) return 'carried forward'
-  if (state === UploadStateEnum.started) return 'started'
-}
-
-function deleteDuplicateCFFUploads({ uploads }) {
-  const nonCFFFlags = []
-  const duplicateUploads = uploads && [...uploads]
```
ajay-sentry, line 15: this copy was unneeded

src/shared/utils/extractUploads.js
```diff
@@ -1,86 +0,0 @@
...
-
-  // Get all the flags from "uploaded" uploads
-  duplicateUploads?.forEach((upload) => {
-    if (
-      upload?.uploadType !== UploadTypeEnum.CARRIED_FORWARD &&
-      upload?.flags
-    ) {
-      nonCFFFlags.push(...upload.flags)
```
ajay-sentry, line 23: converted this to a set from an array so we can use .has() instead of .includes()

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
...
+      hasNoUploads: true,
+    }
+  }
+
+  const uploads = deleteDuplicateCFFUploads({ uploads: unfilteredUploads })
+  const hasNoUploads = !uploads || uploads.length === 0
+
+  const {
+    uploadsOverview,
+    providerGroups: groupedUploads,
```
ajay-sentry, line 123: groupedUploads is a better name than "sorted" uploads, because that's what we're doing!

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
...
+      groupedUploads: {},
+      uploadsProviderList: [],
+      uploadsOverview: '',
+      erroredUploads: [],
+      hasNoUploads: true,
+    }
+  }
+
+  const uploads = deleteDuplicateCFFUploads({ uploads: unfilteredUploads })
+  const hasNoUploads = !uploads || uploads.length === 0
```
ajay-sentry, line 119: debated doing an early return for this case, but figured it's a minor optimization at best

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  const {
+    uploadsOverview,
+    providerGroups: groupedUploads,
+    errorProviderGroups: erroredUploads,
+  } = createUploadGroups({ uploads })
+
+  const uploadsProviderList = Object.keys(groupedUploads)
```
ajay-sentry, line 127: we could also do this in the createUploadGroups function, but the O(k) where k = number of keys in groupedUploads is bounded by the number of providers, which is very small in the grand scheme

src/shared/utils/extractUploads.js
```diff
@@ -1,86 +0,0 @@
...
-    }
-  })
-
-  // Filter out cff uploads with repeated flags
-  // We're looping in reverse as splicing reindexes the array, making us
-  // skip the index that's removed. Indices are preserved when you
-  // loop backwards.
-  for (let index = duplicateUploads.length - 1; index >= 0; index--) {
```
ajay-sentry, line 31: Now we're going through all uploads AND all CFF for each upload with a "break" if we find one early. In the worst case this is a quadratic function n*m where n is the number of uploads and m is the number of CFF we have

splice is also an _incredibly_ expensive operation because we are creating a new list each time we call that function.

src/shared/utils/extractUploads.js
```diff
@@ -1,86 +0,0 @@
...
-        }
-      }
-    }
-  }
-
-  return duplicateUploads
-}
-
-const createUploadOverview = ({ uploads }) =>
-  Object.entries(countBy(uploads, (upload) => upload.state))
```
ajay-sentry, line 50: Here we're doing another full pass through all the uploads to generate the "counts" for each state

Then another pass to do the groupBy and filter

Then another pass to do _another_ group by on providers

effectively 3 passes through the uploads to do a couple minor operations. In the new implementation we combine it all into a single pass

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
...
+  return uploads.filter(
+    (upload) =>
+      !(
+        upload?.uploadType === UploadTypeEnum.CARRIED_FORWARD &&
+        upload?.flags?.some((flag) => nonCFFlags.has(flag))
+      )
+  )
+}
+
+const createUploadGroups = ({ uploads }: { uploads: Upload[] }) => {
```
ajay-sentry, line 60: Here we can see the single pass approach, which IMO is just as readable, if not more readable where we have separate "top level" variables for the state counts, provider groups, and errorProviderGroups that all get returned at the the end of the loop

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
...
+      // @ts-ignore this key will always exist if we hit the else; just TS being weird
+      providerGroups[upload.provider].push(upload)
+    }
+
+    if (upload.state === UploadStateEnum.error) {
+      if (!errorProviderGroups[upload.provider]) {
+        errorProviderGroups[upload.provider] = [upload]
+      } else {
+        // @ts-ignore this key will always exist if we hit the else; just TS being weird
+        errorProviderGroups[upload.provider].push(upload)
```
calvin-codecov, line 87: could we use the non-null assertion operator `!` here?

src/shared/utils/extractUploads.ts
```diff
@@ -0,0 +1,136 @@
...
+  const errorProviderGroups: Record<string, Upload[]> = {}
+
+  uploads.forEach((upload) => {
+    // Count the occurrences of each state
+    if (upload.state) {
+      stateCounts[upload.state] = (stateCounts[upload.state] || 0) + 1
+    }
+
+    if (upload.provider === null || upload.provider === undefined) {
+      upload.provider = 'none'
```
calvin-codecov, line 72: could we define `const NONE = 'none'` here and import from UploadsCard?

calvin-codecov (approved):
A couple small comments/suggestions. Nice loop optimizations!

src/shared/utils/extractUploads.ts
ajay-sentry: Good idea, will make the update

src/shared/utils/extractUploads.ts
ajay-sentry: definitely can! Will update

src/shared/utils/extractUploads.ts
ajay-sentry: https://github.com/codecov/gazebo/pull/3156/commits/b1eecfe1b36d9e9ab06e050994e24f78693e4a0a

src/shared/utils/extractUploads.ts
ajay-sentry: https://github.com/codecov/gazebo/pull/3156/commits/b1eecfe1b36d9e9ab06e050994e24f78693e4a0a

PR #3158, opened 2024-09-04
feat: Fire Pendo event on RepoPage for JS and TS users

nicholas-codecov:
# Description

This PR adds in a new `updateOptions` call to Pendo when users navigate to the repo page, and the given repo has JS or TS present. It will only update the options once, and will only run again once the owner value has changed.

Closes codecov/engineering-team#2451

# Notable Changes

- Fix equality check between two objects in current Pendo tracking, this should reduce the amount of times we're calling Pendo, as we're now checking the diff between the owner ids instead of object references
- Add in new hook to trigger `pendo.updateOptions` when JS or TS is present for a given repo
- Create/Update tests

### DIFF
--- a/src/pages/RepoPage/RepoPage.spec.tsx
+++ b/src/pages/RepoPage/RepoPage.spec.tsx
@@ -85,6 +85,58 @@ const mockRepoOverview = ({
   }
 }
 
+const mockOwner = {
+  ownerid: 123,
+  username: 'test-owner',
+  avatarUrl: 'http://127.0.0.1/avatar-url',
+  isCurrentUserPartOfOrg: true,
+  isAdmin: true,
+}
+
+const mockUser = {
+  me: {
+    owner: {
+      defaultOrgUsername: 'codecov',
+    },
+    email: 'jane.doe@codecov.io',
+    privateAccess: true,
+    onboardingCompleted: true,
+    businessEmail: 'jane.doe@codecov.io',
+    termsAgreement: true,
+    user: {
+      name: 'Jane Doe',
+      username: 'janedoe',
+      avatarUrl: 'http://127.0.0.1/avatar-url',
+      avatar: 'http://127.0.0.1/avatar-url',
+      student: false,
+      studentCreatedAt: null,
+      studentUpdatedAt: null,
+      customerIntent: 'PERSONAL',
+    },
+    trackingMetadata: {
+      service: 'github',
+      ownerid: 123,
+      serviceId: '123',
+      plan: 'users-basic',
+      staff: false,
+      hasYaml: false,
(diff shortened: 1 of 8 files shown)


src/pages/RepoPage/RepoPage.spec.tsx
```diff
@@ -245,6 +297,12 @@ describe('RepoPage', () => {
             })
           )
         )
+      }),
+      graphql.query('DetailOwner', (req, res, ctx) => {
+        return res(ctx.data({ owner: mockOwner }))
```
ajay-sentry, line 302: do we need to mock the status here? Curious how this is working otherwise

src/pages/RepoPage/hooks/useJSorTSPendoTracking.spec.tsx
```diff
@@ -0,0 +1,255 @@
...
+        createdAt: 'timestamp',
+        otherGoal: null,
+        typeProjects: [],
+        goals: [],
+      },
+    },
+  },
+}
+
+const mockOverview = (language?: string) => {
```
ajay-sentry, line 54: looking at the tests, I don't think we need this to be optional right? Maybe in the future if we need that behavior we can add it in, it'll help us remove the if statement below and then line 71 can just be languages: [languge]

src/pages/RepoPage/hooks/useJSorTSPendoTracking.ts
```diff
@@ -0,0 +1,78 @@
...
+          isCurrentUserPartOfOrg: ownerData?.isCurrentUserPartOfOrg,
+          isAdmin: ownerData?.isAdmin,
+        }),
+      })
+
+      previousOwner.current = ownerData
+    }
+  }, [ownerData, repoOverview, userData])
+
+  return
```
ajay-sentry, line 77: nit: this should be implicit right?

ajay-sentry (approved):
nothing too pressing, lgtm

src/pages/RepoPage/RepoPage.spec.tsx
nicholas-codecov: I can't find the docs for the version we're on anymore, however iirc it's implicitly set to 200. I did just copy this over from another test.

nicholas-codecov: I can add it in for consistency tho

src/pages/RepoPage/hooks/useJSorTSPendoTracking.ts
nicholas-codecov: yea

PR #3160, opened 2024-09-04
fix: Sentry errors related to unsafe .at()

ajay-sentry:
# Description

Saw a couple sentry errors where we were trying to do an array.at() call on an undefined object

This should fix those couple instances, ~8 users in the last 2 weeks:

https://codecov.sentry.io/issues/?environment=production&environment=dd&environment=enova&environment=epic&environment=roblox&environment=snowflake&project=5514400&project=5215654&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D+.at+is+not+a+function&referrer=issue-list&statsPeriod=14d

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/shared/ContentsTable/utils/adjustListIfUpDir.tsx
```diff
@@ -33,8 +33,12 @@ export function adjustListIfUpDir({
   displayType: 'LIST' | 'TREE'
   rawTableRows: Row[]
 }) {
-  if (treePaths.length > 1 && displayType === displayTypeParameter.tree) {
-    const upDir = treePaths?.at(-2)
+  if (
+    Array.isArray(treePaths) &&
+    treePaths.length > 1 &&
```
ajay-sentry, line 38: if treePaths is a string or is an object with a length key this would fail previously. Just adding an isArray check

src/shared/utils/url.ts
```diff
@@ -10,22 +8,22 @@ export function getFilenameFromFilePath(path: string) {
...
+  if (!path || path === '') {
     return []
   }
-  return path?.split('/') ?? []
+  return path.split('/') ?? []
 }
 
 export function getFilenameFromPath(path?: string) {
-  if (!isString(path)) {
+  if (typeof path !== 'string') {
```
ajay-sentry, line 18: I noticed we use isString a lot, but that's basically just a wrapper for this call... can probably just remove that bloat elsewhere too

src/shared/utils/unsupportedExtensionsMapper.ts
```diff
@@ -348,9 +348,9 @@ const imageFileExtensions = new Set([
   'xpm',
 ])
 
-export function unsupportedExtensionsMapper({ path }) {
-  const fileName = getFilenameFromPath(path)
-  const fileExtension = getFileExtension(fileName)
+export function unsupportedExtensionsMapper({ path }: { path: string }) {
+  const fileName = getFilenameFromPath(path) ?? ''
```
ajay-sentry, line 352: This is the main fix, just default to an empty string if doesn't exist in either of these first two functions

nicholas-codecov (approved):

PR #3161, opened 2024-09-04
fix: hover styles for dark mode on selectors

ajay-sentry:
# Description

This PR aims to quick fix the hover styling on the selectors

This style is also used for the hover on regenerate button, but its a tradeoff I think we're willing to make for now

# Screenshots

https://github.com/user-attachments/assets/e708f47a-91d0-417f-b873-6eb75b5ecf00

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/ui/Select/Select.jsx
+++ b/src/ui/Select/Select.jsx
@@ -15,7 +15,7 @@ const SelectClasses = {
   root: 'relative',
   item: 'block cursor-pointer py-1 px-3 text-sm font-normal',
   button:
-    'flex justify-between items-center w-full rounded bg-ds-background text-left whitespace-nowrap disabled:text-ds-gray-quaternary disabled:bg-ds-gray-primary disabled:border-ds-gray-tertiary focus:outline-1',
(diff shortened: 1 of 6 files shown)


src/ui/Select/Select.jsx
```diff
@@ -15,7 +15,7 @@ const SelectClasses = {
   root: 'relative',
   item: 'block cursor-pointer py-1 px-3 text-sm font-normal',
   button:
-    'flex justify-between items-center w-full rounded bg-ds-background text-left whitespace-nowrap disabled:text-ds-gray-quaternary disabled:bg-ds-gray-primary disabled:border-ds-gray-tertiary focus:outline-1',
+    'flex justify-between items-center w-full rounded bg-ds-container text-left whitespace-nowrap disabled:text-ds-gray-quaternary disabled:bg-ds-gray-primary disabled:border-ds-gray-tertiary focus:outline-1',
```
ajay-sentry, line 18: this is for the historical trend component and probably other stuff too

suejung-sentry (approved):
nice thanks for grabbing that!

PR #3169, opened 2024-09-09
feat: Use new coverage analytics graphql schema

suejung-sentry:
Point fields to the new resolvers for the nested types in CoverageAnalytics (changes in api [here](https://github.com/codecov/codecov-api/pull/806)).

Tested by running against the api changes in staging that the spots this is used (coverage graphs) have no difference in behavior.

Closes https://github.com/codecov/engineering-team/issues/2281

some example side by sides:
Screenshot 2024-09-26 at 8 12 56 PM
Screenshot 2024-09-26 at 8 11 59 PM


src/services/charts/useReposCoverageMeasurements.ts
```diff
@@ -40,15 +44,17 @@ const query = `
     $isPublic: Boolean
   ) {
     owner(username: $owner) {
-      measurements(
```
suejung-sentry, line 43: TODO - remove this use case - owner.measurements is different than owner.repo.measurements (the one i'm after)

src/services/charts/useBranchCoverageMeasurements.tsx
```diff
@@ -33,11 +33,10 @@ const GetBranchCoverageMeasurementsSchema = z.object({
...
       repository: z
         .discriminatedUnion('__typename', [
-          z
-            .object({
-              __typename: z.literal('Repository'),
-            })
-            .merge(MeasurementsSchema),
+          z.object({
+            __typename: z.literal('Repository'),
+            coverageAnalytics: MeasurementsSchema.optional(),
```
nicholas-codecov, line 38: Can we swap this to `nullable`, in GQL there's no concept of `undefined`. So if this returns `null` than the schema parsing will fail.

nicholas-codecov (approved):
One small comment to take a peak.

src/services/charts/useBranchCoverageMeasurements.tsx
suejung-sentry: Great catch - thank you!! today I learned!

PR #3173, opened 2024-09-09
fix: Plan option not updating when plan param is set

nicholas-codecov:
# Description

This PR updates the plan option buttons to set the URL plan param when selecting plan option, previously this was not being done, and we were using it to determine the state of the button.

# Notable Changes

- Set plan param when selecting plan option
- Update tests

# Screenshots

Before:

https://github.com/user-attachments/assets/d624a352-770c-41dd-9ac5-4052079f8eb9

After:

https://github.com/user-attachments/assets/23c2d706-dcdb-4e5a-80cc-19d58c4dde9b


spalmurray-codecov (approved):
Nice

PR #3174, opened 2024-09-10
Tests analytics page V2

RulaKhaled:
# Description
We are developing against this. Please open  a PR of you wish to push changes to this branch. 

Design: https://www.figma.com/design/TcppABudxnslgEmaXm0B9O/GH-2028?node-id=1-2&node-type=CANVAS&t=QWvqJ1kK3WG1pFLS-0

# Link to Sample Entry

# Legal Boilerplate


src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useTestResultsTestSuites/useTestResultsTestSuites.tsx
```diff
@@ -0,0 +1,97 @@
...
+          provider,
+          owner,
+          repo,
+          term,
+        },
+      }).then((res) => {
+        const parsedData = TestResultsTestSuitesSchema.safeParse(res?.data)
+
+        if (!parsedData.success) {
+          return Promise.reject({
```
suejung-sentry, line 75: Not blocking or needed now - with the new pattern per [here](https://github.com/codecov/gazebo/pull/3436) if you use the helper `rejectNetworkError` it will send to sentry which is handy

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/MetricsSection/MetricsSection.tsx
```diff
@@ -0,0 +1,451 @@
...
+              totalFails={aggregates?.totalFails}
+              totalFailsPercentChange={aggregates?.totalFailsPercentChange}
+              updateParams={updateParams}
+              isSelected={queryParams?.parameter === 'FAILED_TESTS'}
+            />
+            <TotalSkippedTestsCard
+              totalSkips={aggregates?.totalSkips}
+              totalSkipsPercentChange={aggregates?.totalSkipsPercentChange}
+              updateParams={updateParams}
+              isSelected={queryParams?.parameter === 'SKIPPED_TESTS'}
```
suejung-sentry, line 442: Not blocking - would be nice to use the constant defined elsewhere (I think `TestResultsFilterParameter`)

ajay-sentry (approved):

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -84,44 +99,74 @@ export function getSortingOption(
...
+    columnHelper.accessor('avgDuration', {
+      header: () => 'Avg duration',
+      cell: (info) => `${(info.renderValue() ?? 0).toFixed(3)}s`,
+    }),
+    columnHelper.accessor('failureRate', {
+      header: () => 'Failure rate',
+      cell: (info) => {
+        const value = (info.renderValue() ?? 0) * 100
+        const isInt = Number.isInteger(info.renderValue())
+        return isInt ? `${value}%` : `${value.toFixed(2)}%`
```
suejung-sentry, line 137: Technically if info.renderValue() is not a number this could end up saying `NaN%`. Maybe an edge case don't need to expect or can add to post-release polish as needed

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useInfiniteTestResults/useInfiniteTestResults.tsx
```diff
@@ -127,6 +163,11 @@ interface UseTestResultsArgs {
   repo: string
   filters?: {
     branch?: string
+    flags?: string[]
+    interval?: MeasurementInterval
+    parameter?: TestResultsFilterParameterType
+    term?: string
+    test_suites?: string[]
```
suejung-sentry, line 170: I was wondering why this one was snake case - guess it's that way in `api` - https://github.com/codecov/codecov-api/blob/7857053efdb545165cbce4ffc0bd9526d82a73aa/graphql_api/types/inputs/test_results_filters.graphql#L4

Seems ok as is - if we ever have time to clean it up can grab it later

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useTestResultsAggregates/useTestResultsAggregates.tsx
```diff
@@ -0,0 +1,140 @@
...
+        .nullable(),
+      repository: z.discriminatedUnion('__typename', [
+        z.object({
+          __typename: z.literal('Repository'),
+          private: z.boolean().nullable(),
+          defaultBranch: z.string().nullable(),
+          testAnalytics: z
+            .object({
+              testResultsAggregates: z.object({
+                totalDuration: z.number(),
```
suejung-sentry, line 26: In too deep with this now but would be nice to have units on these `duration` fields next time. Maybe with the polish epic can add comments indicating these (`totalDuration`, `slowestTestsDuration`) represent seconds either here and/or in `api`

suejung-sentry (approved):
Approved this massive effort - nice work, team! 🎉 🎉 

I added some nits that prove I actually read it :joy: but don't need to be addressed before shipping this iteration (in fact, you probably don't want to since it can potentially introduce new errors). I also did a QA sweep last week on the preview deploy and added some small tickets to the polish epic. Amazing stuff! 🚀

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useTestResultsTestSuites/useTestResultsTestSuites.tsx
ajay-sentry: PR up here! https://github.com/codecov/gazebo/pull/3484

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/MetricsSection/MetricsSection.tsx
ajay-sentry: PR up here! https://github.com/codecov/gazebo/pull/3484

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
RulaKhaled: For the failure rate (the value here), the API always returns a float, so we're safe!
ref: https://github.com/codecov/codecov-api/blob/main/graphql_api/types/test_results/test_results.graphql#L5

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useInfiniteTestResults/useInfiniteTestResults.tsx
RulaKhaled: Here's the issue [The test_suites key in the test results filter needs to be camel-cased to align with naming standards #2878](https://github.com/codecov/engineering-team/issues/2878)

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useTestResultsAggregates/useTestResultsAggregates.tsx
RulaKhaled: For sure, could be a nice-to-have in the GQ schema playground too! [Add comments indicating these (totalDuration, slowestTestsDuration) represent seconds #2879](https://github.com/codecov/engineering-team/issues/2879)

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useTestResultsFlags/useTestResultsFlags.tsx
```diff
@@ -0,0 +1,97 @@
...
+        },
+      }).then((res) => {
+        const parsedData = TestResultsFlagsSchema.safeParse(res?.data)
+
+        if (!parsedData.success) {
+          return rejectNetworkError({
+            status: 404,
+            data: {},
+            dev: 'useTestResultsFlags - 404 Failed to parse data',
+          } satisfies NetworkErrorObject)
```
nicholas-codecov, line 79: Can we add in `error: parsedData.error` to this object (and the other unsuccessful parsing of data errors in the new hooks), the way that the helper function works is by checking the status value, and if an error is passed. You can see the underlying logic [here](https://github.com/codecov/gazebo/blob/main/src/shared/api/helpers.ts#L18-L39).

We only have an error for when parsing is unsuccessful unlike our other errors, so that is why the error value is not required, sorry for any confusion, will be working on better docs to help address these shortfalls.

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -84,44 +99,74 @@ export function getSortingOption(
...
+  hideFlakeRate: boolean
+  interval?: MeasurementInterval
+}) => {
+  const baseColumns = [
+    columnHelper.accessor('name', {
+      header: () => 'Test name',
+      cell: (info) => info.renderValue(),
+    }),
+    columnHelper.accessor('avgDuration', {
+      header: () => 'Avg duration',
```
nicholas-codecov, line 129: nit: Should `Avg` have a `.` at the end: `Avg.`?

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -153,26 +216,84 @@ const FailedTestsTable = () => {
     ordering: getSortingOption(sorting),
     filters: {
       branch: branch ? getDecodedBranch(branch) : undefined,
+      flags: flags as string[],
+      // eslint-disable-next-line camelcase
+      test_suites: testSuites as string[],
```
nicholas-codecov, line 221: Could these values actually be undefined when being passed here, and would that cause any issues when querying? 

Could they be set to empty arrays instead of `undefined` above, and get rid of the type casting?

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -153,26 +216,84 @@ const FailedTestsTable = () => {
...
+
+        return {
+          name: result.name,
+          avgDuration: result.avgDuration,
+          failureRate: result.failureRate,
+          flakeRate: FlakeRateContent,
+          commitsFailed: result.commitsFailed,
+          updatedAt: result.updatedAt,
+        }
+      }) ?? []
```
nicholas-codecov, line 275: nit: Is this nullish coalescing operator needed here?

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/BranchSelector/BranchSelector.tsx
```diff
@@ -76,20 +76,30 @@ const BranchSelector = () => {
...
 
+  const sortedBranchList = overview?.defaultBranch
+    ? [
+        // Pins the default branch to the top of the list always, filters it from results otherwise
+        { name: overview.defaultBranch, head: null },
+        ...(branchList?.branches?.filter(
+          (branch) => branch.name !== overview.defaultBranch
+        ) ?? []),
+      ]
+    : (branchList?.branches ?? [])
```
nicholas-codecov, line 87: `m`: What are your thoughts on utilizing `Set` here? I think we'd be able to drop the `filter` which would help us skip iterating over the entire array on each render, and reduce the overall complexity.

Also, what do you think about memo'izing this `sortedBranchList`?

Docs for `Set` can be found [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#description), if you're looking for more details.

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useFlakeAggregates/useFlakeAggregates.tsx
```diff
@@ -0,0 +1,120 @@
...
+              flakeAggregates(interval: $interval) {
+                flakeCount
+                flakeCountPercentChange
+                flakeRate
+                flakeRatePercentChange
+              }
+          }
+        }
+        ... on NotFoundError {
+          message
```
nicholas-codecov, line 53: `m`: Wondering if we can add in handle this potential union type `OwnerNotActivatedError`? What do you think about handling the error, with our schema validation, and also in our resolver, until it has been removed from the schema fully?

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useTestResultsAggregates/useTestResultsAggregates.tsx
```diff
@@ -0,0 +1,140 @@
...
+                totalFails
+                totalFailsPercentChange
+                totalSkips
+                totalSkipsPercentChange
+                }
+          }
+        }
+        ... on NotFoundError {
+          message
+        }
```
nicholas-codecov, line 78: Same thing here, regarding handling `OwnerNotActivatedError` for the time being.

nicholas-codecov (commented):
Couple of comments to take a peak at before merging.

There's a couple of duplicates that I didn't want to add in multiple times but they are:

- Handling `OwnerNotActivatedError` for the time being in queries that hit the `repository` field until it's removed from the schema
- Add `error: parsedData.error` to `rejectNetworkError` when unsuccessful parsing occurs

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useFlakeAggregates/useFlakeAggregates.tsx
RulaKhaled: I'm certain we don't need this union type, as it's already handled in the repository-level validations. Duplicating it here would just add dead code

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
ajay-sentry: good catch! With line 239 its not needed, we can make an update

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/BranchSelector/BranchSelector.tsx
ajay-sentry: I think it's actually a similar amount of complexity with using Set vs. Array, both from a runtime front and the code readability front; wasn't too sure which you meant.

Runtime front both are still O(n) since you need to populate the set with N elements before doing a removal, sets also don't respect ordering out the both so we'd still need to do a conversion back to array after right? It may end up being slower actually.

Readability front, were you thinking something like this?

`
const branchSet = Set(...branchList.branches)
Set.remove({ name: overview.defaultBranch, head: null })

 { name: overview.defaultBranch, head: null },
       ...branchSet,
 `

Memoizing is a good idea though, I think it'd help with either the Set or the Array approach.

nicholas-codecov: I was more thinking along the lines of:

```
const branchSet = Set({ name: overview.defaultBranch, head: null }, ...branchList.branches)
```

items added later won't override former items

ajay-sentry: > items added later won't override former items

ahh okay I didn't realize sets were orderedSets by default, didn't get your point originally but that makes sense. So we're looking at something like this then right? 

```
  const sortedBranchList = useMemo(() => {
    return branchList?.branches
      ? [
          // Pins the default branch to the top of the list always, filters it from results otherwise
          new Set([
            overview?.defaultBranch && {
              name: overview.defaultBranch,
              head: null,
            },
            ...branchList.branches,
          ]),
        ]
      : []
  }, [branchList?.branches, overview?.defaultBranch])
```

It's around the same complexity imo haha but since I already did that thought exercise I can make the change for ya later

ajay-sentry: I lied it's actually this

```
  const sortedBranchList = useMemo(() => {
    if (!branchList?.branches) return []

    return [
      ...(overview?.defaultBranch
        ? [
            {
              name: overview.defaultBranch,
              head: null,
            },
          ]
        : []),
      ...branchList.branches,
    ]
  }, [branchList?.branches, overview?.defaultBranch])
```

&& operator isn't so kind

ajay-sentry: Yeah actually this doesn't seem to work bc of the && operator, if you have an elegant solution happy to try it out otherwise I'll leave as is

nicholas-codecov: This is def not a blocker, we can go with this for the launch today and take a peak as a fast follow

PR #3179, opened 2024-09-10
feat: Update assets table to use paginated hook

nicholas-codecov:
# Description

This PR updates the assets table on the bundles tab to use the newly added paginated hook to load only 20 assets at a time, and enable infinite scrolling to help reduce the load times on the bundles tab.

Closes codecov/engineering-team#2489

# Notable Changes

- Update `AssetsTable` to hook into paginated hook and add infinite scrolling
- Update `useBundleAssetsTable` to hook into paginated hook
- Update types and args for `usePagedBundleAssets`
- Update tests


src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.spec.tsx
```diff
@@ -349,39 +352,39 @@ describe('AssetsTable', () => {
 
     describe('renders table rows', () => {
       it('renders asset column', async () => {
```
ajay-sentry, line 354: curious why we have these tests split out into separate component renders vs. just confirming all the columns in a single test? This is a common pattern I've noticed throughout the codebase and just wanted some clarity on

src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.spec.tsx
```diff
@@ -404,31 +407,100 @@ describe('AssetsTable', () => {
...
+        it('sorts the table by size', async () => {
+          const { user, mockOrdering } = setup({ multipleAssets: false })
+          render(<AssetsTable />, { wrapper })
+
+          const loadTimeColumn = await screen.findByText(
+            'Estimated load time (3G)'
+          )
+          await user.click(loadTimeColumn)
+
+          await waitFor(() => expect(mockOrdering).toHaveBeenCalledWith('SIZE'))
```
ajay-sentry, line 458: shouldn't this be load Time? Or still size?

nicholas-codecov: So ... load time is directly related to the size, so we just use the size

src/services/bundleAnalysis/usePagedBundleAssets.tsx
```diff
@@ -207,8 +207,8 @@ interface UsePagedBundleAssetsArgs {
     reportGroups?: string[]
     loadTypes?: string[]
   }
-  orderingDirection: 'ASC' | 'DESC'
-  ordering: 'NAME' | 'SIZE' | 'TYPE'
+  orderingDirection?: 'ASC' | 'DESC'
+  ordering?: 'NAME' | 'SIZE' | 'TYPE'
```
spalmurray-codecov, line 211: It slightly annoys me that we don't have the ordering direction defined somewhere shared lol

src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.spec.tsx
ajay-sentry: It's not immediately clear from why that would be the case :) Maybe I just missed a comment somewhere though

src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.spec.tsx
nicholas-codecov: I personally prefer the more verbose style like the above, because it explains exactly what the test is checking, so if that test fails I know instantly what is going on. Whereas if it's checking all the columns I have to look deeper into which test failed.

As well it probably plays a bit nicer with test analytics so we can easily see which test is failing since we display it by test name/description.

src/services/bundleAnalysis/usePagedBundleAssets.tsx
nicholas-codecov: I'll fix in a follow up

src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.spec.tsx
nicholas-codecov: I'll add a comment in for more context

spalmurray-codecov (approved):
Code lgtm, tested preview deploy and seems to be working as expected!

src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.spec.tsx
ajay-sentry: I see what you mean, though the error message from the test should still be quite clear when we do the expect() statement and a certain line fails.

The pro with this approach, as you mentioned, is if multiple columns are failing we see all fails at once. I'm fine either way, though I personally prefer consolidation for the reduced render and verbosity pros

spalmurray-codecov: It does come at the cost of more tests/longer CI runs. Can't imagine we'd save a huge amount of time consolidating these though. If it were significant then it may not be worth the time saved debugging. Could be interesting to compare the runtimes.

nicholas-codecov: Also with the upcoming move to Vitest, we should hopefully see test CI times go down.

ajay-sentry: Yeah at the end of the day it's mostly personal preference.

Knowing that our GH bill for the runners isn't too large either, that's a valid approach as well (in the short/med term at least)

PR #3194, opened 2024-09-12
fix: Users unable to cancel when have adblock

ajay-sentry:
# Description

**PROBLEM**
- Users with adblock on weren't able to cancel their Codecov subscriptions.
- This was due to the script being correctly "injected" into the DOM, but that being the "guard" for us to signal that "baremetrics was blocked," which is actually incorrect
- Maybe that worked previously where adblockers were preventing scripts from loading altogether; but now I guess it does inject scripts but just blocks them from executing.

Today I learned that scripts have .onload() and .onerror() handlers that allow you to throw errors if they aren't injected properly into the DOM!

Previously we were relying on doing all of that on "our own" in a bit of a hacky way; and it turns out just relying on those aspects and removing the unneeded code was all we needed to get baremetrics to work with and without adblock.

**References:**
https://app.baremetrics.com/cancellations/configure/responses/widget/embed-code
https://help.baremetrics.com/en/articles/5380579-embedding-cancellation-insights-code
https://help.baremetrics.com/en/articles/5380562-cancellation-insights-your-101-setup-guide

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/DowngradePlan/CancelButton/useBarecancel.js
+++ b/src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/DowngradePlan/CancelButton/useBarecancel.js
@@ -11,25 +11,29 @@ export function useBarecancel({ customerId, callbackSend, isModalOpen }) {
   useEffect(() => {
     let unMounted = false
     if (isModalOpen) {
-      loadBaremetrics().then(() => {
-        window.barecancel.params = {
-          /* eslint-disable camelcase */
-          access_token_id: config.BAREMETRICS_TOKEN,
-          customer_oid: customerId,
-          comment_required: true,
-          callback_send: () => {
-            memoizedSuccess()
-          },
-          callback_error: (error) => {
-            // You can also catch any errors that happen when sending the cancellation event to Baremetrics.
-            // For example, if Baremetrics returns that the customer does not have an active subscription.
-            console.error(error)
-          },
-          /* eslint-enable camelcase */
-        }
-        if (unMounted) return
-        setWasBlocked(false)
-      })
+      loadBaremetrics()
+        .then(() => {
+          window.barecancel.params = {
+            /* eslint-disable camelcase */
+            access_token_id: config.BAREMETRICS_TOKEN,
+            customer_oid: customerId,
+            comment_required: true,
+            callback_send: () => {
+              memoizedSuccess()
+            },
+            callback_error: (error) => {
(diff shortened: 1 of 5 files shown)


src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/DowngradePlan/CancelButton/utils.js
```diff
@@ -17,8 +17,17 @@ export function loadBaremetrics() {
     script.src =
       'https://baremetrics-barecancel.baremetrics.com/js/application.js'
     script.dataset.testid = 'baremetrics-script'
+
+    // This stuff controls the logic in useBareCancel to make sure if the script isn't loaded to cancel without Baremetrics
+    script.onload = function () {
+      return resolve()
+    }
+
+    script.onerror = function () {
```
ajay-sentry, line 26: This is needed to trigger the catch and "block" stuff in useBarecancel.js

src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/DowngradePlan/CancelButton/useBarecancel.js
```diff
@@ -11,25 +11,29 @@ export function useBarecancel({ customerId, callbackSend, isModalOpen }) {
...
+              // You can also catch any errors that happen when sending the cancellation event to Baremetrics.
+              // For example, if Baremetrics returns that the customer does not have an active subscription.
+              console.error(error)
+            },
+            /* eslint-enable camelcase */
+          }
+          if (unMounted) return
+          setWasBlocked(false)
+        })
+        .catch(() => {
```
ajay-sentry, line 34: only diff here is to set blocked to true if we have an error

src/pages/PlanPage/subRoutes/CancelPlanPage/subRoutes/DowngradePlan/CancelButton/utils.js
```diff
@@ -8,17 +8,22 @@ export function getEndPeriod(unixPeriodEnd) {
 }
 
 export function loadBaremetrics() {
-  return new Promise((resolve) => {
-    if (window.barecancel && window.barecancel.created) {
```
ajay-sentry, line 12: I didn't realize til this morning that we were incorrectly resolving this promise when we shouldn't have been because this attribute _was_ in the DOM but didn't actually mean the script was "working"

calvin-codecov (approved):
LGTM

PR #3196, opened 2024-09-13
feat: Render okta errors based on error param

RulaKhaled:
# Description
Surfacing okta errors, a full list of the error codes can be found here: https://developer.okta.com/docs/reference/error-codes/#example-errors-for-openid-connect-and-social-login

issue: https://github.com/codecov/engineering-team/issues/2347 

PS. The design was discussed with Kyle and dismiss button visibility in dark mode/light will be handled in a different issue. 

# Notable Changes
- Added a new variant for the top banner 
- Created new top banners

# Screenshots
Screenshot 2024-09-13 at 10 45 37 AM
Screenshot 2024-09-13 at 1 49 49 PM

# Legal Boilerplate

### DIFF
--- a/src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
+++ b/src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
@@ -1,0 +1,46 @@
+enum ErrorType {
+  UnauthorizedClient = 'unauthorized_client',
+  AccessDenied = 'access_denied',
+  UnsupportedResponseType = 'unsupported_response_type',
+  UnsupportedResponseMode = 'unsupported_response_mode',
+  InvalidScope = 'invalid_scope',
+  ServerError = 'server_error',
+  TemporarilyUnavailable = 'temporarily_unavailable',
+  InvalidClient = 'invalid_client',
+  LoginRequired = 'login_required',
+  InvalidRequest = 'invalid_request',
+  UserCanceledRequest = 'user_canceled_request',
+}
+
+export const errorMessages: Record<ErrorType, string> = {
+  [ErrorType.UnauthorizedClient]:
+    "Unauthorized client: The client isn't authorized to request an authorization code using this method. Please reach out to your administrator.",
+  [ErrorType.AccessDenied]:
+    'Access denied: The resource owner or authorization server denied the request. Please reach out to your administrator.',
+  [ErrorType.UnsupportedResponseType]:
+    "Unsupported response type: The authorization server doesn't support obtaining an authorization code using this method. Please reach out to your administrator.",
+  [ErrorType.UnsupportedResponseMode]:
+    "Unsupported response mode: The authorization server doesn't support the requested response mode. Please reach out to your administrator.",
+  [ErrorType.InvalidScope]:
+    'Invalid scope: The requested scope is invalid, unknown, or malformed. Please reach out to your administrator.',
+  [ErrorType.ServerError]:
+    'Server error: The authorization server encountered an unexpected condition that prevented it from fulfilling the request. Please try again later or contact support.',
+  [ErrorType.TemporarilyUnavailable]:
+    'Temporarily unavailable: The authorization server is currently unable to handle the request due to temporary overloading or maintenance. Please try again later.',
+  [ErrorType.InvalidClient]:
+    "Invalid client: The specified client isn't valid. Please reach out to your administrator.",
+  [ErrorType.LoginRequired]:
+    "Login required: The client specified not to prompt, but the user isn't signed in. Please sign in and try again.",
+  [ErrorType.InvalidRequest]:
+    "Invalid request: The request parameters aren't valid. Please try again or contact support.",
+  [ErrorType.UserCanceledRequest]:
+    'Request canceled: User canceled the social sign-in request. Please try again if this was unintentional.',
+}
+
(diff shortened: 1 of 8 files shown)


src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
```diff
@@ -0,0 +1,45 @@
...
+    "Invalid request: The request parameters aren't valid. Please try again or contact support.",
+  [ErrorType.UserCanceledRequest]:
+    'Request canceled: User canceled the social sign-in request. Please try again if this was unintentional.',
+} as const
+
+export const getErrorMessage = (error: string): string => {
+  if (error in errorMessages) {
+    return errorMessages[error as keyof typeof errorMessages]
+  }
+  return 'An unknown error occurred. Please try again or contact support.'
```
nicholas-codecov, line 44: We could log this to Sentry via `Sentry.captureMessage` and include the error type so we can keep tabs on cases we might have missed (or if they change for some reason)

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
```diff
@@ -0,0 +1,45 @@
...
+  UnsupportedResponseType: 'unsupported_response_type',
+  UnsupportedResponseMode: 'unsupported_response_mode',
+  InvalidScope: 'invalid_scope',
+  ServerError: 'server_error',
+  TemporarilyUnavailable: 'temporarily_unavailable',
+  InvalidClient: 'invalid_client',
+  LoginRequired: 'login_required',
+  InvalidRequest: 'invalid_request',
+  UserCanceledRequest: 'user_canceled_request',
+} as const
```
nicholas-codecov, line 13: Is there a benefit to using this enum object here? We don't seem to be accessing the value outside of the object definition, could it just be simplified to one object?

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
nicholas-codecov, line 1: Is there a benefit to having these split out from the component where they're being used? Exporting `getErrorMessage` by itself may be confusing for others as it's not clear what error messages it's getting, and may be accidentally imported elsewhere.

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
RulaKhaled: We're aligning with JavaScript's property naming conventions by receiving snake_case values and converting them to the preferred camelCase format

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
RulaKhaled: The object is big and messy. I think it’s cleaner this way. We don’t have a rigid structure for enum files in the codebase, so I don’t think it’s confusing, especially since it’s a file nested within the banner folder

nicholas-codecov: Yeah, they tend to get large and a little messy, and that's alright. If the error messages only relates to Okta and this component specifically, it might make sense to just move the code definitions inside the banner so they are more tightly coupled or rename the export to getOktaErrorMessage.

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
nicholas-codecov: I think I might not have explained myself well, but what I wanted to say is that since we are not actually accessing the enum properties and checking for individual keys inside the getErrorMessage function or anywhere else, there is no need to align this and we can just type it as:
```ts
const OktaErrorMappings = { 'server_error': "...", } as const
function getOktaErrorMessage(status){
  return OktaHTTPErrorMappings[status] ?? 'default status'
}
```
This way we can drop the enum and make keep the definition in a single type as opposed to having to define enum + record to make changes in the future.

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
RulaKhaled: Cool I'll rename it to make it even more explicit for its use

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
RulaKhaled: I understand your point; the way you listed it is more direct. However, this goes against our ESLint rules and JavaScript's preferred key structure. It will also make more sense if we ever need to access the enum keys, instead of having to refactor it later

Screenshot 2024-09-16 at 10 21 06 AM

src/shared/GlobalTopBanners/OktaErrorBanners/enums.ts
nicholas-codecov: Another thing you may want to add in alongside the fingerprint is a tag. I added one to the `prismLanguageMapper` capture message and this is what it provides in the Sentry UI:

Screenshot 2024-09-16 at 09 20 16

And the code is pretty simple to add in:

```ts
Sentry.captureMessage(`Unsupported language type for filename ${fileName}`, {
  fingerprint: ['unsupported-prism-language'],
  tags: {
    'file.extension': fileExtension,
  },
})
```

PR #3200, opened 2024-09-16
fix: Show tests tab for public repos without a user being logged in

RulaKhaled:
# Description
As long as
- a repo is sending test failures
- is public

Show the tab even if a user is not logged in (icognito mode etc).

# Notable Changes
- Update repo tabs 
- Update repo routes 

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/RepoPage/RepoPage.tsx
+++ b/src/pages/RepoPage/RepoPage.tsx
@@ -316,7 +316,7 @@ function RepoPage() {
           isRepoPrivate={isRepoPrivate}
           isCurrentUserActivated={isCurrentUserActivated}
           testAnalyticsEnabled={repoOverview?.testAnalyticsEnabled}
-          isCurrentUserPartOfOrg={repoData?.isCurrentUserPartOfOrg || false}
+          isCurrentUserPartOfOrg={!!repoData?.isCurrentUserPartOfOrg}
         />
       </Suspense>
     </div>


src/pages/RepoPage/RepoPage.tsx
```diff
@@ -310,6 +316,7 @@ function RepoPage() {
           isRepoPrivate={isRepoPrivate}
           isCurrentUserActivated={isCurrentUserActivated}
           testAnalyticsEnabled={repoOverview?.testAnalyticsEnabled}
+          isCurrentUserPartOfOrg={repoData?.isCurrentUserPartOfOrg || false}
```
ajay-sentry, line 319: do we need the || false check here? if repoData?.isCurrentUserPartOfOrg is undefined it'll still be a falsey value

src/pages/RepoPage/RepoPageTabs.tsx
```diff
@@ -98,11 +98,18 @@ export const useRepoTabs = ({ refetchEnabled }: UseRepoTabsArgs) => {
     })
   }
 
-  if (isCurrentUserPartOfOrg) {
+  if (repoOverview?.testAnalyticsEnabled) {
```
ajay-sentry, line 101: Nice catch!

src/pages/RepoPage/RepoPageTabs.spec.tsx
```diff
@@ -1009,4 +1009,79 @@ describe('useRepoTabs', () => {
...
+
+        const expectedTab = [
+          {
+            pageName: 'failedTestsOnboarding',
+            children: expect.anything(),
+          },
+        ]
+
+        await waitFor(() =>
+          expect(result.current).not.toEqual(
```
ajay-sentry, line 1080: Should we put what is expected here rather than .not() and have the variable name "expectedTab?" Would reduce cognitive burden here imo

Alloneway30 (approved):

src/pages/RepoPage/RepoPage.tsx
RulaKhaled: yes if we want to explicitly set it to boolean, it's the pattern used in this component (see isRepoActivated, isRepoActive)

src/pages/RepoPage/RepoPageTabs.spec.tsx
RulaKhaled: Sure i will add that as part of the test.

I'm specifically checking for the tab's existence because if analytics isn't enabled, it shouldn't appear—unless it's the specific case being tested. It's a bit more unique to this scenario.

RulaKhaled: Also nice catch for the variable name!

src/pages/RepoPage/RepoPage.tsx
ajay-sentry: Ah okay makes sense! Only thing I'd say then if we're looking for a boolean is we can utilize the !! syntax to cast a variable to its boolean equivalent

But that's a nit if anything

ajay-sentry (approved):
looks good!

src/pages/RepoPage/RepoPage.tsx
RulaKhaled: sure thing!

PR #3207, opened 2024-09-18
fix: Guard when calling match() on undefined

ajay-sentry:
# Description

Aiming to fix: https://codecov.sentry.io/issues/5616514924

Not sure exactly where this sentry issue is coming from, but my hunch is it's from the only place we're calling .match() on something here: https://github.com/codecov/gazebo/blob/7a43fe74ff12c04adefdbb971009a9282e79cc37/src/ui/A/A.jsx#L37-L38

Tracing that function up though, even weirder because we have a guard checking for if href exists here:
https://github.com/codecov/gazebo/blob/7a43fe74ff12c04adefdbb971009a9282e79cc37/src/ui/A/A.jsx#L78-L80

Technically it will resolve to true for _any_ truthy value, such as true, 1, '', etc. In those cases, we also wouldn't be throwing an error saying "can't read properties of undefined" either, but something more like these:

Screenshot 2024-09-18 at 9 36 24 AM

In any case, I don't think this PR "hurts" us. If the URL doesn't exist, we should nav the user back to the top of the application.

**TESTING**

- Went through and clicked around everywhere to try and reproduce, couldn't find anything 
- The app still functions correctly

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
+++ b/src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
@@ -47,7 +47,7 @@ const renderItem = ({ item }) => {
   if (item?.isProvider) {
     return (
       <div className="flex h-8 items-center gap-2">
-        <A pathname={{ pageName: 'codecovAppInstallation' }}>
+        <A to={{ pageName: 'codecovAppInstallation' }}>
           <Icon name="plus-circle" />
           <span>Install Codecov GitHub app</span>
         </A>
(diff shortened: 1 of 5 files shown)


src/pages/DefaultOrgSelector/DefaultOrgSelector.jsx
```diff
@@ -47,7 +47,7 @@ const renderItem = ({ item }) => {
   if (item?.isProvider) {
     return (
       <div className="flex h-8 items-center gap-2">
-        <A pathname={{ pageName: 'codecovAppInstallation' }}>
+        <A to={{ pageName: 'codecovAppInstallation' }}>
```
ajay-sentry, line 50: interestingly, this was the only spot I could find that was using something outside of "href" or "to." Just changed it to "to" for parity

nicholas-codecov (approved):

PR #3241, opened 2024-09-23
chore: Update ui/SettingsDescriptor to Vitest

nicholas-codecov:
# Description

This PR updates the `ui/SettingsDescriptor` to TS and tests to Vitest


calvin-codecov (approved):
LGTM

Are the new files too different for github to keep thinking it's a filename change instead of a delete/add?

PR #3251, opened 2024-09-23
chore: Remove ui/List component

nicholas-codecov:
# Description

Noticed that we aren't using this component anymore, so we might as well get rid of it.


spalmurray-codecov (approved):
NICE

PR #3278, opened 2024-09-23
chore: Update shared/api tests to Vitest

nicholas-codecov:
# Description

This PR updates the tests in `shared/api` to Vitest. I had to do some extra work to get the tests all playing nicely in the api tests hence the amount of changes.


src/shared/api/api.test.ts
```diff
@@ -410,49 +445,25 @@ describe('when using a graphql mutation', () => {
...
+              error: null,
+              token: 123,
+            },
           },
         })
       )
     })
   })
-
-  describe('when graphql query returns a 401 error', () => {
```
suejung-sentry, line 431: do we no longer need this test?

suejung-sentry (approved):

src/shared/api/api.test.ts
nicholas-codecov: Ahh, if you read a couple lines below, it's actually checking for a 403 error. I renamed, and moved this test to a more suitable location.

nicholas-codecov: Line 230 on the right.

suejung-sentry: ah gotcha - thanks sounds good!!

PR #3287, opened 2024-09-24
feat: Tokenless banner for admin/member views

RulaKhaled:
# Description
Also no API work for tokenless. This is showing tokenless banner if banner is required for uploads.

# Notable Changes
- New banners, hidden behind a feature flag 
- Tests 

# Screenshots
https://github.com/user-attachments/assets/1ca9dc07-fb3c-42a9-b05f-2dfdea3cfd9f

Issue: [[Gazebo] Create token banners for both admin and user](https://github.com/codecov/engineering-team/issues/2514)
Design: https://www.figma.com/design/CRM2ODN4ihBWITjC3xryF1/GH-2100?node-id=2001-11&node-type=frame&t=7OQWaHkGOjqDO8op-0 

# Legal Boilerplate


src/shared/GlobalTopBanners/GlobalTopBanners.tsx
```diff
@@ -8,18 +8,20 @@ const TrialBanner = lazy(() => import('./TrialBanner'))
 const TeamPlanFeedbackBanner = lazy(() => import('./TeamPlanFeedbackBanner'))
 const ProPlanFeedbackBanner = lazy(() => import('./ProPlanFeedbackBanner'))
 const OktaBanners = lazy(() => import('./OktaBanners'))
+const TokenlessBanner = lazy(() => import('./TokenlessBanner'))
 
 const GlobalTopBanners: React.FC = () => {
   return (
     <SilentNetworkErrorWrapper>
       {/* These are listed in priority order: if multiple banners are rendering, only the bottommost will display. */}
-      <div className="[&>*:last-child]:block [&>*]:hidden">
```
ajay-sentry, line 16: was this removed on purpose?

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
```diff
@@ -0,0 +1,160 @@
...
+}
+
+interface UseOrgUploadTokenParams {
+  provider: string
+  owner: string
+}
+
+function OrgUploadTokenTooltip({ orgUploadToken }: { orgUploadToken: string }) {
+  const [isTokenVisible, setIsTokenVisible] = useState(true)
+  const encodedToken = orgUploadToken.replace(/\w|[^-]/g, 'x')
```
ajay-sentry, line 26: This should probably be encoded at the time of calling the hook right? Otherwise it's still available in the network tab

If the purpose is to just blankly show and hide the token with a string then I think we can use a variable with the const "xxx-xxx" to mimic this functionality without spending cycles doing the replace

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
```diff
@@ -0,0 +1,160 @@
...
+  return (
+    <TopBanner>
+      <TopBanner.Start>
+        <p className="items-center gap-1 md:flex">
+          <span className="flex items-center gap-1 font-semibold">
+            <Icon name="informationCircle" />
+            Uploading with token is now required.
+          </span>
+          You must upload with{' '}
+          {isString(orgUploadToken) ? (
```
ajay-sentry, line 89: nit: I think we should try and move off of using Lodash for things like this; in this case, we're importing a dependency that is just doing a check around typeof orgUploadToken === 'string' right?

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
```diff
@@ -0,0 +1,160 @@
...
+    owner,
+  })
+
+  return (
+    <TopBanner>
+      <TopBanner.Start>
+        <p className="items-center gap-1 md:flex">
+          <span className="flex items-center gap-1 font-semibold">
+            <Icon name="informationCircle" />
+            Uploading with token is now required.
```
ajay-sentry, line 126: nit: with "A" token?

ajay-sentry: or "The" actually to stay in line with the below copy

ajay-sentry:
General question -- if the token is "encoded" should clicking the copy button still copy the actual token?

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
nicholas-codecov: +1

src/shared/GlobalTopBanners/GlobalTopBanners.tsx
RulaKhaled: yes, it was overriding other banners instead of stacking them on top of each other.

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
RulaKhaled: Nope, we don’t need to go that far. This is purely for accessibility, such as when someone is nearby and you want to ensure it's only visible to you. I think replacing it with the exact number of characters makes more sense, as that's how most 'hide password' functionality typically works. It's mimicking the functionality of the global upload token settings tab.

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
RulaKhaled: i don't mind! For future ref are we not supposed to ever use it then?

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
RulaKhaled: https://www.figma.com/design/CRM2ODN4ihBWITjC3xryF1/GH-2100?node-id=2001-11&node-type=frame&t=a6FXMXCubyJ4NJxJ-0 your suggestion makes more sense though!

RulaKhaled:
> General question -- if the token is "encoded" should clicking the copy button still copy the actual token?

yes sir. Check upload token settings tab for a reference

src/shared/GlobalTopBanners/GlobalTopBanners.tsx
ajay-sentry: ooo, can we update the comment above too then? It sounds like that may have been the intended behavior at one point?

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
ajay-sentry: Fair enough, it's not too expensive either way

ajay-sentry (approved):
looks good! Just the one comment update

src/shared/GlobalTopBanners/GlobalTopBanners.tsx
RulaKhaled: good point!

src/shared/GlobalTopBanners/TokenlessBanner/TokenlessBanner.tsx
ajay-sentry: If we can help it, that'd be my preference tbh. There are very few lodash functions that are actually "worth it"

I like omit and merge off the top of my head

PR #3288, opened 2024-09-24
chore: Update shared/GlobalBanners to Vitest

nicholas-codecov:
# Description

This PR updates all of the tests in `shared/GlobalBanners` to Vitest, and adds in a global override to the Vitest config to write to a jest object in the global space for RTL to advance fake timers automatically.


spalmurray-codecov (approved):
"Pain is the only real thing" - Copilot 2024

PR #3325, opened 2024-09-26
ref: Remove old checkbox component

spalmurray-codecov:
Removes the old checkbox component, replacing with the new one. Had to fix the one checkbox instance to make it work with the new component. Note that the new checkbox is not supposed to look *identical* to the old one, there are some subtle differences.

Before:
Screenshot 2024-09-26 at 16 59 01

After:
Screenshot 2024-09-26 at 16 57 19


ajay-sentry (approved):
looks good!

PR #3347, opened 2024-10-03
chore: Ensure that Vite build supports larger browser set

nicholas-codecov:
# Description

This PR adds in a new Vite plugin `@vitejs/plugin-legacy`, this will automatically create a two bundles of our application. One for modern browsers that support modules, etc. and one for legacy browsers. 

You can read more about this plugin [here](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy)


vite.config.mjs
```diff
@@ -65,6 +66,9 @@ export default defineConfig((config) => {
         REACT_APP_PENDO_KEY: process.env.REACT_APP_PENDO_KEY,
       }),
       tsconfigPaths(),
+      legacy({
```
suejung-sentry, line 69: Does this work - skimming the doc, we can call `legacy()` here and that will default to reading the `browserslist` lines from package.json [here](https://github.com/codecov/gazebo/blob/bd8b644fff48fc5bcdd06ed1ac1170b258b847da/package.json#L41C2-L52C5)?
May be nicer so we don't have a diverging/overridden list for this?

[doc](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy#targets)
> The [targets] query is also [Browserslist compatible](https://github.com/browserslist/browserslist). See [Browserslist Best Practices](https://github.com/browserslist/browserslist#best-practices) for more details.
> If it's not set, plugin-legacy will load [the browserslist config sources](https://github.com/browserslist/browserslist#queries) and then fallback to the default value.

disclaimer - have not tried!

suejung-sentry (approved):

vite.config.mjs
nicholas-codecov: I don't see anything around the docs about this, and reading the source code it does not look like something along these lines exists. Once we move off of CRACO we can remove it from the `package.json`

suejung-sentry: Sounds good - wonder why they had that description there in `targets` saying it does support it then? This thing

[doc](https://github.com/vitejs/vite/tree/main/packages/plugin-legacy#targets)

The [targets] query is also [Browserslist compatible](https://github.com/browserslist/browserslist). See [Browserslist Best Practices](https://github.com/browserslist/browserslist#best-practices) for more details.
If it's not set, plugin-legacy will load [the browserslist config sources](https://github.com/browserslist/browserslist#queries) and then fallback to the default value.

PR #3364, opened 2024-10-08
feat: Update FailedTests Table to use FlakeRate

ajay-sentry:
# Description

This PR aims to update the existing table query to now fetch flakeRate property and links that property to the table as well.

Additionally we update the UTs to match

# Screenshots

Screenshot 2024-10-07 at 5 26 36 PM

# Link to Sample Entry

# Legal Boilerplate


src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -129,11 +133,14 @@ const columns = [
...
     cell: (info) => {
+      const value = (info.renderValue() ?? 0) * 100
+      const isInt = Number.isInteger(info.renderValue())
+
       return (
         <Tooltip delayDuration={0} skipDelayDuration={100}>
           <Tooltip.Root>
             <Tooltip.Trigger className="underline decoration-dotted decoration-1 underline-offset-4">
-              {info.renderValue() ? `${info.renderValue()}%` : '9%'}
+              {isInt ? `${value}%` : `${value.toFixed(2)}%`}
```
nicholas-codecov, line 143: Thoughts on swapping over to using [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)? 

It's a really nifty built-in tool in all browsers for handling just this sort of thing, so you shouldn't have to do any ternaries etc.

nicholas-codecov (commented):
One comment, and one sneaky test failing ... tho I think it's actually a flake 😞

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
ajay-sentry: Looked into it a bit and it looks like the equivalent code would be something like

```
      const formatter = new Intl.NumberFormat('en-US', {
        style: 'percent',
        maximumFractionDigits: 2,
      }).format(info.renderValue() ?? 0)
```

It's a neat tool for sure, but I think the cognitive load from deconstructing this code is actually higher than what I had previously

suejung-sentry (approved):
lgtm

PR #3371, opened 2024-10-09
fix: Fix commit detail page error boundary

suejung-sentry:
# Description

Customer reported issue on Commit Detail page > Coverage report > Files changed tab > Files Changed table
Screenshot 2024-10-09 at 7 18 53 PM

The issue is that the component shows a loading spinner, then after a few seconds, the whole page is replaced with an upside down umbrella error page.
Screenshot 2024-10-09 at 7 25 05 PM

This component calls the `useCommit` hook, which calls a `useQuery` (GraphQL query `Commit`) and `useCompareTotals` (GraphQL query `CompareTotals`). The latter of these runs the GraphQL query only when `shouldPoll =
      commitQuery?.data?.commit?.compareWithParent?.state === 'pending'`.
It seems when data gets into that delicate state of the report still being `pending` (not yet `processed`), but the second hook fails for some reason (unknown why - but I see 404 with null data in the console), we end up with the error bubbling up to trigger the full page 404.
The full page 404 is because `useCompareTotals` uses our [default](https://github.com/codecov/gazebo/blob/fff945ced4afa7a03ef1dd9aa37d04fea3b0e3f9/src/index.tsx#L41) react-query v4 `suspense: true` ([doc](https://tanstack.com/query/v4/docs/framework/react/guides/suspense)) which relies on React error boundaries for any thrown errors. The next up error boundary for this component is all the way up the stack with the full page umbrella error page. In this PR I introduced a local boundary with some error text that looks like below to render instead.

I also populate the data returned with the 404 so we can see what the cause of the error is for that zod schema violation. That can inform if we need any follow-up work to avoid this state. I had a lot of trouble getting anything to replicate this state other than the customer account, so the true test will be once we can gather more data about what's causing the failure there (my suspicion is something related to longer processing times than we see in our repos). The actual error causing this appears transient as I'm not able to consistently get commits to that state and even ones that I do see, don't stay that way across a page refresh. Anyway in the meantime, this new error state is less jarring than the full page upside down umbrella one and gives us more information about any underlying errors.

Here's a picture of the new error state (and if you open console, we can grab more logs now I hope)
Screenshot 2024-10-10 at 12 49 34 AM

Closes https://github.com/codecov/internal-issues/issues/825

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,11 +111,11 @@ Mocks are located in `src/mocks/handlers`
 Example:
 
 ```js
-import { rest } from 'msw'
+import { http, HttpResponse } from 'msw2'
 
 export const handlers = [
-  rest.get('internal/user', (req, res, ctx) => {
-    return res(ctx.status(200), ctx.json({ mock: 'data' }))
+  http.get('internal/user', (req, res, ctx) => {
+    return HttpResponse.json({ data: { mock: 'data' } })
   }),
 ]
 ```
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -47,7 +47,7 @@ RUN apk -U add git
 COPY . /home/workspace
 RUN corepack enable
 RUN yarn install
-RUN yarn build && rm -f build/mockServiceWorker.js
+RUN yarn build:vite && rm -f build/mockServiceWorker.js
 
 
 FROM alpine:3.17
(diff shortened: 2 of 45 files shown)


spalmurray-codecov (approved):

suejung-sentry:
[This follow-up PR](https://github.com/codecov/gazebo/pull/3422) fixes the cause of this issue - the `useCompareTotals` hook doesn't run often (only polls under certain circumstances) and appears to potentially have been always failing due to mixup across the `coverage` vs. `percentCovered` keys.

Also there was an issue in ImpactedFiles should be nullable and zod schema adjusted as such

The ErrorBoundary was removed in favor of a more holistic pattern in [this PR](https://github.com/codecov/gazebo/pull/3423).

PR #3380, opened 2024-10-10
chore: Move Dockerfile over to Vite

nicholas-codecov:
# Description

This PR moves the Dockerfile over to Vite, and as well updates the inline replacement string so that we can ensure that we replace the given strings in the correct directory.

Built and ran the docker image locally and it was all happy (minus API requests 😅)


spalmurray-codecov (approved):
lgtm, I'll try this out with the rest of the selfhosted images when I get to the office (internet on train no good for downloading images lol)

trent-codecov (approved):

PR #3384, opened 2024-10-10
feat: Glue in pass/fail/skip params into tooltip via test results query

ajay-sentry:
# Description

This PR glues in the pass fail and skip counts from the test results query to the tooltip shown when you hover the flake rate field.
A minor refactor had to be done when generating the test results "values" to show on the table because multiple values from the query are being populated onto a single column.

I basically just moved over that component logic to the column and had the column render the value itself instead of having it on the column.

# Screenshots

Screenshot 2024-10-10 at 4 52 37 PM

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useInfiniteTestResults/useInfiniteTestResults.tsx
+++ b/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useInfiniteTestResults/useInfiniteTestResults.tsx
@@ -21,6 +21,9 @@ const TestResultSchema = z.object({
   failureRate: z.number().nullable(),
   flakeRate: z.number().nullable(),
   avgDuration: z.number().nullable(),
+  totalFailCount: z.number(),
+  totalSkipCount: z.number(),
+  totalPassCount: z.number(),
 })
 
 export const OrderingDirection = {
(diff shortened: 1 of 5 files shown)


src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useInfiniteTestResults/useInfiniteTestResults.tsx
```diff
@@ -21,6 +21,9 @@ const TestResultSchema = z.object({
   failureRate: z.number().nullable(),
   flakeRate: z.number().nullable(),
   avgDuration: z.number().nullable(),
+  totalFailCount: z.number(),
```
RulaKhaled, line 24: Just to confirm, these cannot be null right?

RulaKhaled (commented):
let's goo

RulaKhaled (approved):

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/hooks/useInfiniteTestResults/useInfiniteTestResults.tsx
ajay-sentry: that's right!

PR #3398, opened 2024-10-16
chore: Rename msw2 to msw

nicholas-codecov:
# Description

This PR updates the imports for `msw2` over to `msw` since we have fully moved to Vitest and MSW 2 so we can just use the normal pattern again.


ajay-sentry (approved):
lgtm ✅

PR #3399, opened 2024-10-16
feat: Change cancellation confirmation banner and handle auto-refund …

calvin-codecov:
…cancel

# Description

Closes https://github.com/codecov/engineering-team/issues/2507
Figma: https://www.figma.com/design/WispRoZQs6rGwy0nrwN9xB/GH-1507?node-id=1-2&t=AnqFQfPjam4O0AF8-1

Copy of the banner and the "alert" component has changed.

Related to https://github.com/codecov/codecov-api/pull/894

# Code Example

# Notable Changes

In order to be able to figure out that we have issued a Stripe auto-refund and immediately canceled a subscription, we are utilizing Plan Page context while also checking that there is no `subscriptionDetail` populated, also known as being on the free plan. Without the context, we don't know when a cancellation has been fired because we have no signal that a cancellation has already happened and without checking the non-existence of subscriptionDetail, we don't know that it is not a "cancel at period end" cancellation.

# Screenshots

Updating seat count:
Update_seats

Cancel at period end outside of grace period: (disregard width: it will be inline with the cards like the screenshot below)
cancel_period_end

Cancel immediately with refund:
Screenshot 2024-10-16 at 3 04 10 PM

# Link to Sample Entry

# Legal Boilerplate


src/pages/PlanPage/subRoutes/CurrentOrgPlan/InfoMessageCancellation/InfoMessageCancellation.tsx
```diff
@@ -14,19 +14,26 @@ function getPeriodEnd(
...
-        <p className="text-sm">
-          Your subscription has been cancelled and will become inactive on{' '}
-          {periodEnd}
-        </p>
-      </Message>
+    <div className="col-start-1 col-end-13 mb-4 sm:flex-initial md:w-2/3 lg:w-3/4">
+      <Alert variant="info">
+        <Alert.Title className="text-sm">Cancellation confirmation</Alert.Title>
+        <Alert.Description className="text-sm">
+          Your subscription has been successfully canceled. Your account{' '}
```
ajay-sentry, line 29: cancelled vs. canceled -> only calling it out cuz the banner has 2 l's :P

src/pages/PlanPage/subRoutes/CurrentOrgPlan/CurrentOrgPlan.tsx
```diff
@@ -47,8 +47,10 @@ function CurrentOrgPlan() {
 
   return (
     <div className="w-full lg:w-4/5">
+      {/* subscriptionDetail populated means possible cancel at period end cancellation while
+      planUpdatedNotification.isCancellation signifies an auto-refund cancellation */}
       {accountDetails?.subscriptionDetail ||
-      planUpdatedNotification.isRefundedCancellation ? (
+      planUpdatedNotification.isCancellation ? (
```
ajay-sentry, line 53: Im wondering if cancellation or auto refund is the more important aspect of this variable name

ajay-sentry: The thought comes from the variable "isCancellation" only being true for a very small percentage of total cancellations

ajay-sentry (approved):
changes look good! Just some small stuff

src/pages/PlanPage/subRoutes/CurrentOrgPlan/CurrentOrgPlan.tsx
calvin-codecov: isCancellation is always true when we want to show this component regardless of `accountDetails?.subscriptionDetail`'s value so will be removing that check

src/pages/PlanPage/subRoutes/CurrentOrgPlan/InfoMessageCancellation/InfoMessageCancellation.tsx
calvin-codecov: apparently both are correct and `canceled` is prefered in American English? I've always been more drawn to cancelled though until I googled this haha. I'm gonna change it back 🤷

PR #3401, opened 2024-10-16
fix: Separate header render from main app

suejung-sentry:
Our main BaseLayout has an ErrorBoundary and Suspense wrapper for all the react-query suspense: true calls in all the components children of our main app. If there is an error in any of those, it bubbles up the tree and displays the specified error fallback (e.g., upside down umbrella) instead of anything in that tree. 
So in this PR, moving over the Header and Global Top Banners to their own separate suspsense/errorBoundary wrapper, so that tree renders where possible, independent of any error in the main app tree.

Closes https://github.com/codecov/engineering-team/issues/2460


src/layouts/shared/ErrorBoundary/ErrorBoundary.test.tsx
```diff
@@ -1,20 +1,20 @@
 import * as Sentry from '@sentry/browser'
 import { render, screen } from '@testing-library/react'
 import { MemoryRouter, Route } from 'react-router-dom'
-import { vi } from 'vitest'
```
suejung-sentry, line 4: the changes in this file are to convert to TS

spalmurray-codecov: Nice!

spalmurray-codecov (approved):
looks good!

PR #3409, opened 2024-10-18
feat: Adding table header to TA 

RulaKhaled:
# Description
Adding search and reset to default functionality to TA tab 

# Notable Changes
- New header for the table 
- Fixed some default query params 

# Screenshots

https://github.com/user-attachments/assets/26c1fede-3258-46bd-9be2-78c31ba25be7

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/SelectorSection.tsx
+++ b/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/SelectorSection.tsx
@@ -58,7 +58,8 @@ function SelectorSection() {
     if (!params?.testSuites || params.testSuites.length === 0) {
       setSelectedTestSuites([])
     }
-  }, [params])
+    // @ts-expect-error need to type out useLocationParams
+  }, [params?.flags, params?.testSuites])
 
   // This is just here for now to appease linter til we link it up
   console.log(testSuiteSearch, flagSearch)
--- a/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/SelectorSection.tsx
+++ b/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/SelectorSection.tsx
@@ -110,7 +110,11 @@ function SelectorSection() {
                 renderSelected={({ label }: { label: string }) => label}
               />
             </div>
-            <A to="" isExternal hook={'60-day-retention'}>
+            <A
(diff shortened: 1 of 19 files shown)


src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/SelectorSection.tsx
```diff
@@ -15,36 +16,51 @@ import Select from 'ui/Select'
...
   // @ts-expect-error need to type out useLocationParams
-  const [selectedFlags, setSelectedFlags] = useState(params?.flags)
-
+  const [selectedFlags, setSelectedFlags] = useState(params?.flags || [])
   const [flagSearch, setFlagSearch] = useState('')
 
   const [selectedTestSuites, setSelectedTestSuites] = useState(
     // @ts-expect-error need to type out useLocationParams
-    params?.testSuites
+    params?.testSuites || []
```
ajay-sentry, line 48: do we still need these || [] if the default query param is already []? 

Similar story for flags and the useEffect below

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/TableHeader/TableHeader.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  const isParamsDefault =
+    JSON.stringify(params) === JSON.stringify(defaultQueryParams)
+
+  return (
+    <>
+      <hr />
+      <div className="flex items-center justify-between py-2">
+        <h3 className="font-semibold">Tests (50K)</h3>
```
ajay-sentry, line 26: I think this should be h2 looking at the preview deploy

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/SelectorSection.tsx
RulaKhaled: We don't need it for useState, but we do need to update params in useEffect. 

When you call updateParams(defaultQueryParams), the params object updates but this might not automatically trigger a state update for selectedTestSuites/Flags because React doesn't directly tie state variables to external objects like params unless you explicitly update the state:


https://github.com/user-attachments/assets/c71c1a75-5fce-448c-92c0-9239cceba27e

ajay-sentry (approved):

PR #3410, opened 2024-10-18
fix: Move from commit SHA to 22 char hash and polyfill Promise.allSettled

nicholas-codecov:
# Description

Couple of Vite config changes here and a plugin version bump. Goal of this is to resolve some issues with caching, browsers not supporting `Promise.allSettled`, and bumping a plugin version so it works properly with the extended hash values.

Will hopefully resolve: https://codecov.sentry.io/issues/5968807574/events/3a931390316b4d78a3de37186f8306d1/

# Notable Changes

- Polyfill `Promise.allSettled`
- Update Codecov Vite plugin
- Adjust output file names


ajay-sentry (approved):
looks good

PR #3411, opened 2024-10-18
feat: Add CI status tooltip

calvin-codecov:
# Description

Closes https://github.com/codecov/engineering-team/issues/2523

Discussed with Adalene and we decided to do away with the border style to match with the tooltips on their way for TA.

# Code Example

# Notable Changes

# Screenshots
Screenshot 2024-10-22 at 1 42 31 PM

Looked into the reason why we had z-30 in SummaryDropdown and it looks like we don't currently need it.
Screenshot 2024-10-23 at 9 28 28 AM
Screenshot 2024-10-23 at 9 28 12 AM

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/services/navigation/useNavLinks/useStaticNavLinks.js
+++ b/src/services/navigation/useNavLinks/useStaticNavLinks.js
@@ -506,5 +506,12 @@ export function useStaticNavLinks() {
       isExternalLink: true,
       openNewTab: true,
     },
+    requireCIPassDocs: {
+      text: 'Codecov YAML require CI to pass',
+      path: () =>
+        'https://docs.codecov.com/docs/codecovyml-reference#codecovrequire_ci_to_pass',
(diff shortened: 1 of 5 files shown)


src/services/navigation/useNavLinks/useStaticNavLinks.js
```diff
@@ -506,5 +506,12 @@ export function useStaticNavLinks() {
       isExternalLink: true,
       openNewTab: true,
     },
+    requireCIPassDocs: {
```
ajay-sentry, line 509: could you add this to the useStaticNavLinks.test file too?

ajay-sentry: a test assertion I mean

ajay-sentry (approved):

PR #3413, opened 2024-10-21
fix: Truncate sunburst chart's breadcrumbs

suejung-sentry:
Fix bug with sunburst's breadcrumbs show a "..." if too long. It always display just the last element with "..." if it goes more than 1 deep as it seemed like in most cases only 1 fits on the page. This is a quick and dirty solution. For real we should probably give more real estate for this rich information or make a truncatable Breadcrumbs reusable component that has a resize observer.

Closes https://github.com/codecov/engineering-team/issues/2337

Example screenshots:
Screenshot 2024-10-21 at 5 32 02 PM
Screenshot 2024-10-21 at 5 31 51 PM
Screenshot 2024-10-21 at 5 31 45 PM

here are the befores which just need a bandaid:
Screenshot 2024-10-21 at 5 34 19 PM
Screenshot 2024-10-21 at 5 33 38 PM

### DIFF
--- a/src/pages/RepoPage/CoverageTab/OverviewTab/subroute/Sunburst/Sunburst.jsx
+++ b/src/pages/RepoPage/CoverageTab/OverviewTab/subroute/Sunburst/Sunburst.jsx
@@ -56,4 +56,12 @@ function Sunburst() {
   )
 }
 
+export function getPathsToDisplay(breadcrumbPaths) {
+  if (breadcrumbPaths.length <= 1) {
+    return breadcrumbPaths
+  } else {
+    return [breadcrumbPaths[0], { text: '...' }]
+  }
+}
+
 export default Sunburst
--- a/src/pages/RepoPage/CoverageTab/OverviewTab/subroute/Sunburst/Sunburst.jsx
+++ b/src/pages/RepoPage/CoverageTab/OverviewTab/subroute/Sunburst/Sunburst.jsx
@@ -23,6 +23,13 @@ function Sunburst() {
 
   const breadcrumbPaths = useConvertD3ToBreadcrumbs(currentPath)
 
+  let pathsToDisplay
+  if (breadcrumbPaths.length <= 1) {
+    pathsToDisplay = breadcrumbPaths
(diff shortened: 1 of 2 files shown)


src/pages/RepoPage/CoverageTab/OverviewTab/subroute/Sunburst/Sunburst.jsx
```diff
@@ -23,6 +23,13 @@ function Sunburst() {
 
   const breadcrumbPaths = useConvertD3ToBreadcrumbs(currentPath)
 
+  let pathsToDisplay
+  if (breadcrumbPaths.length <= 1) {
+    pathsToDisplay = breadcrumbPaths
+  } else {
+    pathsToDisplay = [breadcrumbPaths[0], { text: '...' }]
```
nicholas-codecov, line 30: Can we extend this to more than just one level? Something like 2 or 3 so it matches the depth you can travel while looking at the sunburst? I find it pretty difficult to get any context as to "where" i am in the file structure with just the current level.

nicholas-codecov: Could also poke the design folks, and get their input on this.

suejung-sentry: Thanks, I'll ping design. This is just a quick and dirty. Really we have too much information trying to be displayed in too little of space. I added some screenshots to the PR description to show what it's like. If it's any consolation, there's the hover tooltip on the sunburst - I can add one to the breadcrumbs as well perhaps.
If we want to do this for real a reusable truncated breadcrumbs component would be ideal imo

suejung-sentry: here's a note from talking to adalene - sounds like we can get some design love on this in general to make it better for users
Screenshot 2024-10-22 at 1 42 13 PM

PR #3415, opened 2024-10-21
feat: Add TS Query v5 to Gazebo

nicholas-codecov:
# Description

This PR adds in TS Query V5 as a new dependency and adds in the new query client and provider to the app so the transition to V5 can happen smoothly.

Issue: codecov/engineering-team#2128

# Notable Changes

- Add TS Query Dep
- Add client and provider to app


spalmurray-codecov (approved):
Yeehaw 🤠

PR #3422, opened 2024-10-22
fix: Results array for impacted files is nullable

nicholas-codecov:
# Description

This PR fixes some zod schema issues where a field was being marked as non-nullable, however it is in fact nullable. This was most likely the root cause for: #3371. I've gone through and updated all instances of this `results` field to be nullable, and all the places where it's used. This should resolve the issue raised in #3371, so i've gone ahead and removed that error boundary so we start collecting these network errors up in the `NetworkErrorBoundary` again.

# Notable Changes

- Update zod schemas in hooks calling `impactedFiles.results` from the API to be nullable
- Update code using this field to handle if it is `null`
- Update `CoverageObjSchema` in `useCompareTotals` to be `coverage` not `percentCovered`
- Remove error boundary from `FilesChangedTab`

### DIFF
--- a/src/services/commit/useCompareTotals.tsx
+++ b/src/services/commit/useCompareTotals.tsx
@@ -18,7 +18,7 @@ import { type NetworkErrorObject } from 'shared/api/helpers'
 import A from 'ui/A'
 
 const CoverageObjSchema = z.object({
-  percentCovered: z.number().nullable(),
+  coverage: z.number().nullable(),
 })
 
 const ImpactedFileSchema = z
--- a/src/services/commit/useCompareTotals.tsx
+++ b/src/services/commit/useCompareTotals.tsx
@@ -14,6 +14,7 @@ import {
   RepoOwnerNotActivatedErrorSchema,
 } from 'services/repo/schemas'
 import Api from 'shared/api'
+import { type NetworkErrorObject } from 'shared/api/helpers'
 import A from 'ui/A'
 
 const CoverageObjSchema = z.object({
@@ -32,7 +33,7 @@ const ImpactedFileSchema = z
 const ImpactedFilesSchema = z.discriminatedUnion('__typename', [
   z.object({
     __typename: z.literal('ImpactedFiles'),
-    results: z.array(ImpactedFileSchema),
+    results: z.array(ImpactedFileSchema).nullable(),
   }),
   z.object({
     __typename: z.literal('UnknownFlags'),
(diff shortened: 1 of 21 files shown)


src/services/pull/usePullCompareTotalsTeam.tsx
```diff
@@ -23,14 +24,16 @@ const CoverageObjSchema = z.object({
...
+    results: z
+      .array(
+        z.object({
+          headName: z.string().nullable(),
+          missesCount: z.number(),
+          isCriticalFile: z.boolean(),
+          patchCoverage: CoverageObjSchema,
+        })
+      )
+      .nullable(),
```
suejung-sentry, line 36: Ah, I see that `ImpactedFiles` object is nullable and that was missing across all the usages. It looks like it is marked as nullable on the `api` side as well so there may have been drift across the repos

src/services/commit/useCompareTotals.tsx
```diff
@@ -14,10 +14,11 @@ import {
   RepoOwnerNotActivatedErrorSchema,
 } from 'services/repo/schemas'
 import Api from 'shared/api'
+import { type NetworkErrorObject } from 'shared/api/helpers'
 import A from 'ui/A'
 
 const CoverageObjSchema = z.object({
-  percentCovered: z.number().nullable(),
+  coverage: z.number().nullable(),
```
suejung-sentry, line 21: On the api side it seems like these fields (`percentCovered`, `coverage`) return the same data and `percentCovered` is preferred. Are we seeing data returned with one but not the other? Because that would be a good thing for us to fix if so

https://github.com/codecov/codecov-api/blob/324fd728cd905c073e96c6657ac3d363a92c6070/graphql_api/types/coverage_totals/coverage_totals.py#L5

https://github.com/codecov/codecov-api/blob/324fd728cd905c073e96c6657ac3d363a92c6070/graphql_api/types/coverage_totals/coverage_totals.graphql#L10

suejung-sentry (approved):
thanks for looking over that

src/services/commit/useCompareTotals.tsx
nicholas-codecov: `coverage` is deprecated, however because that would require a lot of FE changes, the simple solution that was implemented was to alias the fields.

```graphql
coverage: Float @deprecated(reason: "Use `percentCovered`")
```

suejung-sentry: Got it, thanks! I missed that there was a mismatch somewhere in the flow causing this problem.

PR #3423, opened 2024-10-22
feat: Capture certain network errors to Sentry

nicholas-codecov:
# Description

This PR updates the `NetworkErrorBoundary` so that certain errors will be sent to Sentry so we can keep better track of them. Typically this will only be related to zod parsing errors as they're ones that we control ourselves, and it probably a good idea that we keep track of them and resolve them when able.  

# Example Code

```tsx
const parsedData = RequestSchema.safeParse(res?.data)

if (!parsedData.success) {
  return Promise.reject({
    status: 404,
    data: {},
    dev: 'useBundleSummary - 404 Failed to parse data',
    // including the error here will trigger Sentry to capture the message and create an issue
    error: parsedData.error,
  } satisfies NetworkErrorObject)
}

const data = parsedData.data

if (data?.owner?.repository?.__typename === 'NotFoundError') {
  // No error included here, so no message will be sent to Sentry
  return Promise.reject({
    status: 404,
    data: {},
    dev: 'useBundleSummary - 404 Not found error',
  } satisfies NetworkErrorObject)
}
```

# Notable Changes

- Remove Sentry metrics related to network errors
- Update `NetworkErrorBoundary` to send certain errors to Sentry
- Update `NetworkErrorObject` to have conditional `error` field
- Update tests


src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
```diff
@@ -174,12 +171,26 @@ class NetworkErrorBoundary extends Component {
...
+            level: 'error',
+            data: error.error,
+          },
+        })
+      }
       return { hasNetworkError: true, error }
     }
 
     if (Object.keys(graphQLErrorToUI).includes(error.__typename)) {
-      sendGraphQLErrorMetrics(error.__typename)
```
suejung-sentry, line 182: Losing out on the metrics is a bummer. Hopefully things are getting captured on the `api` prometheus side

src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
```diff
@@ -174,12 +171,26 @@ class NetworkErrorBoundary extends Component {
     // if the error is not a network error, we don't do anything and
     // another error boundary will take it from there
     if (Object.keys(errorToUI).includes(String(error.status))) {
-      sendNetworkErrorMetrics(error.status)
+      // only capture network errors if they are not a rate limit error
+      // this will typically only be schema parsing errors
+      if (error.status !== 429 && error.dev && error.error) {
```
suejung-sentry, line 176: Thoughts on making it easier for us in the case we forget to include both `error.dev` and `error.errror` and that means our errors don't get tracked? I could definitely see myself doing that. A fallback catchall fingerprint name perhaps?

suejung-sentry (approved):
This is really nice and useful, thank you for putting it together!

src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx
nicholas-codecov: When you use the `NetworkErrorObject` it will enforce you to add `dev` to the rejection object.

nicholas-codecov: The only issue with having a catchall is that we're likely to catch errors that are out of our control, such as repo not activated, etc. which we don't want to be flooding Sentry with.

PR #3426, opened 2024-10-22
chore: Add yarn test all script

calvin-codecov:
Adding an extra script shortcut since our current ones for running vitest use the --changed flag.


package.json
```diff
@@ -12,6 +12,7 @@
     "test": "vitest run --changed --config ./vitest.config.mjs",
     "test:vite": "yarn test",
     "test:watch": "vitest watch --changed --config ./vitest.config.mjs",
+    "test:all": "vitest run --config ./vitest.config.mjs",
```
suejung-sentry, line 15: Thoughts on swapping `test` to be all and `test:changed` for `--changed`?
My personal flow is `yarn test <nameOfFileImWorkingOn>` even if I don't make changes across commits

PR #3430, opened 2024-10-23
feat: Migrate bundle analysis services to Query V5

nicholas-codecov:
# Description

This PR updates the hooks/services in the `services/bundleAnalysis` directory to Query V5. Some of these updates include moving fully over to the `queryOptions` API format that we will be moving towards, I didn't do this for all of these, as the PR would have grown significantly, and can be accomplished in a follow up PR.

Another change I made in this PR was the removal of the `index.ts` file from the `services/bundleAnalysis` directory. Some of the benefits of this are:

- Removing barrel files (not good for bundling)
  - They will typically include all of the functions being exported as it struggles to properly tree shaking, especially if you're using `export * from './...'`
- Explicit imports, if you `cmd + click` on the import you're taken to the function declaration not the `index.ts` file
- Removes potential conflicting type imports with similarly named types

Closes: codecov/engineering-team#2964

# Notable Changes

- Update all services in `services/bundleAnalysis` to Query V5
- Update components using these hooks to use the newer version or use the `queryOption` API format
- Update tests
  - Some of these were just updating missed MSW mocks or incorrect MSW mocks


src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/AssetsTable.tsx
```diff
@@ -161,7 +161,7 @@ const createColumns = (totalBundleSize: number | null) => [
             />{' '}
             <ChangeOverTime
               change={value?.change?.size.uncompress}
-              hasMeasurements={value.measurements.length > 0 ?? false}
+              hasMeasurements={value.measurements.length > 0}
```
nicholas-codecov, line 164: This was something that Vite kept bringing up as an optimization so finally threw it in.

src/services/bundleAnalysis/useBranchBundleSummary.tsx
```diff
@@ -127,28 +130,20 @@ const query = `query BranchBundleSummaryData(
   }
 }`
 
-export interface UseBranchBundleSummaryArgs {
+export interface BranchBundleSummaryQueryOptsArgs {
   provider: string
   owner: string
   repo: string
-  branch?: string
+  branch: string | null | undefined
```
spalmurray-codecov, line 137: Why this instead of `branch?: string | null`?

src/services/bundleAnalysis/useBundleAssets.tsx
```diff
@@ -266,32 +271,34 @@ export const useBundleAssets = ({
           dateBefore,
           dateAfter,
           filters,
-          assetsAfter: pageParam,
+          assetsAfter,
```
spalmurray-codecov, line 274: Does this mean that `assetsAfter=null` doesn't get along well with the query? I'd think it might be worthwhile updating this API side if not. Would rather do that than do the few lines above on every infinite query.

src/services/bundleAnalysis/useBranchBundleSummary.tsx
nicholas-codecov: This is an attempt to be a bit more explicit with function args and their types. `?` represents an optional argument so TS doesn't complain if you never provide it. However, if you type it as above, TS will complain if you don't provide the value:

Screenshot 2024-11-01 at 12 17 48

spalmurray-codecov: Oh so you just want this to always be called with a branch param?

spalmurray-codecov (changes requested):
Looks good. Honestly thought the transition would require more changes, glad to see it's not so bad. Requesting changes bc I want to talk through my comments and make sure we're aligned 🫡

src/services/bundleAnalysis/useBundleAssets.tsx
nicholas-codecov: Sorry, `null` works fine, and I'm updating things rn to be set to `null` instead of `undefined`. The type for `pageParam` is inferred through the `initialPageParam` so we have to assign it to a `string` for the remaining cursors to keep the types happy.

spalmurray-codecov: If null is fine, doesn't the old code work? `assetsAfter: pageParam`

spalmurray-codecov: I may be misunderstanding though

src/services/bundleAnalysis/useBranchBundleSummary.tsx
nicholas-codecov: Yes I would like all the arguments provided, this still makes allowances however if `branch` is undefined.

spalmurray-codecov: Ok cool beans

src/services/bundleAnalysis/useBundleAssets.tsx
nicholas-codecov: Soooo just did some testing, and the API is okay if just an empty string is provided ... so I'll get rid of all of this.

spalmurray-codecov: NICE

PR #3433, opened 2024-10-23
feat: Update Jest TA onboarding copy to include filepath setting

ajay-sentry:
# Description

This PR ships a copy update to the TA onboarding page around adding a new setting specific for Jest TA setups. We now need to include this new env variable when running jest so we can capture the file path and output it correctly

Needed to basically "copy paste" all the clipboard values into a separate value since the copy may be a JSX element now, and when copying a JSX element you get [Object object] which is not what we want. Thanks @spalmurray-codecov for pointing that out!

Closes https://github.com/codecov/engineering-team/issues/2605

# Screenshots

Screenshot 2024-10-23 at 12 44 46 PM

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/RepoPage/FailedTestsTab/FrameworkTabsCard/FrameworkTabs/FrameworkTabs.tsx
+++ b/src/pages/RepoPage/FailedTestsTab/FrameworkTabsCard/FrameworkTabs/FrameworkTabs.tsx
@@ -16,8 +16,24 @@ const FrameworkCopy = {
   [Frameworks.PYTEST]:
     'pytest --cov --junitxml=junit.xml -o junit_family=legacy',
   [Frameworks.VITEST]: 'vitest --reporter=junit',
+  [Frameworks.JEST]: (
+    <p data-testid="jest-framework-copy">
+      npm i --save-dev jest-junit <br />
+      JEST_JUNIT_CLASSNAME=
+      <span className="text-codecov-code">{'"{filepath}"'}</span> jest
+      --reporters=jest-junit
+    </p>
+  ),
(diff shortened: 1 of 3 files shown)


src/pages/RepoPage/FailedTestsTab/FrameworkTabsCard/FrameworkTabs/FrameworkTabs.tsx
```diff
@@ -13,11 +13,27 @@ const Frameworks = {
...
+const FrameworkClipboard = {
   [Frameworks.PYTEST]:
     'pytest --cov --junitxml=junit.xml -o junit_family=legacy',
   [Frameworks.VITEST]: 'vitest --reporter=junit',
   [Frameworks.JEST]:
-    'npm i --save-dev jest-junit \njest --reporters=jest-junit',
+    'npm i --save-dev jest-junit \n JEST_JUNIT_CLASSNAME="{filepath}" jest --reporters=jest-junit',
+
   [Frameworks.PHP_UNIT]: './vendor/bin/phpunit --log-junit junit.xml',
 } as const
```
calvin-codecov, line 38: could we use
```
const FrameworkClipboard = {
  ...FrameworkCopy,
  [Frameworks.JEST]: 'npm i --save-dev jest-junit \n JEST_JUNIT_CLASSNAME="{filepath}" jest --reporters=jest-junit',
}
? I think it would be cleaner and clearer

ajay-sentry: I like that a lot!

calvin-codecov (approved):
beautiful

PR #3436, opened 2024-10-24
feat: Rework how we capture service errors to sentry

nicholas-codecov:
# Description

This PR changes the way that was recently introduced as to how we capture data fetching service errors, and send them to Sentry to tidy things up a little bit, and reduce what is logged out to the console. 

# Code Example

Use with a data fetching hook:

```tsx
if (!parsedRes.success) {
  return rejectNetworkErrorToSentry({
    status: 404,
    data: {},
    dev: 'useCommit - 404 failed to parse',
    error: parsedRes.error,
  })
}
```

# Notable Changes

- Create a function that handles, rejects, and sends info to Sentry for service parsing errors
- Update two examples hooks using this new function
- Create/Update tests

### DIFF
--- a/src/shared/api/helpers.ts
+++ b/src/shared/api/helpers.ts
@@ -23,6 +23,7 @@ export function rejectNetworkError(error: NetworkErrorObject) {
       scope.addBreadcrumb({
         category: 'network.error',
         level: 'error',
+        message: error.dev,
         data: error.error,
       })
       scope.setFingerprint([error.dev])
--- a/src/shared/api/helpers.ts
+++ b/src/shared/api/helpers.ts
@@ -15,7 +15,7 @@ export interface NetworkErrorObject {
   error?: Error
 }
 
-export function rejectNetworkErrorToSentry(error: NetworkErrorObject) {
+export function rejectNetworkError(error: NetworkErrorObject) {
   // only capture network errors if they are not a rate limit error
   // this will typically only be schema parsing errors
   if (error.status !== 429 && error.dev && error.error) {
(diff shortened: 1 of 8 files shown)


src/shared/api/helpers.ts
```diff
@@ -14,6 +15,28 @@ export interface NetworkErrorObject {
...
 
+export function rejectNetworkError(error: NetworkErrorObject) {
+  // only capture network errors if they are not a rate limit error
+  // this will typically only be schema parsing errors
+  if (error.status !== 429 && error.dev && error.error) {
+    Sentry.withScope((scope) => {
+      scope.addBreadcrumb({
+        category: 'network.error',
+        level: 'error',
+        data: error.error,
```
suejung-sentry, line 26: the previous [version](https://github.com/codecov/gazebo/blob/82b3d3ef75c7ff01919953fa41dbeb5b7836d104/src/layouts/shared/NetworkErrorBoundary/NetworkErrorBoundary.jsx#L183) had also included `message: error.dev` in the "breadcrumb" if we wanted to keep that in (honestly not sure where that shows up in Sentry dashboard, but based on [this](https://docs.sentry.io/product/issues/issue-details/breadcrumbs/#breadcrumb-attributes) it seems like it can with search?)

nicholas-codecov: Nice catch, I'll add that back in

suejung-sentry (approved):
the changes make sense to me - we can get any input/feedback from team on next time we meet perhaps

PR #3438, opened 2024-10-24
chore: Convert to TS - shared context, plan, propTypes, treePaths

suejung-sentry:
Convert a group of files from JS --> TS
https://github.com/codecov/engineering-team/issues/2745


src/shared/plan/ScheduledPlanDetails/ScheduledPlanDetails.test.tsx
```diff
@@ -8,7 +8,7 @@ const queryClient = new QueryClient({
   defaultOptions: { queries: { retry: false } },
 })
 
-const wrapper = ({ children }) => (
+const wrapper = ({ children }: { children: React.ReactNode }) => (
```
ajay-sentry, line 11: nit: We usually type this guy out as wrapper: React.FC<React.PropsWithChildren>

src/shared/plan/ScheduledPlanDetails/ScheduledPlanDetails.tsx
```diff
@@ -1,12 +1,23 @@
 import { format, fromUnixTime } from 'date-fns'
-import PropType from 'prop-types'
 
-export function getScheduleStart(scheduledPhase) {
+export function getScheduleStart(scheduledPhase: {
```
ajay-sentry, line 3: thoughts on turning this into a shared interface for these two functions?

src/shared/treePaths/useTreePaths.test.tsx
```diff
@@ -70,7 +71,11 @@ const overviewMock = {
 }
 
 describe('useTreePaths', () => {
-  function setup({ repoOverviewData }) {
+  function setup({
+    repoOverviewData,
+  }: {
+    repoOverviewData: typeof overviewMock
```
ajay-sentry, line 77: this would actually be the type of the zod schema right? 
https://github.com/codecov/gazebo/blob/68a1e3a485c8c8b97783e8b1a15747f9f4dc0fc1/src/services/repo/useRepoOverview.tsx#L9

Also if the mock isn't changing, could modify line 81 to just use the mock directly and remove the need to pass it in outright

src/shared/treePaths/useTreePaths.ts
```diff
@@ -1,42 +1,48 @@
...
-  } = useParams()
+  } = useParams<URLParams>()
 
-  const { data: repoOverview } = useRepoOverview(
-    {
-      provider,
-      repo,
-      owner,
-    },
-    { suspense: false }
```
ajay-sentry, line 38: was this removed on purpose?

src/shared/plan/ScheduledPlanDetails/ScheduledPlanDetails.test.tsx
suejung-sentry: Ah thanks! I followed the wrong pattern - fixed it in other files I saw using the wrong way, too, so others don't mistake that too!

src/shared/plan/ScheduledPlanDetails/ScheduledPlanDetails.tsx
suejung-sentry: Yeah, I like that - fixed!

src/shared/treePaths/useTreePaths.ts
suejung-sentry: Yeah, `useRepoOverview` doesn't take that argument (and can't find where it ever did?) So removing this would be a no-op

src/shared/treePaths/useTreePaths.test.tsx
```diff
@@ -18,7 +22,7 @@ const queryClient = new QueryClient({
 
 const wrapper =
   (initialEntries = '/gh/owner/coolrepo/tree/main/src%2Ftests') =>
-  ({ children }) => (
+  ({ children }: { children: ReactNode }) => (
```
suejung-sentry, line 25: left this one as I thought this notation was a bit clearer

src/shared/treePaths/useTreePaths.ts
ajay-sentry: Got it! Nice fix :)

ajay-sentry (approved):
lgtm!

PR #3446, opened 2024-10-29
fix: Scrolling issues with virtual renderers

nicholas-codecov:
# Description

This PR fixes some issues with the scrolling on the virtual renderers. Basically the issue was that on the diff renderer the last line was being covered up by the scroll bar so you're unable to see it (this is common in virtualized things like this). The solution was to create our own scroll bar that comes after the virtualized content and hook it up to the scrolling behaviour.

I've also tweaked the hit counters slightly, giving them a little extra breathing room.

# Notable Changes

- Create `useIsOverflowing` hook to determine if we need our custom scroll bar
- Create `useSyncScrollLeft` to sync the scroll left value between a base ref, and an array of refs to update when the base ref is scrolled
- Update styles to remove scroll bars from the `textarea` and virtualized elements
- Create a custom scroll bar component
- Give the hit counters a bit more breathing room when stacked 

# Screenshots

Left side hit counter:

Screenshot 2024-10-29 at 12 25 21

Scrolling example on diff viewer:

https://github.com/user-attachments/assets/3f0623f2-cb5a-4204-911b-6a7bdddf07df

Scrolling example on file viewer:

https://github.com/user-attachments/assets/7171b786-197d-45d1-9766-3c938599eb66


src/ui/VirtualRenderers/VirtualFileRenderer.tsx
```diff
@@ -281,16 +319,19 @@ function VirtualFileRendererComponent({
             fileType={fileType}
             coverage={coverage}
             codeDisplayOverlayRef={codeDisplayOverlayRef}
+            wrapperWidth={wrapperWidth}
+            setWrapperRefState={setWrapperRefState}
           />
         </div>
       </div>
+      {isOverflowing ? (
+        <ScrollBar scrollBarRef={scrollBarRef} wrapperWidth={wrapperWidth} />
```
spalmurray-codecov, line 328: Could we get test for this line in both files?

src/ui/VirtualRenderers/VirtualFileRenderer.tsx
```diff
@@ -225,9 +242,28 @@ function VirtualFileRendererComponent({
...
+  const isOverflowing = useIsOverflowing(codeDisplayOverlayRef)
+
+  // sync the scroll position of the text area with the code display overlay and scroll bar
+  useSyncScrollLeft({
+    scrollingRef: textAreaRef,
+    refsToSync: [codeDisplayOverlayRef, scrollBarRef],
+  })
+  useSyncScrollLeft({
+    scrollingRef: scrollBarRef,
+    refsToSync: [codeDisplayOverlayRef, textAreaRef],
```
spalmurray-codecov, line 265: This implies it's not possible to scroll the codeDisplayOverlayRef manually, right?

src/ui/VirtualRenderers/VirtualFileRenderer.tsx
```diff
@@ -225,9 +242,28 @@ function VirtualFileRendererComponent({
...
-  useLeftScrollSync({ textAreaRef, overlayRef: codeDisplayOverlayRef })
+
+  // sync the width of the wrapper with the width of the text area
+  useSyncTotalWidth({ textAreaRef, widthDivRef })
+
+  // check if the code display overlay is overflowing, so we can conditionally render the scroll bar
+  const isOverflowing = useIsOverflowing(codeDisplayOverlayRef)
+
+  // sync the scroll position of the text area with the code display overlay and scroll bar
+  useSyncScrollLeft({
```
spalmurray-codecov, line 259: I found the name of this hook confusing until I read the definition. Didn't know that `scrollLeft` was an element attribute. IMO `useSyncScroll` could be more clear, even though it's less specific. Nit though, your call.

src/ui/VirtualRenderers/VirtualFileRenderer.tsx
nicholas-codecov: Yes, the `textarea` overlays it, so that's what the user interacts with.

spalmurray-codecov: Cool ok! just wanted to sanity check my understanding tyty

nicholas-codecov: They never actually "touch" the virtualized/tokenized code.

src/ui/VirtualRenderers/VirtualFileRenderer.tsx
nicholas-codecov: Gonna go with `useScrollLeftSync` so it reads a bit better, but I personally prefer the slightly more specific naming for clarity.

PR #3456, opened 2024-10-30
feat: Default color mode to user's current system theme

ajay-sentry:
# Description

This PR aims to add logic to default the user's appearance mode to whatever their current system theme is at the time they come to Codecov.

Originally I wanted to have the user's system theme be swappable, but I was running into some funky behavior with the instances where we had the "dark:" class name not being picked up correctly. In the end, I feel like that'd be a cute little easter egg at best, and just having the default system theme setup for when the user first comes onto the site is probably a nice Quality of Life adjustment anyway.

Worst case they just push the light/dark mode button 😄 

Closes https://github.com/codecov/engineering-team/issues/2403

# Notable Changes

- Uses the "prefers-color-scheme" media query to default the user's appearance on their first visit to Codecov

# Screenshots

https://github.com/user-attachments/assets/b78dfd2e-bacc-495f-b6cb-38d0ea9c33ff

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/globals.css
+++ b/src/globals.css
@@ -253,6 +253,8 @@
     --color-code-operator: 219 178, 89;
   }
 
+  /* Also used for prefers-color-scheme: light */
+  /* explicitly define it below if we want to change behavior */
   body {
     @apply lightMode;
   }
@@ -261,7 +263,7 @@
     @apply darkMode;
   }
 
-  /* new body theme can be applied here
+  /* new body theme can be applied here */
 
   @media (prefers-color-scheme: dark) {
     body {
@@ -271,9 +273,7 @@
     body.light {
       @apply lightMode;
     }
-
-     new body theme added above should be applied here as well to cover specificity
-  } */
+  }
 }
 
 #sentry-feedback {
--- a/src/shared/ThemeContext/ThemeContext.tsx
+++ b/src/shared/ThemeContext/ThemeContext.tsx
@@ -29,8 +29,15 @@ interface ThemeContextProviderProps {
 export const ThemeContextProvider: FC<ThemeContextProviderProps> = ({
   children,
 }) => {
-  const currentTheme = (localStorage.getItem('theme') as Theme) || Theme.LIGHT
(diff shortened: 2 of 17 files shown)


src/globals.css
```diff
@@ -261,7 +263,7 @@
     @apply darkMode;
   }
 
-  /* new body theme can be applied here
+  /* new body theme can be applied here */
 
   @media (prefers-color-scheme: dark) {
```
ajay-sentry, line 268: This is uncommented now

src/shared/ThemeContext/ThemeContext.tsx
```diff
@@ -29,8 +29,15 @@ interface ThemeContextProviderProps {
 export const ThemeContextProvider: FC<ThemeContextProviderProps> = ({
   children,
 }) => {
-  const currentTheme = (localStorage.getItem('theme') as Theme) || Theme.LIGHT
-  const [theme, setTheme] = useState<Theme>(currentTheme)
+  const prefersDark = window.matchMedia('(prefers-color-scheme: dark)').matches
```
ajay-sentry, line 32: This is the only logic changes

nicholas-codecov (commented):
Instead of mocking this out in all of these tests, can you mock it out in the `vitest.config.mjs` similar to how we're globally mocking out some Sentry stuff?

src/shared/ThemeContext/ThemeContext.tsx
ajay-sentry: I did consider that approach, though I ended up on preferring to have all the theme logic in one place

ajay-sentry:
> Instead of mocking this out in all of these tests, can you mock it out in the `vitest.config.mjs` similar to how we're globally mocking out some Sentry stuff?

For sure, updated in [5a47a97](https://github.com/codecov/gazebo/pull/3456/commits/5a47a97daa3f51e32ff2192d686a0afa7fb779cf)

PR #3472, opened 2024-11-07
fix: Responsiveness tweaks for TA Selector section

ajay-sentry:
# Description

This PR aims to fix some responsiveness issues that occurred when on some smaller screen sizes on the Test Analytics page.

# Screenshots

https://github.com/user-attachments/assets/caa264d1-be0f-4e4b-83e6-1965ad6503a1

Screenshot 2024-11-08 at 9 33 23 AM

# Link to Sample Entry

# Legal Boilerplate


ajay-sentry:
> Hrmrm, some of these test path names are pushing the select box out far enough that it's pushing width too far breaking things, i wonder if vertically stacking the selects could help a bit with this 🤔
> 
> Screenshot 2024-11-07 at 15 40 36

If we want to loop in design for a Mobile iteration that could be a worthwhile pursuit after we go live. For now I think we can treat that as a nit

nicholas-codecov (approved):

PR #3488, opened 2024-11-12
chore: Upgrade http-proxy-middleware to ^2.0.7

calvin-codecov:
# Description

Closes https://github.com/codecov/internal-issues/issues/971

Upgraded our own use of the package for
├─ gazebo@workspace:.
│  └─ http-proxy-middleware@npm:2.0.6 [6563e] (via npm:^2.0.6 [6563e])

but adding a resolution as it is also referenced here
gazebo@workspace:.
   └─ react-scripts@npm:5.0.1 [6563e] (via npm:^5.0.1 [6563e])
      └─ webpack-dev-server@npm:4.15.2 [37b33] (via npm:^4.6.0 [37b33])
         └─ http-proxy-middleware@npm:2.0.6 [edb85] (via npm:^2.0.3 [edb85])

We can remove this resolution when we pull out react-scripts as part of Gazebo V2 work tickets.

### DIFF
--- a/package.json
+++ b/package.json
@@ -176,8 +176,7 @@
     "micromatch": "^4.0.8",
     "express": "^4.21.0",
     "path-to-regexp@^1.7.0": "^1.9.0",
-    "path-to-regexp@^6.2.0": "^6.3.0",
-    "http-proxy-middleware": "^2.0.7"
+    "path-to-regexp@^6.2.0": "^6.3.0"
   },
   "packageManager": "yarn@4.5.0",
   "engineStrict": false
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -176,7 +176,8 @@
     "micromatch": "^4.0.8",
     "express": "^4.21.0",
     "path-to-regexp@^1.7.0": "^1.9.0",
-    "path-to-regexp@^6.2.0": "^6.3.0"
+    "path-to-regexp@^6.2.0": "^6.3.0",
+    "http-proxy-middleware": "^2.0.7"
```
ajay-sentry, line 180: do we still need this if we're also updating it above?

calvin-codecov: ahh sorry, I should have clarified in the PR comment that we also use the pckage. The update above was for our own use but we need the resolution for react-scripts since we can't upgrade that.

spalmurray-codecov (approved):

ajay-sentry (approved):

PR #3492, opened 2024-11-13
fix: Fix link to custom gh app for self-hosted

suejung-sentry:
Self-hosted deployments (`config.IS_SELF_HOSTED`, for Dedicated Enterprise Cloud customers and self-hosted) cannot use the default GitHub codecov app that can be installed from this link: `https://github.com/apps/codecov/installations/select_target`

Instead, that slug `codecov` needs to be replaced by the custom GitHub app name as explained in our docs [here](https://docs.codecov.com/docs/manual-github-app-creation)
So something like `https://github.com/apps/my-custom-app-name/installations/select_target`
Note that the slug ("GitHub App Name") is a lowercase hyphen separated string version of the entered name

Update the "Install Codecov GitHub app" link in the header to redirect to the custom app based on the existing env var (`REACT_APP_GH_APP=my-custom-app-name`).

Screenshot 2024-11-12 at 6 25 07 PM

Closes https://github.com/codecov/engineering-team/issues/2251


src/services/navigation/useNavLinks/useStaticNavLinks.test.tsx
```diff
@@ -60,7 +60,7 @@ describe('useStaticNavLinks', () => {
       ${links.uploaderCLI}                   | ${'https://docs.codecov.com/docs/codecov-uploader#using-the-cli-to-upload-reports-with-codecovio-cloud'}
       ${links.integrityCheck}                | ${'https://docs.codecov.com/docs/codecov-uploader#integrity-checking-the-uploader'}
       ${links.codecovGithubApp}              | ${'https://github.com/apps/codecov'}
```
spalmurray-codecov, line 62: maybe replace this one too? Not sure where used

spalmurray-codecov (approved):

src/services/navigation/useNavLinks/useStaticNavLinks.test.tsx
suejung-sentry: Oh yeah good catch. I updated to pull from the var (which is set to `codecov` by default). It's used in some banners it looks like - I won't hide in those cases since the close-enough link is probably better than no link

suejung-sentry:
Removed the "Switch context" row per this convo

Screenshot 2024-11-13 at 2 27 13 PM

suejung-sentry:
I'll wait for this before merging this - https://github.com/codecov/internal-issues/issues/1073

PR #3499, opened 2024-11-14
build: Vendor Split Popular Packages

nicholas-codecov:
# Description

This PR utilizes a practice called "vendor splitting", where packages that are used site wide are split out into their own chunk files. This brings some optimizations because these files change rarely and are able to be cached after first load reducing the amount of JS being loaded on initial load when the user visits the site.

# Notable Changes

- Use `rollupOptions.output.manualChunks` to split out packages into their own chunks


vite.config.mjs
```diff
@@ -68,6 +68,31 @@ export default defineConfig((config) => {
...
+              'lodash/isString',
+              'lodash/isNumber',
+              'lodash/isBoolean',
+              'lodash/isUndefined',
+              'lodash/isArray',
+              'lodash/gt',
+              'lodash/omit',
+              'lodash/flatMap',
+              'lodash/groupBy',
+            ],
```
calvin-codecov, line 94: would we be able to leverage the use of `['lodash']` from https://rollupjs.org/configuration-options/#output-manualchunks so that we don't need to list each individual function one by one?

calvin-codecov (approved):
Nice!

vite.config.mjs
nicholas-codecov: I attempted that, however it did not work and included everything. I think mainly because we're not using `@rollup/plugin-node-resolve` we're using whatever is built into Vite

calvin-codecov: thanks for checking

nicholas-codecov: Wanted to double double check, tried using that plugin and it did not work, so we're stuck with this 😢

PR #3522, opened 2024-11-20
add PROCESSING_TIMEOUT upload error code

matt-codecov:
a significant number of "unknown errors" that we report to users are timing out during processing. this PR creates an error code for it, and we can later choose to display different text in gazebo for that error code

note: worker will attempt to set this error code and save everything when it gets the "soft timeout" signal. but there is no guarantee it won't be forcibly killed before it can save everything. there will still be some unknown uploads from that

sibling PR in shared: https://github.com/codecov/shared/pull/431


nicholas-codecov (approved):
Send it

PR #3525, opened 2024-11-25
fix: Seperate org name from error banner on 404 page

nicholas-codecov:
# Description

This PR updates a couple things in the `BaseLayout`, `Header`, `Navigator`, and `GlobalTopBanners` to maintain previously resolved issues, and codecov/engineering-team#2830.

Closes codecov/engineering-team#2830

# Notable Changes

- Create new hook to determine information for the `Navigator` component
- Update logic in the `Navigator` to handle new cases
  - Repo Page
    - User has access to repo -> render breadcrumb
    - User and Owner exist but private repo -> render context switcher (so they can escape)
    - None of the above -> render nothing (there is the guest header as their escape hatch)
  - Any page when the user is not logged in and owner does exist
    - Render the context switcher
  - User not logged in and org doesn't exist -> don't render anything (guest header is the escape hatch)
- Wrap each banner in `GlobalTopBanners` to have their own `SilentNetworkErrorWrapper`
- Fetch data, and pass down from `BaseLayout`
- Create/Update tests
  - Had to rework the `wrapper` in the `Navigator` tests as paths were being rendered even if they didn't match

# Screenshots

Updated private repo page

Screenshot 2024-11-25 at 09 32 34


src/shared/GlobalTopBanners/GlobalTopBanners.tsx
```diff
@@ -14,13 +14,33 @@ const GlobalTopBanners: React.FC = () => {
...
+          <ProPlanFeedbackBanner />
+        </SilentNetworkErrorWrapper>
+
+        <SilentNetworkErrorWrapper>
+          <BundleFeedbackBanner />
+        </SilentNetworkErrorWrapper>
+
+        <SilentNetworkErrorWrapper>
+          <TokenlessBanner />
+        </SilentNetworkErrorWrapper>
```
nicholas-codecov, line 43: Switching this up, so that we're able to show some banners and not others, for instance if we're on a repo that the user needs to auth with Okta with, before this would just hide all the banners.

src/layouts/Header/components/Navigator/Navigator.tsx
```diff
@@ -57,7 +67,8 @@ function Navigator({ currentUser }: NavigatorProps) {
   }
 
   // Fallback instead of MyContextSwitcher if not logged in
-  if (!currentUser) {
+  // If the owner doesn't exist, don't show anything
+  if (!currentUser && ownerData) {
```
nicholas-codecov, line 69: If the user is viewing an owner that doesn't exist, it just rendered the Avatar icon with a question mark, so just removing that entirely, but leaving in the user controls on the right hand side.

src/layouts/BaseLayout/hooks/NavigatorDataQueryOpts.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+}`
+
+interface NavigatorDataQueryArgs {
+  provider: string
+  owner: string
+  repo: string
+}
+
+export const NavigatorDataQueryOpts = ({
```
spalmurray-codecov, line 54: Why are these lines not covered? Tests failing or something?

nicholas-codecov: It looks like the CI hasn't ran or something

spalmurray-codecov (approved):
looks good. Seems like a lot of stuff is not covered according to my browser extension though

spalmurray-codecov (approved):

PR #3535, opened 2024-11-28
feat: Grab the plugin name for the given bundle

nicholas-codecov:
# Description

This PR updates the query options for grabbing a list of assets used to display the asset table, so that we can determine if we want to render the new path column in the asset table

Closes codecov/engineering-team#2957

# Notable Changes

- Grab plugin name in assets table query options
  - also updated to be the TSQ V5 style we're moving to
- Update `useBundleAssetsTable`
- Update tests


calvin-codecov (approved):

src/pages/RepoPage/BundlesTab/BundleContent/AssetsTable/useBundleAssetsTable.ts
```diff
@@ -54,21 +55,23 @@ export function useBundleAssetsTable({
     }
   }, [overview?.oldestCommitAt, today, trend])
 
-  return useBundleAssets({
```
calvin-codecov, line 57: i like this change

PR #3552, opened 2024-12-04
feat: new UI to clarify adding token

calvin-codecov:
# Description

Closes https://github.com/codecov/engineering-team/issues/2698

Figma: https://www.figma.com/design/ZJmSdJUIz03zJdF22kLhAs/GH-2694?node-id=2058-3163&t=koQcNbcBaHcFN7rk-1 There are two versions in the file. Check the higher up version that is labeled "Latest".

With the addition of org level tokens and token auth possibly being required for an org, we have multiple variations as outlined in the figma file.

Landing on this repo onboarding page...
- the org token may or may not have been previously generated (could have been generated in their settings)
- the org may or may not require token authentication (this is set in their settings)
- the user may be an admin or just a regular member

Easiest way to test on preview:
If you have any repos on your personal org, go to the onboarding for one.
- Check what the steps look like without org token generated for both "token required" and "not token required" (setting can be changed in in the "Settings" tab -> Global Upload Token on org level.
- Click "Generate" on onboarding step and check what the steps look like for both token required/not required

# Notable Changes

I changed the Step components to move away from naming with number specific names and I also split them out from GitHubActions.tsx as that file was getting quite large.

# Screenshots

I've only expanded the screenshot example for "Token required - Prev Generated" but all of the others would look the same.

Token required - Not Prev Generated
Required - No Prev
Token required - Not Prev Generated but newly generated in this flow
Required - Generated
Token required - Repo instruction chosen (Not Prev Generated)
Required - Repo Chosen
Token required - Prev Generated
Required - Prev

Token optional - Not Prev Generated
Optional - No Prev
Token optional - Prev Generated
Optional - Prev
Token optional - Not Prev Generated but newly generated in this flow
Optional - Generated

Non-admin Disabled Generated Button
NonAdmin

Light mode screenshot
Screenshot 2024-12-12 at 3 12 17 PM

# Link to Sample Entry

# Legal Boilerplate


nicholas-codecov (approved):
One small comment, rest looks great!

PR #3558, opened 2024-12-05
fix: Broken tests when custom GH_APP set locally

spalmurray-codecov:
Was testing our the new changes to `GH_APP` and found some tests breaking when I had the config value set locally. This PR fixes those tests.


suejung-sentry (approved):
ooh sorry about that - thank you so much for fixing!!

PR #3563, opened 2024-12-10
Convert shared P5 to TypeScript

calvin-codecov:
# Description

Closes https://github.com/codecov/engineering-team/issues/2749

# Notable Changes

Main change was for AppLink component. The props it takes in and the split of the types of components it renders wasn't ideal for translation as we render a `<Component>...</Component>` where Component could be an `a` or `NavLink` or `Link`. It was a clever way to reuse code but I think it is worth consideration that we split these out for clarity, especially for parameter type specificity. Currently, all 3 instances receive the same props even though not all of them are needed for each type of component rendered.

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/RepoPage/ConfigTab/tabs/ConfigurationManager/components/FeatureGroup/FeatureGroup.tsx
```diff
@@ -1,10 +1,13 @@
+import { useNavLinks, useStaticNavLinks } from 'services/navigation'
 import A from 'ui/A'
 import Button from 'ui/Button'
 import { Card } from 'ui/Card'
 
 interface FeatureGroupProps extends React.PropsWithChildren {
   title: string
-  getStartedLink?: string // navLink key
+  getStartedLink?:
+    | keyof ReturnType<typeof useNavLinks>
```
ajay-sentry, line 9: do we need that first `|`?

src/shared/GlobalBanners/SelfHostedLicenseExpiration/SelfHostedLicenseExpiration.test.tsx
```diff
@@ -32,7 +32,7 @@ afterAll(() => {
 
 const wrapper =
```
ajay-sentry, line 33: nit: We usually type these as React.FC<React.PropsWIthChildren>

src/pages/RepoPage/ConfigTab/tabs/ConfigurationManager/components/FeatureGroup/FeatureGroup.tsx
calvin-codecov: Nope, it's optional - it's just for stylistic consistency.

src/shared/featureFlags/featureFlag.ts
```diff
@@ -36,12 +36,12 @@ export function useFlags(fallback) {
         'Warning! Self hosted build is missing a default feature flag value.'
       )
     }
-    return fallback
+    return fallback ?? {}
```
ajay-sentry, line 39: nit: could set this to the initial param as well since fallback is optional to begin with

ajay-sentry (approved):

PR #3573, opened 2024-12-12
chore: Migrate useSyncProviders to TS Query V5

nicholas-codecov:
# Description

This PR migrates the `useSyncProviders` to the TS Query queryOptions version `SyncProvidersQueryOpts`

Ticket: codecov/engineering-team#2962

# Notable Changes

- Migrate `useSyncProviders` to `SyncProvidersQueryOpts`
- Update usage in `SyncProviderPage`
- Update tests


src/pages/SyncProviderPage/SyncProviderPage.tsx
```diff
@@ -1,20 +1,17 @@
...
 import A from 'ui/A'
 
 import SyncButton from './SyncButton'
 
-interface SyncButtonProps {
-  hasSynced: boolean
-}
-
-const SyncButtons: React.FC<SyncButtonProps> = ({ hasSynced }) => {
-  const { data: syncProviders } = useSyncProviders({ enabled: !hasSynced })
```
suejung-sentry, line 17: do we lose any business logic by removing this `hasSynced` condition? Like the result of this operation is different now than before?

suejung-sentry (approved):
looks good! one question

src/pages/SyncProviderPage/SyncProviderPage.tsx
nicholas-codecov: No we we're not losing any business logic here, this was just being disabled, however the case where it would've been disabled the user is being redirected anyways so it won't be called.

suejung-sentry: Got it, okay thank you for clarifying!

PR #3576, opened 2024-12-12
feat: Migrate isEnterprisePlan to GQL Field

ajay-sentry:
# Description

This PR aims to remove the isEnterprisePlan helper function in Gazebo in favor of the new GraphQL field for the same.

As part of this PR there are also some changes removing the accountDetails hook has the SOT for retrieving a user's plan in favor of the usePlanData hook, which seems to be the "go-to" pattern for more recently created components anyway.

Outside of that, all the tests that touch plan details had to be updated to include this new field, and in the cases where we deprecated the useAccountDetails hook in favor of usePlanData we also had to add or update the mocks to reflect the differences in the plan mocks as well.

Closes https://github.com/codecov/engineering-team/issues/3061

This PR will probably be the "biggest" of these migration PRs because it's the first one. I don't expect others to be as meaty, though they may have just as many files touched

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/AccountSettings/AccountSettings.test.jsx
+++ b/src/pages/AccountSettings/AccountSettings.test.jsx
@@ -39,7 +39,6 @@ const mockPlanData = {
   pretrialUsersCount: 0,
   planUserCount: 1,
   hasSeatsLeft: true,
-  isEnterprisePlan: false,
 }
 
 const mockCurrentUser = (username) => ({
@@ -162,6 +161,8 @@ describe('AccountSettings', () => {
               plan: {
                 ...mockPlanData,
                 value: planValue,
+                isEnterprisePlan:
+                  planValue === Plans.USERS_ENTERPRISEM ? true : false,
               },
             },
           },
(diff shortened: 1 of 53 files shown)


src/pages/AccountSettings/AccountSettings.test.jsx
```diff
@@ -161,6 +161,8 @@ describe('AccountSettings', () => {
               plan: {
                 ...mockPlanData,
                 value: planValue,
+                isEnterprisePlan:
```
RulaKhaled, line 164: nit: could just do `planValue === Plans.USERS_ENTERPRISEM`

RulaKhaled (approved):
lgtm!

rohitvinnakota-codecov (approved):

src/pages/AccountSettings/AccountSettings.test.jsx
ajay-sentry: great point

PR #3577, opened 2024-12-12
chore: Migrate useSelfHostedSeatsAndLicense to TS Query V5

nicholas-codecov:
# Description

This PR migrates the `useSelfHostedSeatsAndLicense` to the TSQ V5 query options version `SelfHostedSeatsAndLicenseQueryOpts`, as well as a TS migration

Ticket: codecov/engineering-team#2961

# Notable Changes

- Migrate `useSelfHostedSeatsAndLicense` to `SelfHostedSeatsAndLicenseQueryOpts`
- Update `SelfHostedLicenseExpiration`
  - Refactor to TS
  - Add in tiny wrapper component so that we can short circuit rendering the actual banner and triggering the data fetching, as suspense queries cannot be controlled via an `enabled` property

### DIFF
--- a/src/shared/GlobalBanners/SelfHostedLicenseExpiration/SelfHostedLicenseExpiration.tsx
+++ b/src/shared/GlobalBanners/SelfHostedLicenseExpiration/SelfHostedLicenseExpiration.tsx
@@ -1,10 +1,11 @@
+import { useSuspenseQuery as useSuspenseQueryV5 } from '@tanstack/react-queryV5'
 import { differenceInCalendarDays } from 'date-fns'
 import { useState } from 'react'
 import { useParams } from 'react-router-dom'
 
 import config from 'config'
 
-import { useSelfHostedSeatsAndLicense } from 'services/selfHosted/useSelfHostedSeatsAndLicense'
+import { SelfHostedSeatsAndLicenseQueryOpts } from 'services/selfHosted/SelfHostedSeatsAndLicenseQueryOpts'
 import { Provider } from 'shared/api/helpers'
 import Banner from 'ui/Banner'
 import Button from 'ui/Button'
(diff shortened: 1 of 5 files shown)


src/shared/GlobalBanners/SelfHostedLicenseExpiration/SelfHostedLicenseExpiration.tsx
```diff
@@ -118,4 +116,18 @@ const SelfHostedLicenseExpiration = () => {
...
+  const isDedicatedNamespace = !!config.IS_DEDICATED_NAMESPACE
+
+  if (!isSelfHosted || !isDedicatedNamespace) {
+    return null
+  }
+
+  return <SelfHostedLicenseExpiration />
+}
+
+export default SelfHostedLicenseExpirationWrapper
```
suejung-sentry, line 133: I am probably missing it - where is this wrapper component meant to be used?

Was just wondering in case perhaps there's a way to create a generic "conditionalSuspense" or similar wrapper to handle if we have this scenario in other places

suejung-sentry (approved):
approved! Had a Q about suspense

src/shared/GlobalBanners/SelfHostedLicenseExpiration/SelfHostedLicenseExpiration.tsx
nicholas-codecov: So i played around a bit with this idea and came up with something like:

```tsx
// ConditionalRenderer.tsx
export const ConditionalRenderer = ({ condition, children }) => {
  if (condition) {
    return <>{children}</>
  }

  return null
}

// SelfHostedLicenseExpiration.tsx
function SelfHostedLicenseExpirationWrapper() {
  const isSelfHosted = !!config.IS_SELF_HOSTED
  const isDedicatedNamespace = !!config.IS_DEDICATED_NAMESPACE

  return (
    <ConditionalRenderer condition={isSelfHosted && isDedicatedNamespace}>
      <SelfHostedLicenseExpiration />
    </ConditionalRenderer>
  )
}

export default SelfHostedLicenseExpirationWrapper
```

Because of the way TanStack suspense queries work, we're still required to creating this wrapping component to conditionally render the hook call, so the resulting abstraction becomes fairly simple. Because of this I'd rather avoid trying to abstract it, and just handle the conditions directly in the same file. 

The main motivating factor behind keeping the conditional rendering in the same file, is that everything is defined up front right in front of the engineer, and if there are more complex condition cases we're not stuck trying to create a "one size fits all" conditional rendering component that is a pain to maintain and makes the app overall more fragile.

PR #3588, opened 2024-12-17
ref: Migrate isAnnualPlan && isMonthlyPlan to use BillingRate property from GQL

ajay-sentry:
# Description

This PR is part of the migration effort to remove the billing helpers from Gazebo, milestone 2: https://github.com/codecov/engineering-team/issues/2989 of the streamline plan creation work.

In particular, we remove the usage of three helpers:
- isAnnualPlan
- isMonthlyPlan
- isPaidPlan

and instead use the billingRate property off the Plan interface. In many cases you'll see in the code that where prior we were able to rely on "hardcoded" PlanNames, we now need to have the full Plan object to do any business logic against.

There are very few areas where we needed to do this billing rate check, but instead of having that layer of abstraction over the names, we can now more intuitively look at the billing rate property on the plan object itself to make that call. This means that in the future, when we have more plans with billing rates of annual or monthly, or even quarterly or whatever, we don't need to create a new helper function in Gazebo or make any modifications really at all. It'll be as simple as updating the Billing const created in billing.ts, and adding conditional logic to any of the now-plumbed-through Plan functions to make whatever minor changes needed.

In most cases, and until we add a new billing rate type we shouldn't need to make _any changes from here on out for billing rate!_

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/Controllers/Controller.tsx
```diff
@@ -10,8 +10,7 @@ import { UpgradeFormFields } from '../UpgradeForm'
 
 interface BillingControlsProps {
   seats: number
-  newPlan?: PlanName
-  selectedPlan: PlanName
+  newPlan?: Plan
```
ajay-sentry, line 13: This now takes a plan, and selectedPlan is no longer needed in this function either.

IN CASE ANYONE STUMBLES UPON THIS COMMENT LATER:
- SelectedPlan never "changes values" like newPlan does... so having newPlan be the source of truth here makes the most sense

src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/Controllers/ProPlanController/BillingOptions/BillingOptions.tsx
```diff
@@ -34,17 +28,23 @@ const BillingControls: React.FC<BillingControlsProps> = ({
 
   const currentPlanBillingRate = planData?.plan?.billingRate
   const [option, setOption] = useState<OptionPeriod>(() =>
-    currentPlanBillingRate === 'monthly'
+    currentPlanBillingRate === BillingRate.MONTHLY
```
ajay-sentry, line 31: using the const now instead of a magic string

src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/Controllers/ProPlanController/PriceCallout/PriceCallout.tsx
```diff
@@ -20,7 +19,7 @@ import Icon from 'ui/Icon'
 import { UpgradeFormFields } from '../../../UpgradeForm'
 
 interface PriceCalloutProps {
-  newPlan?: PlanName
+  newPlan?: Plan
```
ajay-sentry, line 22: NewPlan is now **ALWAYS** a Plan object and never just the plan "name"... it's more inline with what the variable says anyway lol

src/services/account/useUpgradePlan.ts
```diff
@@ -24,12 +25,12 @@ export function useUpgradePlan({ provider, owner }: useUpgradePlanParams) {
...
 
   return useMutation({
-    mutationFn: (formData: { seats: number; newPlan: string }) => {
+    mutationFn: (formData: { seats: number; newPlan: Plan }) => {
       const path = getPathAccountDetails({ provider, owner })
       const body = {
         plan: {
           quantity: formData?.seats,
-          value: formData?.newPlan,
+          value: formData?.newPlan.value,
```
ajay-sentry, line 33: You can see that even though we're passing in a Plan, we haven't modified the mutation object that the backend receives at the end of the day

src/shared/utils/billing.test.ts
```diff
@@ -535,7 +458,7 @@ describe('findTeamPlans', () => {
     const expectedResult = {
       marketingName: 'Team',
       value: Plans.USERS_TEAMY,
-      billingRate: 'yearly',
```
ajay-sentry, line 538: sometimes we were using 'yearly' and some other random stuff in our mocks for billingRate sadge

src/shared/utils/upgradeForm.ts
```diff
@@ -236,19 +234,24 @@ export const getDefaultValuesUpgradeForm = ({
     plans,
   })
 
-  const isMonthlyPlan = accountDetails?.plan?.billingRate === 'monthly'
+  const isMonthlyPlan =
+    accountDetails?.plan?.billingRate === BillingRate.MONTHLY
+  const isPaidPlan = !!accountDetails?.plan?.billingRate // If the plan has a billing rate, it's a paid plan
```
ajay-sentry, line 239: The only part of this diff that really matters is the new isPaidPlan const which is doing what the helper function was doing previously... this was also the only place that helper was used 😅

src/pages/PlanPage/subRoutes/UpgradePlanPage/UpgradeForm/Controllers/ProPlanController/PriceCallout/PriceCallout.tsx
ajay-sentry: Update: This is actually IndividualPlanSchema rather than Plan to better reflect the actual type

src/services/account/useAccountDetails.ts
```diff
@@ -117,9 +117,9 @@ export const PlanSchema = z
   .object({
     baseUnitPrice: z.number(),
     benefits: z.array(z.string()),
-    billingRate: z.string().nullable(),
+    billingRate: z.nativeEnum(BillingRate).nullish(),
     marketingName: z.string(),
-    monthlyUploadLimit: z.number().nullish(),
+    monthlyUploadLimit: z.number().nullable(),
```
ajay-sentry, line 122: this is actually nullable() vs. nullish(), just matching all the other schema types too

src/pages/PlanPage/subRoutes/CancelPlanPage/CancelPlanPage.test.tsx
```diff
@@ -20,15 +20,15 @@ const teamPlans = [
...
-    billingRate: 'monthly',
+    billingRate: BillingRate.MONTHLY,
     marketingName: 'Users Team',
     monthlyUploadLimit: 2500,
     value: Plans.USERS_TEAMM,
   },
   {
     baseUnitPrice: 5,
     benefits: ['Up to 10 users'],
-    billingRate: 'yearly',
```
adrian-codecov, line 31: I noticed some billingRates say `yearly` while others say `annually` (before your changes), did you confirm the new logic accounts for this?

adrian-codecov (approved):
Read through and left a comment or two, looks good! As spoken on dms, preferable method of testing is through the manual scenarios but overall looks good!

spalmurray-codecov (approved):

src/pages/PlanPage/subRoutes/CancelPlanPage/CancelPlanPage.test.tsx
ajay-sentry: yuhhhh we don't actually use yearly anywhere on the backend lol, maybe long ago we did and these mocks just never needed to be updated since we only typed it as a string vs. a const

ajay-sentry:
@adrian-codecov Went through and tested a slew of upgrade / downgrade scenarios and plan updates including payment method / address updates and all seems good :) Will merge this tomorrow morn first thing and monitor sentry / stripe webhooks

PR #3590, opened 2024-12-17
fix: Fix rendered header error

suejung-sentry:
When there is an error in the network calls in the Header component, fail silently (so no Header displays) instead of rendering the upside down umbrella fallback, resulting in double upside down umbrella in some cases, like [here](https://github.com/codecov/engineering-team/issues/3028#issue-2699715869).

This should close [#3028](https://github.com/codecov/engineering-team/issues/3028).

It will still be nice to do the work in [#3081](https://github.com/codecov/engineering-team/issues/3081) to have a more refined fallback strategy to show toasts or other specific behavior upon the different kinds of errors possible in the header. For now with this PR there is a catchall behavior of showing no Header so it is less jarring for the user.


spalmurray-codecov (approved):
nice!

PR #3597, opened 2024-12-19
ref: Migrate off "Plan" object from accountDetails REST endpoint to GQL Plan

ajay-sentry:
# Description

This PR aims to remove all the instances of accountDetails?.plan and similar, as well as removing that object from the accountDetails zod schema as well; to effectively prevent folks in Gazebo from continuing to use that as a SOT for plan data moving forward.

This needs to be done prior to us doing the API side of removing the plan object from the REST endpoint to avoid breaking Gazebo otherwise.

Closes https://github.com/codecov/engineering-team/issues/3126

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/services/account/useAccountDetails.ts
```diff
@@ -136,13 +122,12 @@ export const AccountDetailsSchema = z.object({
...
   name: z.string().nullable(),
   nbActivePrivateRepos: z.number().nullable(),
-  plan: PlanSchema,
   planAutoActivate: z.boolean().nullable(),
   planProvider: z.string().nullable(),
   repoTotalCredits: z.number(),
   rootOrganization: z
     .object({
-      plan: PlanSchema,
+      username: z.string().nullish(),
```
ajay-sentry, line 130: https://github.com/codecov/codecov-api/blob/1d7539e1694eda95d4f321429e75b4bb4b93f2ed/api/internal/owner/serializers.py#L247-L255

From here, was missing before

src/shared/utils/upgradeForm.test.ts
```diff
@@ -130,9 +128,7 @@ describe('getDefaultValuesUpgradeForm', () => {
     })
 
     it('returns sentry year plan if user is sentry upgrade', () => {
-      const accountDetails = {
-        plan: { value: Plans.USERS_BASIC, quantity: 1 },
-      } as z.infer<typeof AccountDetailsSchema>
+      const accountDetails = {} as z.infer<typeof AccountDetailsSchema>
```
RulaKhaled, line 131: could just define this once in the top of the file

RulaKhaled (approved):

src/shared/utils/upgradeForm.test.ts
ajay-sentry: good point, cleaned it up in [91efa0a](https://github.com/codecov/gazebo/pull/3597/commits/91efa0ae3adf26d773cb6fbad71b111f3e7b02c1)

PR #3601, opened 2024-12-20
chore: Migrate useOwner to ts

suejung-sentry:
Migrate useOwner hook to ts, including adding zod schema validation.

Note based on a spot-check, I assumed the api for the useOwner (DetailOwner) graphQL schema declared [here](https://github.com/codecov/codecov-api/blob/ad9247100bf3386252984f236acc452cd0eaa30b/graphql_api/types/owner/owner.graphql#L1C1-L44C2) to be unreliable, so marked all fields here as nullish (can be null or undefined).
In theory if the schema were correct, we would expect below:
```
const OwnerSchema = z
  .object({
    ownerid: z.number().nullable(),
    username: z.string().nullable(),
    avatarUrl: z.string(),
    isCurrentUserPartOfOrg: z.boolean(),
    isAdmin: z.boolean().nullable(),
  })
```
But going through the tests and also the implementation of the owner fields, `avatarUrl` and `username` were missing (only present on the `User` schema, not the `Owner` schema). Also some of the fields didn't look nullable based on the api implementation but were treated as nullable downstream in frontend code.

In general, since the owner related fields are sensitive (can be the difference between whole pages of the app rendering or not), I left it to be looser in what we consider a schema failure in this TS conversion ticket. I spun off this [ticket](https://github.com/codecov/engineering-team/issues/3134) for later to make any necessary logical changes in `api` to make this zod schema stricter.

Closes https://github.com/codecov/engineering-team/issues/2858


src/services/user/useOwner.ts
```diff
@@ -0,0 +1,74 @@
...
+    enabled?: boolean
+  }
+}
+
+export function useOwner({
+  username,
+  opts = { enabled: username !== undefined },
+}: UseOwnerArgs) {
+  const { provider } = useParams<URLParams>()
+  const query = `
```
ajay-sentry, line 39: nit: could you pull the query out of the hook?

src/services/user/useOwner.ts
```diff
@@ -0,0 +1,74 @@
...
+})
+
+export type Owner = z.infer<typeof OwnerSchema>
+
+const RequestSchema = z.object({
+  owner: OwnerSchema.nullish(),
+})
+
+interface URLParams {
+  provider: string
```
ajay-sentry, line 23: nit: This can now be typed out as Provider from `import { Provider } from 'shared/api/helpers'`

ajay-sentry (approved):

src/services/user/useOwner.ts
suejung-sentry: nice! fixed!

src/services/user/useOwner.ts
suejung-sentry: ah yeah good call - done!

PR #3603, opened 2024-12-20
[fix] Update GitHub, GitLab, BitBucket copy

rohitvinnakota-codecov:
Initially intended to fix copy on the login screen. Updated elsewhere as well. 

Screenshot 2025-01-03 at 11 47 20 AM

# Legal Boilerplate


adrian-codecov (approved):
Would it be useful to introduce an enum that holds the correct spelling for GH so it's seamlessly consistent everywhere it's used?

PR #3606, opened 2024-12-23
chore: Migrate onboardUser to ts

suejung-sentry:
https://github.com/codecov/engineering-team/issues/2860


suejung-sentry:
this work isn't needed as this hook isn't in use anywhere. closing

PR #3615, opened 2024-12-30
ref: Migrate useReposTeam to TS Query V5

nicholas-codecov:
# Description

This PR migrates the `useReposTeam` over to the `infiniteQueryOptions` API in TS Query V5, as well as updating its usage and some TS refactors.

Ticket: codecov/engineering-team#2696

# Notable Changes

- Migrate `useReposTeam` to `ReposTeamQueryOpts`
- Rename `services/repos/config` to `services/repos/orderingOptions`
- Update usage in `ReposTableTeam`
- Update tests
- Small TS refactor


src/services/repos/ReposTeamQueryOpts.test.tsx
```diff
@@ -178,13 +197,10 @@ describe('useReposTeam', () => {
...
               repos: [repo3, repo4],
               isCurrentUserPartOfOrg: true,
-              pageInfo: {
-                hasNextPage: false,
-                endCursor: 'aa',
-              },
+              pageInfo: { hasNextPage: false, endCursor: 'aa' },
             },
           ],
-          pageParams: [undefined, 'MjAyMC0wOC0xMSAxNzozMDowMiswMDowMHwxMDA='],
```
suejung-sentry, line 187: This behavior of returning `''` instead of `undefined` seems new and tricky - is that just this test behaving differently now, or has pageParams changed?

suejung-sentry (approved):

src/services/repos/ReposTeamQueryOpts.test.tsx
nicholas-codecov: In TSQ V5 you're required to provide a value to `initialPageParams` which cannot be undefined so we pass an empty string instead

nicholas-codecov: Docs: https://tanstack.com/query/latest/docs/framework/react/reference/useInfiniteQuery

suejung-sentry: nice, thanks!

PR #3617, opened 2024-12-31
ref: Migrate useRepos to TS Query V5

nicholas-codecov:
# Description

This PR migrates the `useRepos` hook over to the `infiniteQueryOptions` version `ReposQueryOpts` and to TS Query V5. 

~~There's some changes with how the imports from the `repos/config.ts` file that will be cleaned up once the prior PR #3615 has been migrated, and this PR can be rebased off of it.~~

Ticket: codecov/engineering-team#2969

# Notable Changes

- Migrate `useRepos` to `ReposQueryOpts`
- Update usage in `ReposTable` and `ChartSelectors`
- Update imports from `services/repos/config` (removing barrel file)
- Update/refactor tests

### DIFF
--- a/src/services/repos/ReposQueryOpts.tsx
+++ b/src/services/repos/ReposQueryOpts.tsx
@@ -132,7 +132,7 @@ function ReposQueryOpts({
   isPublic = null, // by default, get both public and private repos
 }: ReposQueryArgs) {
   const variables = {
-    filters: { activated, term, repoNames, isPublic: Boolean(isPublic) },
+    filters: { activated, term, repoNames, isPublic },
     ordering: sortItem?.ordering,
     direction: sortItem?.direction,
     first,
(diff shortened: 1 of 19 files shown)


calvin-codecov (approved):

src/services/repos/ReposQueryOpts.tsx
```diff
@@ -132,7 +132,7 @@ function ReposQueryOpts({
   isPublic = null, // by default, get both public and private repos
 }: UseReposArgs) {
```
calvin-codecov, line 133: nit: we could change this interface name too since the function is no longer `useRepos`

calvin-codecov (approved):
1 nit if you feel like adding.

Strange that we have that inconsistency. I went down a small rabbit hole to try to see if I could find out what was going on on the API side as it would be nice to keep it consistent.

src/services/repos/ReposQueryOpts.tsx
nicholas-codecov: Nice catch, updated!

PR #3629, opened 2025-01-03
feat: Add appless pre-PAT onboarding container

calvin-codecov:
# Description

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate


src/pages/OwnerPage/OnboardingContainerContext/context.tsx
spalmurray-codecov, line 1: I'm not really following why we need React context and localstorage. Could you explain how they work together?

calvin-codecov: I implemented with React context and localStorage to together solve for two situations.
- Persistence over time: I'm using localStorage to keep track between sessions and refreshes. So if a user hid it previously, it will stay hidden
- Toggling show vs hide in one session: using just the localStorage value to allow "showing" and "hiding" from the UserDropdown menu wasn't working because the react tree wasn't rerendering on just a localStorage value change. So I decided to use a setState variable. However, the onboarding container and the UserDropdown components are part of two different component trees so I added a context to share the state variable. The only shared parent they have is App.tsx and I decided against implementing a setState variable in App.tsx to be passed to each of them as I didn't want to clutter up App.

spalmurray-codecov: Interestinggggg okay. That's annoying but makes sense. In that case, one improvement I think we can do is handle all the local storage stuff inside of the context provider through a useEffect on the state change. That way the only place we interact with local storage is in the provider and the consumers don't have to worry about it.

I was also wondering why the need for the query param? Can't we use local storage/context for that as well?

calvin-codecov: Hmm, you're right. I'll move the extra outside instance of localStorage into the context.

For the query param, I was piggybacking off on our current use of `?source=onboarding` but not that I think about it more, we'll actually need to modify that construct too as we will no longer be showing DefaultOrgSelector anywhere in the flow. I will: 
- make a change for the onboarding container to rely on localStorage
- make a change in useUserAccessGate to remove the showDefaultOrgSelector logic
- make a change in the onboarding flow to set the source param to "onboarding" again as a replacement for ListRepo's use
I considered using the same construct for both bullet 1 and 3, but the localStorage value is tied more closely to whether or not we render the onboarding container while the query param is tied to the idea we came from onboarding

src/layouts/Header/components/UserDropdown/UserDropdown.test.tsx
```diff
@@ -72,18 +76,20 @@ const wrapper: (initialEntries?: string) => React.FC<React.PropsWithChildren> =
...
-            <Route
-              path="*"
-              render={({ location }) => {
-                testLocation = location
-                return null
-              }}
-            />
-          </Route>
-        </Switch>
+        <OnboardingContainerProvider>
```
calvin-codecov, line 79: only change here is wrapping in the Provider component

src/pages/TermsOfService/TermsOfService.tsx
```diff
@@ -114,6 +115,10 @@ export default function TermsOfService() {
       marketingConsent: data?.marketingConsent,
       customerIntent: data?.customerIntent || CustomerIntent.PERSONAL,
     })
+
+    const url = new URL(window.location.href)
+    url.searchParams.set('source', ONBOARDING_SOURCE)
+    window.location.href = url.toString()
```
calvin-codecov, line 121: adding a query param that used to be applied after the DefaultOrgSelector screen

src/assets/onboarding/org_list_install_app.png
spalmurray-codecov, line 1: Do we need both of these images?

src/layouts/BaseLayout/BaseLayout.tsx
```diff
@@ -112,15 +97,11 @@ function BaseLayout({ children }: React.PropsWithChildren) {
         <Suspense>
           <ErrorBoundary errorComponent={<EmptyErrorComponent />}>
             <SilentNetworkErrorWrapper>
-              {isFullExperience || isImpersonating ? (
+              {(isFullExperience || isImpersonating) && (
```
spalmurray-codecov, line 100: In general I think we prefer to use ternaries over && for conditional rendering. Ask Nick if curious there's some reason I can't remember lol.

src/layouts/BaseLayout/BaseLayout.tsx
```diff
@@ -30,20 +28,16 @@ const FullPageLoader = () => (
...
-  isImpersonating: boolean
   isFullExperience: boolean
   showAgreeToTerms: boolean
   redirectToSyncPage: boolean
-  showDefaultOrgSelector: boolean
 }
 
 function OnboardingOrChildren({
   children,
-  isImpersonating,
```
spalmurray-codecov, line 42: I believe the intention here is to never show onboarding when impersonating. What's the status of this for the new flow? It might not be relevant anymore, but I'm not sure. You could ask AJ to confirm whether this is something we want.

src/layouts/BaseLayout/hooks/useUserAccessGate.test.tsx
spalmurray-codecov, line 1: so much red 🥔

spalmurray-codecov (approved):
Some nit picks, looking great! will approve

src/layouts/BaseLayout/BaseLayout.tsx
calvin-codecov: you are correct. Lemme change this and the other instance

PR #3640, opened 2025-01-08
[chore] Add rule to make provider casing more consistent

rohitvinnakota-codecov:
The linter will now warn if we use Github instead of GitHub, same for GitLab. Also fixes casing errors for Bitbucket. 

This closes https://github.com/codecov/engineering-team/issues/3175

Screenshot 2025-01-08 at 11 18 59 AM

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -21,6 +21,7 @@ module.exports = {
     'plugin:tailwindcss/recommended',
   ],
   plugins: ['@vitest', 'jsx-a11y', 'import'],
+
   globals: {
     vi: true,
   },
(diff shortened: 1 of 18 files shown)


.eslintrc.cjs
nicholas-codecov, line 1: Do we want to include Bitbucket with these rules, we may mis-case it in the opposite direction BitBucket 👀

nicholas-codecov (commented):
Small addition to the eslint rule 👀

.eslintrc.cjs
rohitvinnakota-codecov: Sure, I definitely did that before

nicholas-codecov: Awesome thank you!

nicholas-codecov (approved):
Looks all good to me, pending the failing test (could be a flake)

adrian-codecov (approved):
🙌

PR #3642, opened 2025-01-08
Cy/new tos

calvin-codecov:
# Description

Closes https://github.com/codecov/engineering-team/issues/2828

Figma: https://www.figma.com/design/SsoxtY2SB73l0wiLbJ8FhZ/GH-2092?node-id=2288-16704&t=UJWDkZD99C5y1myx-1

Associated API PR: https://github.com/codecov/codecov-api/pull/1094

# Notable Changes

- Modified the TOS screen to ask for both email and name as required fields
- Removed some customerIntent instances as we no longer ask users

# Screenshots

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/pages/TermsOfService/TermsOfService.tsx
+++ b/src/pages/TermsOfService/TermsOfService.tsx
@@ -8,19 +8,18 @@ import config from 'config'
 import { SentryBugReporter } from 'sentry'
 
 import umbrellaSvg from 'assets/svg/umbrella.svg'
-import { CustomerIntent, useInternalUser } from 'services/user'
+import { useInternalUser } from 'services/user'
 import A from 'ui/A'
 import Button from 'ui/Button'
-import RadioInput from 'ui/RadioInput/RadioInput'
 import TextInput from 'ui/TextInput'
 
 import { useSaveTermsAgreement } from './hooks/useTermsOfService'
 
 const FormSchema = z.object({
-  marketingEmail: z.string().email().nullish(),
+  marketingName: z.string(),
+  marketingEmail: z.string().email(),
   marketingConsent: z.boolean().nullish(),
   tos: z.literal(true),
-  customerIntent: z.string(),
 })
 
 interface IsDisabled {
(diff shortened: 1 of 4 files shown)


src/pages/TermsOfService/TermsOfService.tsx
```diff
@@ -182,28 +198,13 @@ export default function TermsOfService() {
...
-              // @ts-expect-error - the types on this need to be updated to match the type of message
-              marketingEmailMessage={formErrors?.marketingEmail?.message}
-              showEmailRequired={
-                watch('marketingConsent') && !currentUser?.email
-              }
-            />
             <div className="flex gap-2">
               <input
-                {...register('tos', { required: true })}
+                {...register('tos')}
```
nicholas-codecov, line 207: Urrm, shouldn't this field still be required, if they don't agree the TOS they shouldn't be able to continue and use the app?

src/pages/TermsOfService/TermsOfService.tsx
```diff
@@ -233,7 +234,6 @@ export default function TermsOfService() {
                     privacy policy
                   </A>
                 </span>{' '}
-                <small className="text-xs">required</small>
```
nicholas-codecov, line 236: Small nit, from the designs it seems like this copy is still present 👀 

Screenshot 2025-01-15 at 09 01 27

nicholas-codecov (commented):
Couple things to peek at 👀

src/pages/TermsOfService/TermsOfService.tsx
calvin-codecov: so i was looking into this and the FormSchema above is already enforcing the validation so it wasn't necessary to have required here.

src/pages/TermsOfService/TermsOfService.tsx
calvin-codecov: hmm this should have been updated in the figma... I'm seeing this 
Screenshot 2025-01-15 at 8 36 39 AM
Are you still seeing the version you pasted?

src/pages/TermsOfService/TermsOfService.tsx
nicholas-codecov: We may want to add this in for accessibility purposes as the fields themselves do not indicate that they are required, only once the form is submitted does the error pop up.

src/pages/TermsOfService/TermsOfService.tsx
calvin-codecov: That is how the existing design looks. Now that we have 3 fields instead of just 1 that is required, we switched to pointing out which of the 4 fields will be optional instead of showing 3 of them are required.

nicholas-codecov: Ahhh okok tyty!

nicholas-codecov (approved):

PR #3650, opened 2025-01-10
feat: Implement EventTracker abstraction and Amplitude implementation

spalmurray-codecov:
Implements a new EventTracker abstraction and an Amplitude specific implementation of it. This provides a generic even tracking interface so we can swap out or change the implementation without major refactoring. This PR also adds some example 'Button Clicked' events that we're interested in tracking for our onboarding funnel.

Depends on https://github.com/codecov/codecov-api/pull/1095
Depends on AMPLITUDE_API_KEY env var getting set (done)

Closes https://github.com/codecov/engineering-team/issues/3178


src/services/events/events.ts
```diff
@@ -0,0 +1,54 @@
+import { Provider } from 'shared/api/helpers'
+
+import { AmplitudeEventTracker, initAmplitude } from './amplitude/amplitude'
+
+const AMPLITUDE_API_KEY = process.env.REACT_APP_AMPLITUDE_API_KEY
+
+export interface EventTracker {
```
ajay-sentry, line 7: Using an abstract class instead of an interface is a little more explicit of this Class's purpose

src/services/events/events.ts
```diff
@@ -0,0 +1,54 @@
...
+  ): void
+  track(
+    eventType: 'Page Viewed',
+    eventProperties: {
+      pageName: 'OwnerPage'
+    }
+  ): void
+}
+
+class StubbedEventTracker implements EventTracker {
```
ajay-sentry, line 31: I love the idea of having a stubbed / generic event tracker instance; maybe we can throw some errors in the functions instead for some higher visibility

Additionally, pulling this out into a separate file would help reduce some cognitive load when parsing this file

src/services/events/events.ts
```diff
@@ -0,0 +1,54 @@
...
+  identify(_userOwnerId: string | number, _username: string): void {}
```
ajay-sentry, line 32: Having this be a string | number seems a little smelly, is there a way around this?

src/services/events/events.ts
```diff
@@ -0,0 +1,54 @@
+import { Provider } from 'shared/api/helpers'
+
+import { AmplitudeEventTracker, initAmplitude } from './amplitude/amplitude'
+
+const AMPLITUDE_API_KEY = process.env.REACT_APP_AMPLITUDE_API_KEY
+
+export interface EventTracker {
+  // Identifies the user this session belongs to.
+  identify(userOwnerId: number | string, username: string): void
```
ajay-sentry, line 9: Personally I really like having functions utilize an object syntax for input parameters for explicit defining instead of relying on proper placement from developers

so in this case we'd have identify({userOwnerId, username}: {userOwnerId: number | string, username: string})

which would prevent a situation where someone accidentally puts the username as the first param unknowingly (since it also accepts a string)

src/services/events/events.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  // Add new events as a new overloaded method signature here.
+  // Please keep the `eventType`s very generic as we have a limited number of
+  // them. Instead, add more detail in `eventProperties` where possible.
+  // Adding event types this way provides type safety for event properties.
+  // E.g., every 'Button Clicked' event must have the buttonType property.
+  track(
+    eventType: 'Button Clicked',
```
ajay-sentry, line 17: Having a central const somewhere of "EventTypes" might help with readability

src/shared/ListRepo/InactiveRepo/InactiveRepo.tsx
```diff
@@ -26,6 +31,11 @@ function InactiveRepo({
           repo: repoName,
         },
       }}
+      onClick={() =>
+        eventTracker(provider, owner, repoName).track('Button Clicked', {
```
ajay-sentry, line 35: Is it possible to instantiate the event tracker somewhere as a singleton and then call track against that instance? Currently every time we want to call track we need to create a new instance of an AmplitudeEventTracker which seems a little expensive

src/services/events/events.ts
spalmurray-codecov: Tried that initially, but went with interface because it handled method overloads a little more cleanly. I agree though - particularly if we go with the Event type instead of overloads

src/services/events/events.ts
spalmurray-codecov: Yep that's fair.

src/services/events/events.ts
spalmurray-codecov: The goal here was to make the API as easy as possible to use. Give us the id in whatever type and we'll handle it. I will look around the codebase again though, I think it may be true that it's almost always a string or almost always a number. In that case we can get rid of the union type.

spalmurray-codecov: Yea. it's pretty much always a number, will update.

src/shared/ListRepo/InactiveRepo/InactiveRepo.tsx
spalmurray-codecov: Yea I considered this. The con of doing a singleton is that we then need some system of keeping the provider/owner/repo the instance is storing up-to-date. This could simply be done by moving those three into the `track()` call. Would like to talk more on this for sure

src/services/events/events.ts
spalmurray-codecov: My current plan is for the StubbedEventTracker to be used in development/testing/staging environments, so I'm hesitant to make it noisy. We can chat more though

Agree on pulling out to its own file.

src/services/events/events.ts
```diff
@@ -0,0 +1,64 @@
...
+let EVENT_TRACKER: EventTracker = new StubbedEventTracker()
+
+export function initEventTracker(): void {
+  // Sets the global EventTracker singleton and calls necessary init functions
+  try {
+    initAmplitude()
+    EVENT_TRACKER = new AmplitudeEventTracker()
+  } catch (e) {
+    if (process.env.REACT_APP_ENV === 'production') {
+      // If in production, we need to know this has occured.
```
ajay-sentry, line 22: nit: occurred

src/services/events/types.ts
```diff
@@ -0,0 +1,76 @@
...
+//  - Try to keep event property names unique to the event type to avoid
+//    accidental correlation of unrelated events.
+//  - Never include names, only use ids. E.g., use repoid instead of repo name.
+//
+
+export type Event =
+  | {
+      type: 'Button Clicked'
+      properties: {
+        buttonType: 'Install GitHub App' | 'Configure Repo'
```
ajay-sentry, line 26: Wondering if we should do a "Type of" and have these buttonTypes (and other event properties that will have a range of set values) live somewhere else for easier parsing and clarity

src/services/events/types.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+  | {
+      type: 'Page Viewed'
+      properties: {
+        pageName: 'Owner Page'
+      }
+    }
+
+export type Identity = {
+  userOwnerId: number
```
ajay-sentry, line 38: is this named as such just cuz owners can be orgs and users?

src/services/events/events.ts
```diff
@@ -0,0 +1,64 @@
...
+
+// Returns the global EventTracker instance.
+export function eventTracker(): EventTracker {
+  return EVENT_TRACKER
+}
+
+// Hook to keep the global EventTracker's context up-to-date.
+export function useEventContext() {
+  const { provider, owner, repo } = useParams<{
+    provider?: string
```
ajay-sentry, line 36: nit: This can be typed to "Provider"

src/services/events/types.ts
spalmurray-codecov: yep. want to be very clear what this is.

src/services/events/amplitude/amplitude.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import { providerToInternalProvider } from 'shared/utils/provider'
+
+import { Event, EventContext, EventTracker, Identity } from '../types'
+
+export function initAmplitude() {
+  const apiKey = config.AMPLITUDE_API_KEY
+  if (!apiKey) {
+    throw new Error(
+      'AMPLITUDE_API_KEY is not defined. Amplitude events will not be tracked.'
```
ajay-sentry, line 13: nit: Should we be telling the consumer they're being tracked if this errors out? :)

ajay-sentry: Actually this might be ok because you capture the exception in event.ts

src/services/events/types.ts
```diff
@@ -0,0 +1,88 @@
...
+//
+// If this union type grows too unwieldy, we can split each event out into its
+// own type.
+//
+
+export type Event =
+  | {
+      type: 'Button Clicked'
+      properties: {
+        buttonType: ButtonType
```
ajay-sentry, line 30: Thinking about this some more, this is almost the "label" for the button rather than the type, right? Or I guess we could consider it the "function" of the button. Type is a bit misleading for me after thinking a bit longer because I was thinking this would be "info" or "alert" or something.

Having some way to group buttons of the same "type" though is needed, I agree with the underlying intent

src/services/events/types.ts
```diff
@@ -0,0 +1,88 @@
...
+        buttonLocation?: string
```
ajay-sentry, line 31: another nit for these two properties in general is you could probably drop the "button" prefix

src/services/events/types.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export type Identity = {
+  userOwnerId: number
+  provider: Provider
+}
+
+// Describes the context to be attached to events. We can extend this as needs
+// arise in the future.
+export type EventContext = {
+  // owner the event is being performed ON, not BY.
```
ajay-sentry, line 49: Do we also have the "BY" being tracked somewhere?

src/services/user/useOwner.ts
```diff
@@ -73,7 +73,7 @@ export function useOwner({
           })
         }
 
-        return res?.data?.owner
```
ajay-sentry, line 76: nice catch wowie

src/services/events/types.ts
ajay-sentry: It would also be great to have some standards around what each of these properties should contain. Like what does "location" mean in this context? Is it page, component, etc. And when we input a value into this field should it be of some consistent format?

src/services/events/amplitude/amplitude.ts
spalmurray-codecov: Going to test this. I think Sentry might log it anyway lol.

spalmurray-codecov: Looks like it does not. We good.

src/services/events/types.ts
spalmurray-codecov: Amplitude best practices say to keep prefix because without it you run the risk of grouping unrelated event types. For example if we have type property on button click and a type on Checkout Completed - we might accidentally group on type, not knowing that it includes different event types. Make sense?

Fair point on standards. In this case, buttonLocation is used a non-technical description of where the button is located. Examples are 'User dropdown' or 'Repo list'. IMO it's function is to be a unique identifier of the button. E.g., presumably there is only one possible GH app install button in the user dropdown, but there are multiple GH app install buttons across the whole app.

Happy to chat more on this and am open to suggestions!

src/services/events/types.ts
spalmurray-codecov: Yep! This is tracked by the user session and `identify` method.

src/services/user/useOwner.ts
spalmurray-codecov: Lol yep 😭

nicholas-codecov (commented):
Two small tweaks

src/services/events/types.ts
ajay-sentry: what if we use "hook" which is something that I think is used for the <A/> and <AppLink/> components, which are basically just buttons if you think about them that way 🤔

ajay-sentry: buttonName might work too

src/services/events/types.ts
ajay-sentry: I was thinking we could go a step further even. Like position and component within the app potentially. and the _might_ be a way to derive that information for free via the stack trace (long term pipe dream style thing)

Thinking of buttonLocation to be like the button's location within the "DOM"

src/services/events/types.ts
ajay-sentry: Got it got it

ajay-sentry (approved):
S U P E R S O L I D 

send it

src/services/events/types.ts
spalmurray-codecov: We discussed this in Slack and chose to stick with `buttonLocation` as a non-technical description of the button's location, but we added a global event property called `path` which contains the current route.

src/services/events/events.ts
```diff
@@ -0,0 +1,33 @@
+import { captureException } from '@sentry/react'
+
+import config from 'config'
+
+import { AmplitudeEventTracker, initAmplitude } from './amplitude/amplitude'
+import { Event, EventContext, EventTracker, Identity } from './types'
+
+export class StubbedEventTracker implements EventTracker {
```
ajay-sentry, line 8: This file is so awesome

PR #3662, opened 2025-01-16
feat: Add ACH related hooks

suejung-sentry:
Add hooks related to project to add ACH payment method.

The below changes are made:
1. `useAccountDetails` [edit existing] - add usBankAccount as a possible return value
2. `useUpdateBillingEmail` [edit existing] - adds a knob to propagate the updated email to the default payment method
3. `useCreateStripeSetupIntent` [new] - This will be in order to cut over from the existing Stripe CardElement to their PaymentElement. PaymentElement requires a clientSecret from api which establishes a Stripe SetupIntent (done in api [here](https://github.com/codecov/codecov-api/pull/1083))
4. `updatePaymentMethod` [new] - pretty similar to existing updateCard except uses stripe's "confirmSetup" api instead of "createPayment". We can deprecate updateCard once we cut over to this one. This is needed to hook in with the PaymentElement stripe component (vs. CardElement)

Other notes:
Added the `Provider` type instead of `string` and that needed a bunch of files touched FYI

Closes https://github.com/codecov/engineering-team/issues/3109


src/services/account/useCreateStripeSetupIntent.ts
```diff
@@ -0,0 +1,84 @@
...
+  provider,
+  owner,
+  opts = {},
+}: UseCreateStripeSetupIntentArgs) {
+  return useQuery({
+    queryKey: ['setupIntent', provider, owner],
+    queryFn: ({ signal }) =>
+      createStripeSetupIntent({ provider, owner, signal }).then((res) => {
+        const parsedRes = CreateStripeSetupIntentSchema.safeParse(res.data)
+        if (!parsedRes.success) {
```
ajay-sentry, line 72: Do we want to return different stuff if one of the "error" types comes back?

src/services/account/useUpdatePaymentMethod.ts
```diff
@@ -0,0 +1,89 @@
...
+  provider: Provider
+  owner: string
+  email?: string
+}
+
+interface useUpdatePaymentMethodReturn {
+  reset: () => void
+  error: null | Error
+  isLoading: boolean
+  mutate: (variables: any, data: any) => void
```
ajay-sentry, line 21: is variables actually "any" ?

ajay-sentry (approved):
lgtm, couple thoughts on things

src/services/account/useUpdatePaymentMethod.ts
```diff
@@ -0,0 +1,89 @@
...
+              billing_details: {
+                email: email,
+              },
+            },
+            // eslint-disable-next-line camelcase
+            return_url: `${config.BASE_URL}/plan/${provider}/${owner}`,
+          },
+        })
+        .then((result) => {
+          if (result.error) return Promise.reject(result.error)
```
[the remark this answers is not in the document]
suejung-sentry: It would take the shape of a stripe error like below. And our current error framework would have that propagate to the user as the generic error page.
I don't think the contents of the error are ever shown to the user. I'll take a pass in the next PRs on the error flows that not missing something in there, good callout.
```
{
  "type": "some_error",
  "message": "Some message",
  "code": "some_code",
  "payment_intent": {
    "id": "pi_xyz",
    "status": "requires_payment_method"
  }
}
```

src/services/account/useCreateStripeSetupIntent.ts
suejung-sentry: Hm true, I'll have those errors reject as well (seems that is the pattern in other hooks too). I'll do a pass at the error pathways too to make sure these get reacted to correctly as well

src/services/account/useUpdatePaymentMethod.ts
suejung-sentry: Ah just copied that over from the updateCard implementation, good catch of a lazy moment lol. Updated.

PR #3666, opened 2025-01-17
ref: Remove tier name, use isTeamPlan 

RulaKhaled:
# Description
There's no need to fetch the tier name anymore; we should unify the queries to use isTeamPlan, as it's consistently used throughout the codebase.

# Notable Changes
- Refactored useTiers -> useIsTeamPlan
- Updated components and tests to align with the new hook structure.

closes: [#3240](https://github.com/codecov/engineering-team/issues/3240)

# Legal Boilerplate

### DIFF
--- a/src/services/useIsTeamPlan/useIsTeamPlan.ts
+++ b/src/services/useIsTeamPlan/useIsTeamPlan.ts
@@ -23,7 +23,7 @@ export interface UseIsTeamPlanArgs {
 }
 
 const query = `
-  query OwnerPlan($owner: String!) {
+  query IsTeamPlan($owner: String!) {
     owner(username:$owner){
       plan {
         isTeamPlan
@@ -34,7 +34,7 @@ const query = `
 
 export const useIsTeamPlan = ({ provider, owner }: UseIsTeamPlanArgs) =>
   useQuery({
-    queryKey: ['OwnerPlan', provider, owner, query],
+    queryKey: ['IsTeamPlan', provider, owner, query],
     queryFn: ({ signal }) =>
       Api.graphql({
         provider,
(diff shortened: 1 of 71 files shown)


src/services/useIsTeamPlan/useIsTeamPlan.ts
```diff
@@ -62,6 +53,15 @@ export const useTier = ({ provider, owner }: UseTierArgs) =>
           })
         }
 
-        return parsedRes.data?.owner?.plan?.tierName ?? null
+        return parsedRes.data?.owner?.plan?.isTeamPlan ?? null
       }),
   })
+
+export const TierNames = {
```
ajay-sentry, line 60: Should we move this const over to billing.ts to be with the other consts?

src/services/useIsTeamPlan/useIsTeamPlan.ts
```diff
@@ -3,47 +3,38 @@ import { z } from 'zod'
...
 
-export interface UseTierArgs {
+export interface UseIsTeamPlanArgs {
   provider: string
   owner: string
 }
 
 const query = `
-  query OwnerTier($owner: String!) {
+  query OwnerPlan($owner: String!) {
```
ajay-sentry, line 26: Should this query be called OwnerPlan or isTeamPlan?

The former implies this query does more than return isTeamPlan imo

src/services/useIsTeamPlan/useIsTeamPlan.ts
RulaKhaled: good call

ajay-sentry (approved):

src/shared/utils/billing.ts
```diff
@@ -149,3 +149,10 @@ export function lastTwoDigits(value: number | string) {
   }
   return null
 }
+
+export const TierNames = {
```
ajay-sentry, line 153: nit: Can we move these to the top of the file with the other consts? 😅

PR #3667, opened 2025-01-17
Cy/non pat appless

calvin-codecov:
# Description

Combination of https://github.com/codecov/engineering-team/issues/2828, https://github.com/codecov/engineering-team/issues/2734, and https://github.com/codecov/gazebo/pull/3629, which were all previously approved, with a couple final UI tweaks.

These 3 PRs together encapsulate the entire Non-PAT Appless feature (aka p1).

# Code Example

# Notable Changes

# Screenshots

# Link to Sample Entry

# Legal Boilerplate

Look, I get it. The entity doing business as "Sentry" was incorporated in the State of Delaware in 2015 as Functional Software, Inc. In 2022 this entity acquired Codecov and as result Sentry is going to need some rights from me in order to utilize my contributions in this PR. So here's the deal: I retain all rights, title and interest in and to my contributions, and by keeping this boilerplate intact I confirm that Sentry can use, modify, copy, and redistribute my contributions, under Sentry's choice of terms.

PR #3684, opened 2025-01-27
Cy/non pat appless 2

calvin-codecov:
# Description

Second attempt at rollout for the appless experience as the original (https://github.com/codecov/gazebo/pull/3667) was reverted. Closes https://github.com/codecov/internal-issues/issues/1162. The new changes are after the 7th commit, starting with b61b6468b7cd0426bdf57c9d135305f95fa0330e.

This PR fixes problem #2 of the ticket as well as adds more amplitude events. The plan is to monitor DB events coming through checking on users that are created (GH login) vs terms agreed (onboarded).

# Legal Boilerplate


src/services/account/usePlanData.ts
```diff
@@ -52,7 +52,7 @@ export const PlanDataSchema = z
   .object({
     owner: z
       .object({
-        hasPrivateRepos: z.boolean(),
+        hasPrivateRepos: z.boolean().nullish(),
```
calvin-codecov, line 55: this field is `hasPrivateRepos: Boolean` on the Owner graphql type and was causing unrelated zod errors

spalmurray-codecov (approved):

PR #3687, opened 2025-01-28
ref: Flatten Amplitude context properties

spalmurray-codecov:
While implementing the python side of things, we found that the object notation
was overcomplicating things there for little to no benefit in Gazebo or 
Amplitude. This PR flattens the Gazebo context object. This will change the keys
used in Amplitude, but it hasn't been too long and we're not even using these 
properties for anything yet :)


ajay-sentry (approved):
lgtm!

PR #3691, opened 2025-01-29
fix: Only show GH app config banner if admin

calvin-codecov:
# Description

Closes https://github.com/codecov/engineering-team/issues/3287

Background info copied from ^ for context:
In the new non-PAT appless experience, we made a shift of rendering the GH Config Banner from `<HeaderBanners>` to `<ListRepo>` which moved it outside of scope of the `SilentNetworkErrorWrapper`. The `account-details` API (used for getting data about if org has GH app installed) call 404s inside of that ErrorBoundary wrapper so anything `isGHApp` related would just error and be caught by the boundary. Now that we've moved `<GithubConfigBanner>` out, it needs more boolean checks to make sure we only show it for admins on their org that does not have the GH app already installed. We also don't want to show this at the same time as the "Welcome" banner when someone just onboards. 

# Notable Changes

Check that the user viewing the org page is an admin to get the GH Config Banner. Rest of the changes are just minor refactors and TS conversion made while I was there.

# Screenshots

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/shared/ListRepo/ListRepo.tsx
+++ b/src/shared/ListRepo/ListRepo.tsx
@@ -1,5 +1,3 @@
-/* */
-import PropTypes from 'prop-types'
 import { Suspense, useContext } from 'react'
 import { useParams } from 'react-router-dom'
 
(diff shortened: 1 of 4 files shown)


src/shared/ListRepo/ListRepo.tsx
```diff
@@ -30,40 +29,47 @@ export const repoDisplayOptions = Object.freeze({
...
+  const cameFromOnboarding = source === ONBOARDING_SOURCE
   const isMyOwnerPage = currentUser?.user?.username === owner
   const showDemoAlert = cameFromOnboarding && isMyOwnerPage
 
   return (
     <>
-      {/* we only want one of these two banners to show at a time */}
-      {!hasGhApp && !showDemoAlert && <GithubConfigBanner />}
+      {/* we only want one of this or DemoAlert banners to show at a time */}
+      {isAdmin && !hasGhApp && !showDemoAlert ? <GithubConfigBanner /> : null}
```
calvin-codecov, line 70: This is the main change.

spalmurray-codecov (approved):
Seems to be working! v nice

PR #3696, opened 2025-01-31
fix: Handle ach case

suejung-sentry:
Fixes scenario where user sets up an initial checkout session with an instant payment method (like credit card), then they go to put in a new delayed notification payment method (e.g., ACH microdeposits). Then during these 2 days of waiting, they go to upgrade their Codecov account from Team to Pro. This should NOT show the modal that says it will cancel their existing incomplete upgrade. The modal should only show if this is the initial checkout session. This PR fixes that logic.


ajay-sentry (approved):
looks good!

PR #3697, opened 2025-01-31
ref: Migrate useUpdateOktaConfig to TSQ V5

nicholas-codecov:
# Description

This PR migrates `useUpdateOktaConfig` to TSQ V5, and tidies up some logic within the hook.

Ticket: codecov/engineering-team#2963

# Notable Changes

- Update `useUpdateOktaConfig` to use TSQ V5
- Remove TSQ V4 from `useUpdateOktaConfig` tests
- Update `OktaConfigForm` to use `Provider` type (as it was updated as the arg to be passed in `useUpdateOktaConfig`)


src/pages/AccountSettings/tabs/OktaAccess/hooks/useUpdateOktaConfig.tsx
```diff
@@ -110,26 +104,21 @@ export const useUpdateOktaConfig = ({ provider, owner }: URLParams) => {
...
+        })
       }
 
       const error = parsedData.data.saveOktaConfig?.error
       if (error) {
-        if (
-          error.__typename === 'ValidationError' ||
-          error.__typename === 'UnauthorizedError' ||
-          error.__typename === 'UnauthenticatedError'
-        ) {
```
nicholas-codecov, line 126: I removed this error, because after testing if the mutation is successful then error is `null`, so we should show  the toast if `error` is defined.

spalmurray-codecov (approved):

PR #3699, opened 2025-01-31
fix: Onboarding continue button on Safari

spalmurray-codecov:
Fixes Safari only race condition in new onboarding flow.


calvin-codecov (approved):
LGTM

PR #3736, opened 2025-02-13
feat: Set a global staleTime value to QueryClients

nicholas-codecov:
# Description

This PR sets a global `staleTime` value for our `QueryClients`. By default this value is set to 0, so our queries are becoming stale instantly, so whenever a component is mounted, it will trigger a new fetch to be requested. By setting this value, it should reduce the amount of duplicate requests we make.

After discussions yesterday, we have set this value to 2mins.

Closes codecov/engineering-team#3350

# Notable Changes

- Set global `staleTime` value on V4 and V5 query clients

### DIFF
--- a/src/index.tsx
+++ b/src/index.tsx
@@ -42,10 +42,18 @@ const TOO_MANY_REQUESTS_ERROR_CODE = 429
 initEventTracker()
 setupSentry({ history })
 
+// setting to 2 minutes, this value will ensure that components that are mounted
+// after suspense do not trigger a new query to be fetched. By default, the
+// stale time value is 0, which means that the query will be re-fetched on every
+// mount.
+const QUERY_STALE_TIME = 2 * (1000 * 60)
+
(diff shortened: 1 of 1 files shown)


src/index.tsx
```diff
@@ -59,14 +67,15 @@ const queryClient = new QueryClient({
...
       },
-      refetchOnWindowFocus: false,
     },
   },
 })
 
 const queryClientV5 = new QueryClientV5({
   defaultOptions: {
     queries: {
+      staleTime: QUERY_STALE_TIME,
```
spalmurray-codecov, line 77: what about GC time?

nicholas-codecov: Value is set to 5mins by default

nicholas-codecov: Which is plenty of time imo

spalmurray-codecov: 🫡

spalmurray-codecov (approved):

PR #3739, opened 2025-02-13
ref: Update Amplitude page view sanitization

spalmurray-codecov:
I was looking at the available properties again and it seems we are missing out on a bunch by doing things in the way I had initially set it up. Now, instead of stating the keys we do want, we'll just delete the keys we don't want. This means we get whatever else amplitude sends, including some useful attribution stuff (utm_params, referrer domain, etc).


ajay-sentry (approved):
makes sense!

PR #3743, opened 2025-02-14
feat: Expand utility of rejectNetworkError

nicholas-codecov:
# Description

TODO

Closes codecov/engineering-team#3328

# Notable Changes

- TODO


package.json
```diff
@@ -6,6 +6,7 @@
     "dev": "yarn start",
     "start": "vite --host",
     "build": "tsc -b && vite build",
+    "preview": "vite preview",
```
spalmurray-codecov, line 9: v cool. Never seen this before

src/shared/api/rejectNetworkError.ts
```diff
@@ -0,0 +1,89 @@
...
+
+const determineSentryLevel = (errorName: NetworkErrorName) => {
+  switch (errorName) {
+    case 'Parsing Error':
+      return 'error'
+    case 'Not Found Error':
+      return 'info'
+    case 'Owner Not Activated':
+      return 'info'
+    default:
```
spalmurray-codecov, line 38: rip coverage. Maybe you could do a test with a bad input and a type ignore? or just leave it lol

spalmurray-codecov: wait the switch below doesn't require a default case, can't we get rid of this one too then?

nicholas-codecov: Whoops, the bottom one should have a default case ... just cause we have type safety ... doesn't mean we have runtime safety 😢

nicholas-codecov: `describe.each` is too powerful 😲

spalmurray-codecov: I mean in the real world. Why is type safety not sufficient? I've definitely trusted only type safety in things I've implemented. If I have a typed function signature, the only way a bad input gets through at runtime is if a developer is messing up no?

src/shared/api/rejectNetworkError.ts
```diff
@@ -27,7 +30,10 @@ type NetworkErrorObject =
   | NotFoundErrorObject
   | OwnerNotActivatedErrorObject
 
-const determineSentryLevel = (errorName: NetworkErrorName) => {
+/**
+ * @private exporting for testing - do not use
```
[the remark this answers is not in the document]
spalmurray-codecov: huh okay. So it doesn't actually prevent imports, it just asks the dev nicely not to 😅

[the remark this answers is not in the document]
nicholas-codecov: exactly

src/shared/api/rejectNetworkError.ts
nicholas-codecov: You see ... we ship raw code to the client to be executed ... and the client can do whatever they want to it during runtime, sooo best to do as much defensive programming as possible

spalmurray-codecov: fair enough

PR #3749, opened 2025-02-19
chore: Remove use of barrel files in services/branches

nicholas-codecov:
# Description

This PR removes the barrel index file from the `services/branches` directory, and updates the respective imports across the app.

Ticket: codecov/engineering-team#3352

# Notable Changes

- Remove `index.ts` from `services/branches`
- Update `services/branches` imports

### DIFF
--- a/src/pages/RepoPage/CoverageTab/OverviewTab/hooks/useBranchSelector.ts
+++ b/src/pages/RepoPage/CoverageTab/OverviewTab/hooks/useBranchSelector.ts
@@ -1,7 +1,6 @@
 import { useParams } from 'react-router-dom'
 
-import { useBranch } from 'services/branches/useBranch'
-import { Branch } from 'services/branches/useBranches'
+import { Branch, useBranch } from 'services/branches/useBranch'
 
 const getDecodedBranch = (branch?: string) =>
   branch ? decodeURIComponent(branch) : branch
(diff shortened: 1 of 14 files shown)


src/pages/RepoPage/CoverageTab/OverviewTab/hooks/useBranchSelector.ts
```diff
@@ -1,6 +1,7 @@
 import { useParams } from 'react-router-dom'
 
-import { Branch, useBranch } from 'services/branches'
+import { useBranch } from 'services/branches/useBranch'
+import { Branch } from 'services/branches/useBranches'
```
ajay-sentry, line 4: This got me curious so I checked the Branch type and it's just a wrapper around the zod schema, but for a different hook

While it still matches https://github.com/codecov/gazebo/blob/0899f21d7fe6eae4248cce5fdd65c7043f4e247f/src/services/branches/useBranch.tsx#L23, we should probably be using that type in the event the branch type in that hook actually being used changes

ajay-sentry: A separate discussion can be had for consolidating the types to one too

nicholas-codecov: Because of GQL, and how we fetch what we need vs a pre-defined set, it's not as easy to "consolidate" these types.

ajay-sentry (approved):
Lgtm!

PR #3751, opened 2025-02-19
feat: Move initial sunburst data into react state

nicholas-codecov:
# Description

This PR moves the initial data from being setup in the `useLayoutEffect` into `useState` so that we only do this calculation once, and is the first step into moving what is being rendered from the DOM into React state.

Closes: codecov/engineering-team#3359

# Notable Changes

- Move `arcVisible` into `utils` function
- Move initial state into `useState`, some tidying up of the `SunburstChart` component


src/ui/SunburstChart/utils.js
```diff
@@ -10,9 +10,16 @@ export const partitionFn = (data) => {
...
+
 export const formatData = format(',d')
+
 // Interpolate in between --color-ds-primary-red, --color-ds-primary-green
 export const colorRange = interpolateHslLong(
   'rgb(254, 0, 0)',
   'rgb(33, 181, 119)'
 )
+
+// Calculate if a arc is visible
```
ajay-sentry, line 22: nit: an vs a

Also, this function is pretty hard to make sense between the name and the calculation being done. How did we land on the magic numbers of 3 and 1 and the x1 value being greater than x0?

What are our thoughts on converting this utils to TS as well?

src/ui/SunburstChart/SunburstChart.jsx
```diff
@@ -24,13 +25,41 @@ function SunburstChart({
   const clickHandler = useRef(onClick)
   const hoverHandler = useRef(onHover)
 
+  const [root] = useState(() => {
```
ajay-sentry, line 28: a code comment would probably be nice for this useState

src/ui/SunburstChart/utils.js
nicholas-codecov: Those details come from this [example sunburst](https://observablehq.com/@d3/zoomable-sunburst)

nicholas-codecov: TS refactor will come later

ajay-sentry: Gotcha, yeah I just noticed the code comment at the top of that file too, thanks for the context!

ajay-sentry (approved):
lgtm

PR #3767, opened 2025-02-24
feat: Add rejectNetworkError to calls (part three)

nicholas-codecov:
# Description

This PR is part three in a series, where we're going through and updating our query requests to utilize the new `rejectNetworkError` helper function.

Ticket: codecov/engineering-team#3329

# Notable Changes

- Update hooks to use `rejectNetworkError` (see [commits](https://github.com/codecov/gazebo/pull/TBD/commits))
- Remove redundant function in `useRepoComponentsArgs`
- Update tests

### DIFF
--- a/src/services/pull/useSingularImpactedFileComparison.tsx
+++ b/src/services/pull/useSingularImpactedFileComparison.tsx
@@ -12,7 +12,7 @@ import {
   RepoOwnerNotActivatedErrorSchema,
 } from 'services/repo/schemas'
 import Api from 'shared/api'
-import { NetworkErrorObject } from 'shared/api/helpers'
+import { rejectNetworkError } from 'shared/api/rejectNetworkError'
 import A from 'ui/A'
 
 import {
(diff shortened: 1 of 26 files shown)


src/services/pull/useSingularImpactedFileComparison.tsx
```diff
@@ -166,11 +167,8 @@ export function useSingularImpactedFileComparison({
           )
         }
 
-        return Promise.reject({
-          status: 404,
-          data: {},
-          dev: 'useSingularImpactedFileComparison - 404 missing data',
-        } satisfies NetworkErrorObject)
+        // we can set to null, and use the error display message we currently have, rather than throwing
```
spalmurray-codecov, line 170: not sure I understand what's going on here. Can you provide some more context?

spalmurray-codecov (approved):
One question, but I trust you know what you're doing lol

src/services/pull/useSingularImpactedFileComparison.tsx
nicholas-codecov: so before we were throwing an error here rather than returning anything. However, in all locations where this is being used, we're checking to ensure data exists and is not a nullish value, and if it is, we display an error message.

PR #3768, opened 2025-02-24
feat: Add rejectNetworkError to API calls (part four)

nicholas-codecov:
# Description

This PR is part four in a series going through Gazebo updating our API calls to utilize `rejectNetworkError` in functions where we call the API.

Ticket: codecov/engineering-team#3329

# Notable Changes

- Update API calls to use `rejectNetworkError` (see [commits](https://github.com/codecov/gazebo/pull/3768/commits))
- Update tests

### DIFF
--- a/src/services/repo/useRepoOverview.tsx
+++ b/src/services/repo/useRepoOverview.tsx
@@ -85,7 +85,6 @@ export function useRepoOverview({
         const parsedData = RequestSchema.safeParse(res?.data)
 
         if (!parsedData.success) {
-          console.debug(parsedData.error)
           return rejectNetworkError({
             errorName: 'Parsing Error',
             errorDetails: {
--- a/src/services/repo/useRepoOverview.tsx
+++ b/src/services/repo/useRepoOverview.tsx
@@ -2,7 +2,7 @@ import { useQuery } from '@tanstack/react-query'
 import { z } from 'zod'
 
 import Api from 'shared/api'
-import { NetworkErrorObject } from 'shared/api/helpers'
+import { rejectNetworkError } from 'shared/api/rejectNetworkError'
 
 import { RepoNotFoundErrorSchema } from './schemas'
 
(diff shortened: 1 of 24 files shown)


src/services/repo/useRepoSettings.tsx
```diff
@@ -145,6 +78,62 @@ export function useRepoSettings() {
...
+          repo,
+        },
+      }).then((res) => {
+        const parsedRes = RequestSchema.safeParse(res?.data)
+
+        if (!parsedRes.success) {
+          return rejectNetworkError({
+            errorName: 'Parsing Error',
+            errorDetails: {
+              callingFn: 'fetchRepoSettingsDetails',
```
ajay-sentry, line 96: General comment -> What do we think about making this a const so we don't accidentally misspell it for one of these rejectNetworkError calls?

src/services/repo/useRepoOverview.tsx
```diff
@@ -85,21 +85,25 @@ export function useRepoOverview({
         const parsedData = RequestSchema.safeParse(res?.data)
 
         if (!parsedData.success) {
-          return Promise.reject({
-            status: 404,
-            data: {},
-            dev: 'useRepoOverview - 404 failed to parse',
-          } satisfies NetworkErrorObject)
+          console.debug(parsedData.error)
```
ajay-sentry, line 88: console.log

ajay-sentry (approved):

src/services/repo/useRepoSettings.tsx
nicholas-codecov: Just playing around with this a bit, instead I think we can actually use the function definition, and the `name` property from it. That way it's typesafe as we're accessing properties off of the function declaration and no magic strings. Sadly there's no way to infer this from the calling function.

nicholas-codecov: Actually nvm, that won't work. Minification will change the function names

nicholas-codecov: Yea we can do that, however I'll do that in a follow up PR and adjust the previous changes that I have already made, and continue with this going forward.

PR #3769, opened 2025-02-24
chore: Replace old onboarding metrics with Amplitude events

spalmurray-codecov:
This PR removes the old onboarding metrics implementation replacing with Amplitude where applicable.

Closes https://github.com/codecov/engineering-team/issues/3125


src/services/events/types.ts
```diff
@@ -29,8 +30,16 @@ export type Event =
       type: 'Button Clicked'
       properties: {
         buttonName: ButtonName
-        buttonLocation?: string
+        buttonLocation?: string // Non-technical description of the button location (e.g., Coverage onboarding)
         loginProvider?: ReturnType<typeof loginProviderToName> // for login buttons only
+        ciProvider?: 'GitHub Actions' | 'CircleCI' | 'Codecov CLI' // E.g., product onboarding pages
+        testingFramework?: Framework // E.g., product onboarding pages
+        copied?:
```
spalmurray-codecov, line 37: could see an argument for using a string type here rather than string literal union. Lmk thoughts if you have any

rohitvinnakota-codecov: Can we put these in an enum and infer the type from there? similar to `loginProvider`

rohitvinnakota-codecov (approved):

src/pages/RepoPage/CoverageOnboarding/CircleCI/CircleCI.test.tsx
```diff
@@ -339,12 +334,11 @@ describe('CircleCI', () => {
...
-      await user.click(copyCommands[1] as HTMLElement)
+      const promises: Promise<void>[] = []
+      copyCommands.forEach((copy) => promises.push(user.click(copy)))
+      await Promise.all(promises)
 
-      await user.click(copyCommands[2] as HTMLElement)
-      await waitFor(() =>
-        expect(mockMetricMutationVariables).toHaveBeenCalledTimes(1)
-      )
+      await waitFor(() => expect(eventTracker().track).toHaveBeenCalledTimes(4))
```
ajay-sentry, line 341: how come this is 4 and not 5 when there's 5 copy commands? :O

src/pages/RepoPage/CoverageOnboarding/GitHubActions/GitHubActions.test.tsx
```diff
@@ -210,22 +204,31 @@ describe('GitHubActions', () => {
...
-      expect(copyCommands.length).toEqual(5)
+      expect(copyCommands.length).toEqual(6)
 
       const promises: Promise<void>[] = []
       copyCommands.forEach((copy) => promises.push(user.click(copy)))
       await Promise.all(promises)
 
       // One of the code-snippets does not have a metric associated with it
-      expect(mockMetricMutationVariables).toHaveBeenCalledTimes(4)
+      expect(eventTracker().track).toHaveBeenCalledTimes(5)
```
ajay-sentry, line 231: Similar here with the off by one

src/pages/RepoPage/CoverageOnboarding/CircleCI/CircleCI.test.tsx
spalmurray-codecov: It's correct as one of the copy buttons doesn't emit an event! This one:

Screenshot 2025-02-25 at 08 39 07

src/services/events/types.ts
spalmurray-codecov: Pulled it out to its own type. Same result, but don't think we need the extra enum features

src/services/events/types.ts
```diff
@@ -92,5 +96,23 @@ type ButtonName =
   | 'Continue'
   | 'Login'
   | 'Sync'
+  | 'Copy'
+
+type ButtonLocation =
```
spalmurray-codecov, line 101: Added this type because I accidentally used both `Coverage onboarding` and `Coverage Onboarding`. I expected these strings to be more or less unique to each call, but that seems to not be the case. If we're reusing locations, we ought to have type safety to avoid mistakes like mine.

PR #3783, opened 2025-02-27
ref: Render only two levels of the sunburst

nicholas-codecov:
# Description

The goal of this PR is to optimize the sunbursts performance overall and reduce the lagginess when viewing the coverage overview page. This was accomplished by changing the way that we render the sunburst. Previously we would render the entire tree structure of the repo, hiding elements that weren't currently visible. This lead to many performance issues as we were creating two elements per file in a repo. When users interacted with the sunburst, under the hood it was actually D3 updating the elements attributes overtime "animating" the transition.

With this refactor we have changed our rendering strategy, and how users interact with the sunburst:

**Initial render:**
1. Process and set root node
2. Set the root node as the selected node
3. Run effect
    1. Call `renderArcs`
        1.  Determine nodes to render
        2. Render two levels of arcs
        3. Set events for folders and files
        4. Format and append title
        5. Setup back circle and back text (i.e. `..`)
        6. Setup click event handlers

**On Click**
1. User clicks on a folder
    1. Update coordinates for selected node (i.e. directory)
    2. Set selected node to the clicked node with updated coordinates
2. Re-run effect
    1. Call `renderArcs`

One regression of this change is the removal of the animation with the sunburst, as the change to the render process doesn't support this. We may explore some potential solutions for bringing this back down the road, however for the time being we will be leaving it as is.

Closes: codecov/engineering-team#3382

# Notable Changes

- Remove work that is not required for processing initial dataset
- Move to storing selected node to React state
    - When user clicks on directory, set as selected node 
- Move D3 drawing into a render function

### DIFF
--- a/src/ui/SunburstChart/SunburstChart.jsx
+++ b/src/ui/SunburstChart/SunburstChart.jsx
@@ -27,38 +27,39 @@ function SunburstChart({
   const hoverHandler = useRef(onHover)
 
   // this state stores the root node of the sunburst chart
-  const [root] = useState(() =>
+  const [{ root }] = useState(() =>
     Sentry.startSpan({ name: 'SunburstChart.createRoot' }, () => {
       // go through the data and add `value` to each node
       const stack = [data]
-      const nodeMap = new Map()
+      const depthIndex = new Map()
 
       // create a new root node with the value of the root node
       const result = { ...data, value: selectorHandler.current(data) }
-      // add the root node to the node map
-      nodeMap.set(data, result)
 
       // while there are nodes to process, pop the last node from the stack
       while (stack.length > 0) {
         const node = stack.pop()
-        const currentNode = nodeMap.get(node)
+
+        if (!('value' in node)) {
+          node.value = selectorHandler.current(node)
+        }
 
         // if the node has children, process them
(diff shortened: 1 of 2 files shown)


src/ui/SunburstChart/SunburstChart.jsx
```diff
@@ -75,24 +67,29 @@ function SunburstChart({
...
     // A color function you can pass a number from 0-100 to and get a color back from the specified color range
     // Ex color(10.4)
-    const color = Sentry.startSpan({ name: 'SunburstChart.color' }, () => {
-      return scaleSequential()
-        .domain([colorDomainMin, colorDomainMax])
-        .interpolator(colorRange)
-        .clamp(true)
-    })
+    function createColorFunction(parentSpan) {
+      return Sentry.startSpan({ name: 'SunburstChart.color', parentSpan }, () =>
```
spalmurray-codecov, line 86: do we use yaml set color indication range for this? seems like no

nicholas-codecov: We do not no

spalmurray-codecov (approved):

PR #3787, opened 2025-03-03
feat: Setup Spotlight integration for local dev

nicholas-codecov:
# Description

Did some exploring with @BYK utilizing Spotlight, and since it was an easy add I thought I'd commit it, with some other updates for Sentry as well.

# Notable Changes

- Conditionally add Spotlight integration into Sentry integrations list
- Add document policy header for dev and preview modes in Vite
- Update to latest version of Sentry JS SDK and Sentry Vite plugin


vite.config.mjs
```diff
@@ -59,9 +59,15 @@ export default defineConfig((config) => {
   return {
     server: {
       port: 3000,
+      headers: {
```
spalmurray-codecov, line 62: haven't worked with vite config before so just sanity checking: this doesn't end up in prod OR it does and we want it to, right?

spalmurray-codecov (approved):
one q. lgtm

vite.config.mjs
nicholas-codecov: So this isn't a worry cause we don't use Vite, to serve anything in prod

nicholas-codecov: And if we did ... we'd want this value being set 😄

spalmurray-codecov: Oh good point, a file being served can't add headers

spalmurray-codecov: Based on your second comment, should we add this header to our file server config?

nicholas-codecov: That's been done for a hot minute, this was just something extra so we could collect profiling info in spotlight from a local instance

spalmurray-codecov: Sick okie doke

PR #3805, opened 2025-03-11
fix: Various usability tweaks on the members page

spalmurray-codecov:
- Adds a spinner to indicate the query is in flight.
- To achieve the spinners on members list I did a refactor that pulls the mutation down into the table component itself. Also gets rid of unnecessary prop drilling.
- Prevents the toggle div from filling the whole card width (whole div is clickable).
- Right aligns toggles in the members list.

Closes https://github.com/codecov/engineering-team/issues/3325


src/pages/MembersPage/MembersList/MembersList.test.jsx
```diff
@@ -339,123 +291,4 @@ describe('MembersList', () => {
       })
     })
   })
-
```
spalmurray-codecov, line 342: Logic for this was moved down to MembersTable, so moved the tests for this logic there as well.

nicholas-codecov (approved):

PR #3807, opened 2025-03-12
fix: sed command for self-hosted git provider replacement

spalmurray-codecov:
Fixes the sed command that replaces GHE/GLE/BBS links in the frontend for self-hosted.

Closes https://github.com/codecov/feedback/issues/600


spalmurray-codecov (commented):
oopsie left echos in

PR #3830, opened 2025-04-01
fix: some flakes

spalmurray:
Fixes a few flakes. There are probably more but CI has passed twice in a row.


suejung-sentry (approved):
oof nice finds - sadge indeed. thanks so much for doing this!

PR #3834, opened 2025-04-03
[ft] Re-Add test gen copy

rohitvinnakota-codecov:
The copy was removed since we were using the Github Copilot extension. After some back and forth, we are supporting it via Codecov AI again. 

# Legal Boilerplate


src/pages/AccountSettings/shared/Header/Header.tsx
```diff
@@ -6,7 +6,7 @@ import TabNavigation from 'ui/TabNavigation'
 
 function Header() {
   const { codecovAiFeaturesTab } = useFlags({
-    codecovAiFeaturesTab: false,
+    codecovAiFeaturesTab: true,
```
ajay-sentry, line 9: just to confirm, we want to default the flag to true now?

ajay-sentry (approved):
niceeee

PR #3839, opened 2025-04-10
fix: Remove duplicate calls in coverage overview page

suejung-sentry:
Remove duplicate GetBranches api call in coverage overview page when search term is evaluated as null vs. "". As in treat those as the same calls so we don't hit api twice but use the cached react-query results.
The call takes around 200 ms so cutting it down from 2 to 1 saves that much time which is nice

Closes https://github.com/codecov/engineering-team/issues/3524

The fix is on the page that looks like:
Screenshot 2025-04-09 at 5 08 20 PM


Swatinem (approved):
nice, seems to be quite a simple change

PR #3842, opened 2025-04-11
feat: add all branches option to TA branch selector

joseph-sentry:
we want to allow users to not filter by any specific branch and aggregate the results from all branches

we also want to make this the default option when users land on the tests tab from now on

i'm putting this behind a feature flag because the backend is not capable of serving this data yet since it relies on the TA timescale data migration, and there's an equivalent API change that needs to be made for this to work correctly as well

fixed a thing where when users were searching through the branch selector for a specific branch we were still pinning the default branch to the top of the selection

also the all branches view has all the nice stuff that the default branch view has: metrics section, reset to default button, etc.

### DIFF
--- a/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
+++ b/src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
@@ -14,6 +14,7 @@ import { useInView } from 'react-intersection-observer'
 import { useLocation, useParams } from 'react-router-dom'
 
 import { MeasurementInterval } from 'pages/RepoPage/shared/constants'
+import { ALL_BRANCHES } from 'services/navigation/useNavLinks'
 import { formatTimeToNow } from 'shared/utils/dates'
 import Icon from 'ui/Icon'
 import Spinner from 'ui/Spinner'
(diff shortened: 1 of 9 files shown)


src/pages/RepoPage/FailedTestsTab/FailedTestsPage/SelectorSection/BranchSelector/BranchSelector.tsx
```diff
@@ -60,37 +69,76 @@ const BranchSelector = () => {
   })
 
   let selection = searchBranchValue?.branch
-  if (!selection) {
-    selection = {
-      name: 'Select branch',
-      head: null,
+  console.log('selection', selection)
```
nicholas-codecov, line 72: 🚨 🚓 🚨

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -308,7 +310,8 @@ const FailedTestsTable = () => {
         <div className="mt-4 text-center text-ds-gray-quinary">
           <p>No data yet</p>
           <p>
-            To see data for the main branch, merge your PR into the main branch.
+            To see data for the {testData?.defaultBranch} branch, merge your PR
+            into the {testData?.defaultBranch} branch.
```
nicholas-codecov, line 314: Should there be some fallback copy if the default branch doesn't exist?

nicholas-codecov (commented):
One small thing to clean up, but looks all good 👍

PR #3843, opened 2025-04-16
fix: Remove redundant call - coverage file

suejung-sentry:
Remove the redundant call to the flags api. Was formerly calling with `filters: {term: ""}` and also `filters: {}` so reduce it down to 1 call. These behave same on the [api side](https://github.com/codecov/codecov-api/blob/795e37f5bb45e311b351af47442f9cad2cdb1afb/graphql_api/actions/flags.py#L27) so that's no issue.
Closes https://github.com/codecov/engineering-team/issues/3529


Swatinem (approved):
a oneliner to avoid a duplicated request… I take it 🎉

PR #3844, opened 2025-04-17
fix: Remove duplicate calls to GetBranchCoverageMeasurements

suejung-sentry:
There is a duplicate call to GetBranchCoverageMeasurements due to milliseconds drift across re-renders when calling new Date(). Remove the milliseconds portion of the date (to the closest second is perfectly fine for this use case of showing the trend with the finest granularity being for the last day). Also uses useRef instead of useMemo in order to in order to guarantee stability across re-renders and component remounts, since useMemo(() => new Date(), []) may still be re-evaluated when the component remounts

Screenshot 2025-04-17 at 11 25 17 AM

### DIFF
--- a/src/shared/utils/timeseriesCharts.ts
+++ b/src/shared/utils/timeseriesCharts.ts
@@ -155,6 +155,9 @@ export function createTimeSeriesQueryVars({
   oldestCommit: Date | null
 }) {
   let after = calculateTrendDate({ today, trend })
+
+  console.log('today', today)
+  console.log('after', after)
   if (after === null) {
     after = oldestCommit
   }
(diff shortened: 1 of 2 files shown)


src/shared/utils/timeseriesCharts.ts
```diff
@@ -155,6 +155,9 @@ export function createTimeSeriesQueryVars({
   oldestCommit: Date | null
 }) {
   let after = calculateTrendDate({ today, trend })
+
+  console.log('today', today)
+  console.log('after', after)
```
nsdeschenes, line 160: 🚨🚓🚨

nsdeschenes (approved):
Some console logs to remove

calvin-codecov (approved):

PR #3867, opened 2025-05-15
fix: Table incorrectly sorting first page of results when sorted

ajay-sentry:
# Description

This PR aims to fix a bug where the first page of results returned from a sorted column on a table were reversed. Interestingly, the table had the correctly sorted results on the ascending case for Firefox and the descending case for Chrome, but failed in the opposite case for either.

After removing this additional table option somehow everything worked again, so probably some weird interaction reversing reversed results or something with the tanstack table library.

Closes https://github.com/codecov/engineering-team/issues/3583

# Screenshots

**CHROME**
https://github.com/user-attachments/assets/08ae42ea-2faf-4c6e-a8a6-73e76bdcb17f

**FIREFOX**
https://github.com/user-attachments/assets/eaa41615-66fc-46c2-ac9e-3ce8ed9423ed

# Link to Sample Entry

# Legal Boilerplate


suejung-sentry (approved):
nice find, that was a weird one

PR #3869, opened 2025-05-16
perf: Minor extract coverage util perf updates

ajay-sentry:
# Description

Just familiarizing myself with the File Explorer code and in the process found a couple minor optimizations we could do.

Namely removing the 2 lodash functions for some JS built-ins. Also potentially a bug fix where we had (commit || branch) which if commit didn't have coverage analytics but branch did would result in an error / premature return when we didn't need to.

# Screenshots
**before**
Screenshot 2025-05-16 at 1 45 02 PM

**after**
Screenshot 2025-05-16 at 1 45 52 PM

These aren't scientific but yeah this is mostly yak shaving anyway

# Link to Sample Entry

# Legal Boilerplate

### DIFF
--- a/src/services/pathContents/useFileWithMainCoverage.tsx
+++ b/src/services/pathContents/useFileWithMainCoverage.tsx
@@ -11,8 +11,14 @@ import {
 } from './constants'
 import { extractCoverageFromResponse } from './utils'
 
-// There's only 1 hook per page table as of Feb 22, 2023. This is a limitation to tailor a table to a particular page. Each "page" should have an independent set of hooks for directory and file entries. Every usePrefetchFile<page name> hook needs to this hook's query and queryKey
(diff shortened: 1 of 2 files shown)


src/services/pathContents/useFileWithMainCoverage.tsx
```diff
@@ -11,8 +11,14 @@ import {
 } from './constants'
 import { extractCoverageFromResponse } from './utils'
 
-// There's only 1 hook per page table as of Feb 22, 2023. This is a limitation to tailor a table to a particular page. Each "page" should have an independent set of hooks for directory and file entries. Every usePrefetchFile<page name> hook needs to this hook's query and queryKey
-// There should be a coverageForFile + prefetch hook set per page. This function, due to how it's written, acts as 1 hook all pages use, and each file implements it's prefetch function accordingly. This is something that needs to be changed, likely in a subsequent PR, as we're getting rid of the individual line file type layout anyway and doing diff line only
+/* There's only 1 hook per page table as of Feb 22, 2023. This is a limitation to
```
ajay-sentry, line 14: This comment was way exceeding the typical line length so just reformatted

spalmurray (approved):

PR #3878, opened 2025-05-28
feat: Fallback to Customer Address if default payment method address doesn't exist

ajay-sentry:
# Description

This PR adds a fallback mechanism to the address component to use the "customer address" if that exists instead. This is actually what stripe / we do under the hood anyway, we just want to make it clear to the customer as well.

FYI If the customer changes their billing address using this form, both their customer address AND the default payment method address are set.

# Link to Sample Entry

# Legal Boilerplate


src/services/account/useAccountDetails.ts
```diff
@@ -102,7 +104,7 @@ export const SubscriptionDetailSchema = z
             expires: z.number().nullable(),
           })
           .nullish(),
-        email: z.string(),
+        name: z.string().nullable(),
```
suejung-sentry, line 107: is name `nullish`? I feel like I've seen it undefined in the past

src/services/account/useAccountDetails.ts
```diff
@@ -93,6 +93,8 @@ export const SubscriptionDetailSchema = z
     currentPeriodEnd: z.number(),
     customer: z
       .object({
+        address: AddressSchema.nullable(),
```
suejung-sentry, line 96: is this one nullish too or is customer address always set? I also feel like i've seen "dirty" old data where not all fields were required so are unset on older customers

suejung-sentry (approved):
Nice thanks for doing this. Looks like we're good to get the dependent [version](https://github.com/codecov/umbrella/pull/189/files) of api deployed in all environments (incl dec) before / with this gazebo PR so good to go there

src/services/account/useAccountDetails.ts
ajay-sentry: It actually should have always been undefined in the past because it was never on the StripeSerializer to begin with, but based off the stripe docs, they call it nullable: https://docs.stripe.com/api/customers/object

suejung-sentry: Ah got it, cool. As long as all our code in-between continues to treat it as `null` should be good to go. Worst case if we see it error can make adjustments as it comes up haha

PR #3898, opened 2025-06-18
feat: add total duration section

joseph-sentry:
We're adding a time spent column to display the total time spent on each test for the selected time period. The goal is to give more granular information about the total impact of each test.

image


spalmurray (dismissed):
lgtm. You're basically a frontend dev

ajay-sentry (dismissed):
changes proposed in the PR look good, had another comment tho

joseph-sentry:
I'm going to end up leaving the slowest tests header in and not make any changes to the header section and just add the time spent column in this change

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -225,11 +235,30 @@ const FailedTestsTable = () => {
     },
   })
 
-  const isDefaultBranch = testData?.defaultBranch === branch
-  const isTeamOrFreePlan = testData?.isTeamPlan || testData?.isFreePlan
-  // Only show flake rate column when on default branch for pro / enterprise plans or public repos
+  const [isDefaultBranch, setIsDefaultBranch] = useState(false)
```
ajay-sentry, line 238: what's up with all this stuff? Maybe we can do a use memo instead if we want to keep the state cached. Something like:

```
const isTeamOrFreePlan = useMemo(() => {
  if (testData?.isTeamPlan !== null && testData?.isFreePlan !== null) {
    return testData.isTeamPlan || testData.isFreePlan
  }
  return true // hide flake stuff by default maybe?
}, [testData?.isTeamPlan, testData?.isFreePlan])
```

etc. etc.

src/pages/RepoPage/FailedTestsTab/FailedTestsPage/FailedTestsTable/FailedTestsTable.tsx
```diff
@@ -352,9 +390,7 @@ const FailedTestsTable = () => {
                   >
                     <div
                       className={cs('flex gap-1', {
-                        'flex-row-reverse': !['name', 'updatedAt'].includes(
-                          header.id
-                        ),
+                        'flex-row-reverse': header.id !== 'name',
```
ajay-sentry, line 393: we still want this change?

PR #3905, opened 2025-07-02
feat: Use new default-org on internal user for default org decision

spalmurray:
Main motivation for this is plan page troubles faced by users. A common flow we're trying to fix is: User clicks Learn more on plans page which takes them to our marketing site. They they click Get started from the marketing site, which takes them back to the app, but the default org logic fails and they end up on their personal org instead of the org they started with. We have had several instances of users upgrading the plan of their personal org instead of the desired org from this flow. We made some other improvements to the checkout page in a previous PR, but we also want to get this part fixed as it's likely the primary problem.

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -13,7 +13,6 @@ import { type Mock, vi } from 'vitest'
 import config from 'config'
 
 import { useLocationParams } from 'services/navigation/useLocationParams'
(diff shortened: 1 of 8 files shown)


ajay-sentry (approved):
niceeeeee, having all this stuff be derived in the API moving forward is a huge win

PR #3909, opened 2025-07-09
fix(address-form): Update error message for address form

suejung-sentry:
Paired with this https://github.com/codecov/umbrella/pull/298/files

Properly show the user the error

Screenshot 2025-07-09 at 4 05 37 PM

### DIFF
--- a/src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/Address/AddressForm.tsx
+++ b/src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/Address/AddressForm.tsx
@@ -134,14 +134,14 @@ export const getErrorMessage = (
     if (error.message) {
       return `Could not save billing address: ${error.message}`
     }
-    return 'Could not save billing address. Please contact support for assistance.'
+    return 'Could not save billing address. Please contact support at support@codecov.io for assistance.'
   }
 
   if (error.data?.detail) {
     return `Could not save billing address: ${error.data.detail}`
   }
 
-  return 'Could not save billing address due to an unknown error. Please contact support for assistance.'
+  return 'Could not save billing address due to an unknown error. Please contact support at support@codecov.io for assistance.'
 }
 
 export default AddressForm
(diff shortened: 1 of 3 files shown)


src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/Address/AddressForm.tsx
```diff
@@ -123,4 +125,23 @@ function AddressForm({
...
+export const getErrorMessage = (
+  error: Error | BillingApiError
+): string | undefined => {
+  if (!error) return undefined
+
+  if (error instanceof Error) {
+    if (error.message) {
+      return `Could not save billing address: ${error.message}`
+    }
+    return 'Could not save billing address. Please contact support for assistance.'
```
ajay-sentry, line 137: nit: we should add the email here too if this is the copy we wanna show. I.e. contact support at support@codecov.io or w/e it is (not sure myself haha)

ajay-sentry: Similar on line 144

src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/Address/AddressForm.tsx
```diff
@@ -50,7 +53,6 @@ function AddressForm({
     mutate: updateAddress,
     isLoading,
     error,
-    reset,
```
ajay-sentry, line 53: do u know why we were able to get rid of this?

ajay-sentry: oh is it because we want the error to persist?

src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/Address/AddressForm.tsx
suejung-sentry: Updated! 
looks like rest of app says support@codecov.io so used that

src/pages/PlanPage/subRoutes/CurrentOrgPlan/BillingDetails/Address/AddressForm.tsx
suejung-sentry: It seemed the former check of `!reset` would always be falsy because `reset` is a function. The form otherwise didn't require it. I think `useMutation` doesn't even return a property `reset: () => void` so that may be a leftover from some long-ago version of this that used react Form

ajay-sentry (approved):

PR #3928, opened 2025-08-29
feat: add all branches as default capability to test analytics page

adrianviquez:
# Description
This PR adds changes so that the default value for TA pages is "All Branches" instead of the default branch.


ajay-sentry (approved):
played around on preview deploy, had a small question for you about reset to default behavior and the null/undefined thing. Otherwise lgtm

adrianviquez:
@ajay-sentry, this is basically the new behavior: instead of defaulting to the default branch, we'll default to "all branches", which we treat as null in the backend and as empty in the URL route. Hence why we're explicitly defining as null

PR #4004, opened 2026-06-18
feat: Add sentry login deprecation banner

calvin-codecov:
Add banner to notify Sentry login users that the option is going to be deprecated soon.

Screenshot 2026-06-18 at 3 35 46 PM


src/services/user/useUser.ts
```diff
@@ -82,14 +83,14 @@ const currentUserFragment = `
 fragment CurrentUserFragment on Me {
   owner {
     defaultOrgUsername
-    isOnlyUsingSentryApp
```
calvin-codecov, line 85: Removing this as it is leftover and was not removed even though unnecessary.

drazisil-codecov (approved):

